PR #2, opened 2023-02-14
Add zod model & form validation

chohner:
First draft for end-to-end schema validation.
Will update var naming & error messages later


joschka:
@chohner very nice :+1: I liked zod already, but the form integration is genius.

PR #67, opened 2023-04-20
Add Box component

joschka:
needs a bit more polishing, but to get the idea...

### DIFF
--- a/app/components/Background.tsx
+++ b/app/components/Background.tsx
@@ -1,54 +1,25 @@
 import classNames from "classnames";
 import React from "react";
+import type { CommonWrapperProps } from ".";
+import { BACKGROUND_COLORS } from ".";
 
 const DEFAULT_PADDING_TOP = "40";
 const DEFAULT_PADDING_BOTTOM = "48";
 
-export interface BackgroundProps {
-  color?: "default" | "white" | "blue" | "yellow" | "green" | "yellow" | "red";
-  paddingTop?:
-    | "default"
-    | "0"
-    | "8"
-    | "16"
-    | "24"
-    | "32"
-    | "40"
-    | "48"
-    | "56"
-    | "64";
-  paddingBottom?:
-    | "default"
-    | "0"
-    | "8"
-    | "16"
-    | "24"
-    | "32"
-    | "40"
-    | "48"
-    | "56"
-    | "64";
+export type BackgroundProps = {
   children: React.ReactNode;
-}
+} & CommonWrapperProps;
 
 export default function Background({
-  color = "default",
+  backgroundColor = "default",
   paddingTop = "default",
   paddingBottom = "default",
   children,
 }: BackgroundProps) {
   const cssClasses = classNames(
+    backgroundColor !== "default" && BACKGROUND_COLORS[backgroundColor],
     `pt-${paddingTop === "default" ? DEFAULT_PADDING_TOP : paddingTop}`,
-    `pb-${
-      paddingBottom === "default" ? DEFAULT_PADDING_BOTTOM : paddingBottom
-    }`,
-    {
-      "bg-white": color === "white",
-      "bg-blue-100": color === "blue",
-      "bg-yellow-300": color === "yellow",
-      "bg-green-200": color === "green",
-      "bg-[#f9e5ec]": color === "red",
-    }
+    `pb-${paddingBottom === "default" ? DEFAULT_PADDING_BOTTOM : paddingBottom}`
   );
 
   return <div className={cssClasses}>{children}</div>;
--- a/app/components/Background.tsx
+++ b/app/components/Background.tsx
@@ -1,0 +1,55 @@
+import classNames from "classnames";
+import React from "react";
+
+const DEFAULT_PADDING_TOP = "40";
+const DEFAULT_PADDING_BOTTOM = "48";
+
+export interface BackgroundProps {
+  color?: "default" | "white" | "blue" | "yellow" | "green" | "yellow" | "red";
+  paddingTop?:
+    | "default"
+    | "0"
+    | "8"
+    | "16"
+    | "24"
+    | "32"
+    | "40"
+    | "48"
+    | "56"
+    | "64";
+  paddingBottom?:
+    | "default"
+    | "0"
+    | "8"
+    | "16"
+    | "24"
+    | "32"
+    | "40"
+    | "48"
+    | "56"
+    | "64";
+  children: React.ReactNode;
+}
+
+export default function Background({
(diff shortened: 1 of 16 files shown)


app/components/Background.tsx
SannyNguyenHung, line 1: Does it make sense to use a complete div for the background? Can't we just use a function that takes care of the classnames and use it like?

`<Container className={generateFancyCSS(context)} />`

app/components/Background.tsx
```diff
@@ -0,0 +1,55 @@
+import classNames from "classnames";
+import React from "react";
+
+const DEFAULT_PADDING_TOP = "40";
+const DEFAULT_PADDING_BOTTOM = "48";
+
+export interface BackgroundProps {
+  color?: "default" | "white" | "blue" | "yellow" | "green" | "yellow" | "red";
+  paddingTop?:
```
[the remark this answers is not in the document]
SannyNguyenHung: Same with background colors

app/components/Box.tsx
```diff
@@ -0,0 +1,38 @@
...
+  heading?: HeadingProps;
+  label?: string;
+  content?: ParagraphProps;
+  button?: ButtonProps;
+}
+
+const Box = ({ heading, label, content, button }: BoxProps) => {
+  return (
+    <div className="ds-stack stack-8">
+      {label ? <div className="ds-label-02-bold">{label}</div> : ""}
```
SannyNguyenHung, line 18: {label && <div className="ds-label-02-bold">{label}</div>} is better to prevent empty strings in the HTML dom or?

app/components/Container.tsx
```diff
@@ -1,28 +1,65 @@
...
+    | "0"
+    | "8"
+    | "16"
+    | "24"
+    | "32"
+    | "40"
+    | "48"
+    | "56"
+    | "64";
+  backgroundColor?: "default" | "white" | "blue" | "yellow";
```
SannyNguyenHung, line 30: We can define standard colors (default, white, blue, green, red) and filter/limiting the selection in the cms

app/components/Container.tsx
```diff
@@ -1,28 +1,65 @@
...
+    `pb-${paddingBottom === "default" ? DEFAULT_PADDING_BOTTOM : paddingBottom}`
+  );
+
+  if (backgroundColor === "default") {
+    return <div className={cssClasses}>{children}</div>;
+  }
+
+  cssClasses = classNames(cssClasses, {
+    "relative before:content-[''] before:absolute before:inset-y-0 before:-left-32 before:-right-32":
+      backgroundColor,
```
SannyNguyenHung, line 52: How does this work? Shouldn't there be a boolean value instead of the default value of backgroundColor?

app/components/PageContent.tsx
```diff
@@ -4,11 +4,57 @@ import Header from "./Header";
...
+
+const wrapInContainer = (
+  cmsData: { [key: string]: any; container?: ContainerCMS },
+  reactElement: ReactNode
+): ReactNode => {
+  if (!cmsData.container) {
+    return reactElement;
+  }
+  const config = transformCmsData(cmsData.container);
+  console.log({ config });
```
SannyNguyenHung, line 43: do need the console log here?

app/services/cms/models/metaComponents.ts
```diff
@@ -0,0 +1,57 @@
...
+    | "px8"
+    | "px16"
+    | "px24"
+    | "px32"
+    | "px40"
+    | "px48"
+    | "px56"
+    | "px64";
+  backgroundColor: "default" | "white" | "blue" | "yellow";
+};
```
SannyNguyenHung, line 55: Why is the format of the padding "px0" instead of "0px"? Can't we use ContainerProps & BackgroundProps here?

```
export type Container = {
  id: number;
  __component: "meta.container";
  ...ContainerProps
}
```

SannyNguyenHung:
Thanks Joschka! This will simplify many things! I luv it :smiley:

app/components/Background.tsx
joschka: The idea is to have a separate container, which does the centering and max-width and might also have a background (yellow). Plus another "container" around it which does just the background color + some padding (light blue).
box-backgrounds

app/components/Box.tsx
joschka: My vage memory told me that ternary operator is safer. But I looked it up in docs and `&&` should be fine here. Colud only be a problem when `label` could equal `0`. Because then "0" would be rendered. 
https://react.dev/learn/conditional-rendering#conditional-ternary-operator-- 
> React considers false as a “hole” in the JSX tree, just like null or undefined, and doesn’t render anything in its place.

app/components/Container.tsx
joschka: When `backgroundColor` equals "default" we already return a few lines above and never reach the line here. But while I'm typing this I realize that we can remove the entire check. In that line we always have a non-default backgroundColor defined. Will fix that.

app/components/PageContent.tsx
```diff
@@ -4,11 +4,57 @@ import Header from "./Header";
...
 };
 
+const transformCmsData = (
+  cmsData: ContainerCMS | BackgroundCMS
+): ContainerProps | BackgroundProps => {
+  console.log({ cmsData });
+  return {
+    ...cmsData,
+    //@ts-ignore
+    paddingTop: cmsData.paddingTop.replace(/^px/, ""),
```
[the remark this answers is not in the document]
joschka: It's `px32` because of a cms limitation. In enumerations values must start with a character. That's why I put the "px" in front. A bit unfortunate.

app/components/PageContent.tsx
joschka: Will clean that up.

joschka:
@SannyNguyenHung Thank you very much for the comments. I appreciate the feedback. As a next step I will fix and clean some things and resolve the conflicts with main.

app/routes/datenschutz.tsx
```diff
@@ -19,9 +18,5 @@ export const loader: LoaderFunction = async ({ request }) => {
 };
 
 export default function Index() {
-  return (
-    <Container className="pt-16 pb-80 bg-blue-100 min-h-[50vh]">
-      <PageContent content={useLoaderData<typeof loader>().content} />
-    </Container>
-  );
+  return <PageContent content={useLoaderData<typeof loader>().content} />;
```
SannyNguyenHung, line 21: We can actually route the privacy & imprint dynamically, like $pageSlug.tsx :grin:

PR #106, opened 2023-05-16
Xstate

SannyNguyenHung:
Hey, 

in this pull request i have replaced the flow logic with xstate. Partially still any as types but we need to take a more dynamic approach than a type safe one. I have removed the treecalculation.ts & flow.server.ts and replaced it with the xstate config at: app/services/flow/consultingcosthelp.json. You can find a stately link here: 
https://stately.ai/registry/editor/1b59901b-5b8e-419e-9c35-a520ba1c3a55?machineId=1a4b8735-cb12-492e-aa85-23da948d32d9

Node conditions are in guard.ts - can be swapped out. What do you think? 

In general I'm looking forward to your feedback!


app/services/flow/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+          stack.push({ node: target, currentStepID: target.key });
+        }
+      }
+    }
+  }
+
+  return possiblePaths;
+}
+
+export function getProgressBar(stepID: string, context: any) {
```
frederike-ramin, line 154: Did someone say "Test"? :D

tests/e2e/vorabcheck.spec.ts
```diff
@@ -88,11 +88,6 @@ test("vorabcheck can be traversed", async ({ page }) => {
   ).toHaveCount(1);
 });
 
-test("funnel: invalid context redirects to start", async ({ page }) => {
-  await page.goto(`${vorabcheck.url}/eigeninitiative`);
-  await expect(page).toHaveURL(`${vorabcheck.url}/${vorabcheck.initialStep}`);
```
frederike-ramin, line 93: I think we can also enable this by checking if it's a valid step in the loader.

frederike-ramin (commented):
Thank you so much for doing this. I really think we should improve the test situation. Everything else seems like optional or possibly later improvements to me

app/lib/vorabcheck/pages.ts
```diff
@@ -5,7 +5,7 @@ import { einkommenKinderStep } from "~/components/form/steps/einkommenKinder";
 import { kinderAnzahlSimpleStep } from "~/components/form/steps/kinderAnzahlSimple";
 import { verfuegbaresEinkommenStep } from "~/components/form/steps/verfuegbaresEinkommen";
 
-export const formPages = {
+export const formPages: { [key: string]: any } = {
```
chohner, line 8: Why do we want to type this even looser? Also, I think `Record<string, string>` is more readable

app/routes/vorabcheck.$stepID.tsx
```diff
@@ -66,9 +65,9 @@ const getReasonsToDisplay = (
 };
 
 export const loader: LoaderFunction = async ({ params, request }) => {
-  const stepID = params.stepID as AllowedIDs;
-  if (!formGraph.hasNode(stepID)) {
-    return redirect(`/vorabcheck/${initialStepID}`);
+  const stepID = params.stepID?.replace("#", "");
+  if (!hasStep(stepID) || stepID === undefined) {
```
chohner, line 69: doesn't `hasStep()` deal with `undefined` already?

app/services/flow/guards.tsx
```diff
@@ -0,0 +1,82 @@
+import { freibetrag } from "~/lib/freibetrag";
+import moneyToCents from "../validation/money/moneyToCents";
+
+function searchEntryInSession(
+  context: any,
+  stepID: any,
+  value: string
+): boolean {
+  let found = false;
+  const values = context[stepID];
```
chohner, line 10: I'm surprised typescript isn't complaining here. I'd expect a check for `stepID in context` first 🤔

chohner: Probably because of all the `any` types  😇
Probably better to include the safety checks?

app/services/flow/guards.tsx
```diff
@@ -0,0 +1,82 @@
...
+  if (values !== undefined) {
+    const propertyValues = Object.values(values);
+    for (const entry in propertyValues) {
+      if (propertyValues[entry] === value) {
+        found = true;
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
chohner: Agreed, or `find()`

app/services/flow/guards.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+function anyNonCriticalWarning(context: any) {
+  return (
+    context.kostenfreieBeratung?.hasTriedFreeServices == "no" ||
+    context.eigeninitiative?.hasHelpedThemselves == "no"
+  );
+}
+
+export function isIncomeTooHigh(context: any) {
```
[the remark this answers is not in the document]
chohner: Could we maybe break this into a normalization and a calculation part? This is going to be very hard to test

app/services/flow/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+        break;
+      }
+    }
+  } else if (possiblePreviousSteps.length === 1) {
+    result = possiblePreviousSteps[0];
+  }
+
+  if (!result) {
+    result = getInitialStep();
+  }
```
chohner, line 69: Maybe easier to just initialize `let result = getInitialStep();` ?

app/services/flow/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const stateMachine = getStateMachine(stepID, context);
+
+  if (!hasStep(stepID)) {
+    return stateMachine.initial;
+  }
+
+  return stateMachine.transition(stepID, { type: "SUBMIT" }).value;
+}
+
+function getPossiblePath(
```
chohner, line 124: we could probably simplify this using https://xstate.js.org/docs/packages/xstate-graph/

app/services/flow/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  const stateMachine = getStateMachine(stepID, undefined);
+  const stateNode = stateMachine.getStateNodeById(
+    `${stateMachine.key}.${stepID}`
+  );
+  return stateNode.transitions.length === 0;
+}
+
+export function getPreviousStep(stepID: string, context: any) {
```
frederike-ramin, line 38: Just wanted to add a comment here because I just remembered our conversation a couple days ago: When testing this, we need to make sure that it also works for not-filled-in-order forms.

app/routes/vorabcheck.$stepID.tsx
```diff
@@ -113,37 +110,29 @@ export const loader: LoaderFunction = async ({ params, request }) => {
...
-      destinationString = link.target as AllowedIDs;
-      break;
-    }
-  }
+  session.set(stepID, validationResult.data);
   const headers = { "Set-Cookie": await commitSession(session) };
-  return redirect(`/vorabcheck/${destinationString}`, { status: 302, headers });
+
+  let destinationString = getNextStep(stepID, session.data);
+  return redirect(`/vorabcheck/${String(destinationString)}`, {
```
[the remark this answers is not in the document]
frederike-ramin: Turns out - it's a symbol, not a string. So I'm just renaming the constant.

app/services/flow/index.tsx
frederike-ramin: I thought this was a really good suggestion. Unfortunately, it does not work with our current form logic:
- when you have nothing filled, the possible graph will only contain one step (as the default always goes the shortest way right now
- if we rework this - that the default is the longest path to success, the funnel logic as it is now will not work (with invalid data you'll just get the previous step instead of the first one)

Therefore, I'd park this for now and merge without this change.

frederike-ramin: I also tried it out here: https://github.com/digitalservicebund/a2j-rechtsantragstelle/tree/graph-progress-bar

PR #118, opened 2023-05-23
CMS Types

joschka:
* getting data from the strapi file dump does now respect a given locale
* many UI components now have a zod schema for their props
* CMS types now have all zod schemas, also renamed things
* simplified the cms service and added an unified "api" (getStrapiPage, getStrapiFooter, ...) where the return values are typed
* Changed the "importing" of content.json, to a static import instead of using fs.readFile (simpler and better performance; only downside: content.json must exist, at least with an empty object inside
* the is now a translation service "props" to translate cms data to component props. Allows components to have another props structure than what comes from Strapi, good example: Image component


app/services/cms/models/BackgroundColor.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from "zod";
+
+export const BackgroundColorSchema = z.union([
+  z.literal("default"),
+  z.literal("white"),
+  z.literal("blue"),
+  z.literal("yellow"),
+  z.literal("green"),
+  z.literal("red"),
+]);
```
chohner, line 10: nit: Maybe easier expressed with an enum  `z.enum(["default", "white", ....]);`?

app/services/cms/models/Background.ts
```diff
@@ -0,0 +1,11 @@
...
+import { WrapperSchema } from "./Wrapper";
+
+export const BackgroundSchema = WrapperSchema.merge(
+  z.object({
+    id: z.number(),
+    __component: z.literal("meta.background"),
+  })
+);
+
+export type Background = z.infer<typeof WrapperSchema>;
```
chohner, line 11: this should export `BackgroundSchema`?

app/services/cms/models/Container.ts
```diff
@@ -0,0 +1,11 @@
...
+import { WrapperSchema } from "./Wrapper";
+
+export const ContainerSchema = WrapperSchema.merge(
+  z.object({
+    id: z.number(),
+    __component: z.literal("meta.container"),
+  })
+);
+
+export type Container = z.infer<typeof WrapperSchema>;
```
chohner, line 11: `ContainerSchema`?

app/services/cms/models/ElementWithId.ts
```diff
@@ -1,11 +1,16 @@
...
-};
+export const ElementWithIdSchema = LocalizableSchema.merge(
+  TimestampableSchema
+).merge(
+  z.object({
+    id: z.number().optional(),
+    elementId: z.string(),
+    element: z.array(FormContentCmsSchema),
+  })
+);
```
chohner, line 14: nit: I think merging the other way around might be easier to read?

```
export const ElementWithIdSchema =  z.object({
    id: z.number().optional(),
    elementId: z.string(),
    element: z.array(FormContentCmsSchema),
  }).merge(LocalizableSchema).merge(TimestampableSchema);
```

app/services/cms/models/Background.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from "zod";
+import { WrapperSchema } from "./Wrapper";
+
+export const BackgroundSchema = WrapperSchema.merge(
+  z.object({
+    id: z.number(),
+    __component: z.literal("meta.background"),
+  })
+);
```
chohner, line 9: nit: for me `z.object({}).merge(WrapperSchema)` might express this type more clear

app/services/cms/models/Heading.ts
```diff
@@ -0,0 +1,36 @@
...
+  tagName: z.union([
+    z.literal("h1"),
+    z.literal("h2"),
+    z.literal("h3"),
+    z.literal("h4"),
+    z.literal("h5"),
+    z.literal("h6"),
+    z.literal("p"),
+    z.literal("div"),
+  ]),
```
chohner, line 16: nit: same here and below - `z.enum` might be a bit more ergonomic

app/services/cms/models/Localizable.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from "zod";
+import { LocaleSchema } from "./Locale";
+
+export const LocalizableSchema = z.object({
```
chohner, line 4: nit: I like the zod naming of `HasX` (so here: `HasLocale`)

app/services/cms/models/BackgroundColor.ts
joschka: Absolutely. I guess I mixed it up with TS enums in my mind.

app/services/cms/models/ElementWithId.ts
joschka: True

app/services/cms/models/Localizable.ts
joschka: Nice, way better than ---able.

app/components/Box.tsx
```diff
@@ -1,16 +1,17 @@
...
+import { z } from "zod";
 
-export interface BoxProps {
-  label?: HeadingProps;
-  heading?: HeadingProps;
-  content?: ParagraphProps;
-  button?: ButtonProps;
-}
+export const BoxPropsSchema = z.object({
+  label: HeadingPropsSchema.optional(),
```
chohner, line 8: `label` & `HeadingPropsSchema`?

app/services/cms/index.ts
```diff
@@ -0,0 +1,73 @@
...
+const getStrapiEntryFromSource =
+  config().CMS === "FILE" ? getStrapiEntryFromFile : getStrapiEntryFromApi;
+
+const getStrapiEntry = async (
+  opts: Pick<GetStrapiEntryOpts, "apiId" | "slug"> & {
+    locale?: StrapiLocale;
+  }
+) => {
+  const data = await getStrapiEntryFromSource({ locale: "de", ...opts });
+  // "remove attributes key"
```
frederike-ramin, line 27: Nit-picky: Why is the comment in quotes? Is there a specific reason/ practice that I'm just no aware of? :D

app/services/cms/models/StrapiBackground.ts
```diff
@@ -0,0 +1,14 @@
+import { z } from "zod";
+import { HasStrapiIdSchema } from "./HasStrapiId";
+import { StrapiWrapperSchema } from "./StrapiWrapper";
+
+export const StrapiBackgroundSchema = z
+  .object({
+    id: z.number(),
+    __component: z.literal("meta.background").optional(),
```
frederike-ramin, line 8: Why is the component literal optional? I think for a Background it should always be given? (same for the other components of course)

frederike-ramin (approved):
Nice. Looks really good, thank you for cleaning that up - feels a lot more stable that way. 👍

app/services/cms/index.ts
joschka: I couldn't find a better wording. I want to say that I kind of remove the attributes key "in between".
`{ bla: { attributes: foobar } }`  => `{ bla: foobar }`. Not removing the whole key with value (`{ bla: {} }`). Remove a layer in between. Any idea?

frederike-ramin: Nono, I got that part. But shouldn't it be 
`  // remove "attributes" key`
then?

app/services/cms/models/StrapiBackground.ts
joschka: :man_shrugging: sometimes it's there, sometimes not


kidding


Components that are nested inside another component don't have the `__component` field in the api response. For example: Heading at the top level inside content area of a page -> has `__component` field. Heading inside InfoBox -> no `__component` field.

app/components/form/steps/kinderAnzahl.tsx
```diff
@@ -25,7 +25,9 @@ export const kinderAnzahlStep = {
             defaultValue={defaultValues?.[inputElement.name] || "0"}
             label={inputElement.label}
             type={inputElement.type}
-            placeholder={inputElement.placeholder}
+            placeholder={
+              inputElement.placeholder ? inputElement.placeholder : undefined
```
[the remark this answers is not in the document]
joschka: And `null`. In cms data, optional things are `null` when not set. But they are defined. In component props world optional things are expected to be `undefined`. Translation from `null` to `undefined`.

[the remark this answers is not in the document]
frederike-ramin: But I thought we're already getting rid of null during the strapi -> props translation?

app/services/cms/index.ts
chohner: How about adding your example there? just so someone else doesn't need to dig through this PR :)

joschka:
```
// remove "attributes" key
// see https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/118 for more information
```

app/components/form/steps/kinderAnzahl.tsx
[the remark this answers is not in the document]
joschka: I changed my mind. Lets keep it for now and make it nicer when we refactor the steps.

PR #146, opened 2023-06-12
ci: split quality gates into parallel jobs

chohner:


joschka:
@chohner This looks really good. Looking forward see how this looks like when a job fails. Do I see directly that for example "prettier" failed?

PR #202, opened 2023-07-14
feat: add CSRF header to steps

chohner:
Adds csfr header validation to all step submissions:
- root returns key in loader and session
- step renders key in hidden form element & checks match in action

### DIFF
--- a/app/services/security/csrf.server.ts
+++ b/app/services/security/csrf.server.ts
@@ -1,0 +1,48 @@
+//https://sergiodxa.com/articles/adding-csrf-protection-to-remix
+import type { Session } from "@remix-run/node";
+import { randomBytes } from "crypto";
+import { getSession } from "~/sessions";
+import { CSRFKey } from "./csrf";
+
+export function createCSRFToken() {
+  return randomBytes(100).toString("base64");
+}
+
+export async function validateCSRFToken(request: Request, session: Session) {
+  // first we parse the body, be sure to clone the request so you can parse the body again in the future
+  const body = Object.fromEntries(
+    new URLSearchParams(await request.clone().text()).entries(),
+  ) as { csrf?: string };
+  // then we throw an error if one of our validations didn't pass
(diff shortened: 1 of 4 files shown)


app/services/security/csrf.server.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export async function createCSRFSession(request: Request) {
+  const session = await getSession(request.headers.get("Cookie"));
+  const csrf = createCSRFToken();
+  session.set(CSRFKey, csrf);
+  return { session, csrf };
+}
+
+export async function validatedSession(request: Request) {
```
pgurusinga, line 36:
```suggestion
export async function validateCSRFSession(request: Request) {
```
just to follow the createCSRFSession

app/services/security/csrf.server.ts
```diff
@@ -0,0 +1,48 @@
...
+export async function validateCSRFToken(request: Request, session: Session) {
+  // first we parse the body, be sure to clone the request so you can parse the body again in the future
+  const body = Object.fromEntries(
+    new URLSearchParams(await request.clone().text()).entries(),
+  ) as { csrf?: string };
+  // then we throw an error if one of our validations didn't pass
+  if (!session.has(CSRFKey))
+    throw new Error("Session: CSRF Token not included.");
+  if (!body.csrf) throw new Error("Body: CSRF Token not included.");
+  if (body.csrf !== session.get(CSRFKey))
```
pgurusinga, line 20: @chohner I think we set `csrf` token in the Remix session, so when we call this function in the `vorabcheck` flow, it would throw an error.

app/services/security/csrf.server.ts
chohner: I wanted to express that the function returns the session itself (or throws an error). happy to iterate the naming though

SannyNguyenHung (approved):

pgurusinga (approved):

PR #212, opened 2023-07-21
Update ci-pipeline.yml

chrkaatz:
Move deployment reporting before the actual deployment

As we do not track initial deployments, we cannot track failed ones later neither


SannyNguyenHung (approved):
Looks fine for me :+1:

joschka (commented):
:+1: top. thank you

PR #227, opened 2023-08-01
Implement session namespace for different contexts

SannyNguyenHung:
In this pull request I want to add a context to our session. This way we now have a total of 3 different session contexts: main, beratungshilfe and onlineverfahren. The 'main' context can store information that is not context specific, such as the CSRF token.

When a user performs a Beratungshilfe, a main & beratungshilfe cookie is created in the frontend and each one has a different unique ID. This ID is also stored in the Redis database under the keys 'main_' and 'beratungshilfe_'.

What do you guys think of this approach? The idea is that we can easily add more contexts by adding another session (like beratungshilfe.session.ts) and calling the corresponding session context.


app/routes/vorabcheck/step.tsx
```diff
@@ -97,6 +100,7 @@ export const action: ActionFunction = async ({ params, request }) => {
     });
   }
 
+  const beratungshilfeSesion = await getSession(request.headers.get("Cookie"));
```
frederike-ramin, line 103: Typo! :D `beratungshilfeSession`

app/services/session/index.ts
```diff
@@ -25,11 +26,23 @@ export async function setReturnToURL(request: Request, uuid: string) {
   }
 }
 
-export function createDatabaseSessionStorage({ cookie }: { cookie: Cookie }) {
+export enum SessionContext {
+  Main = "main",
+  Beratungshilfe = "beratungshilfe",
+  OnlineVerfahren = "onlineverfahren",
```
frederike-ramin, line 32: I think we can/should be more precise with our namespaces. So having not only `beratungshilfe` but `beratungshilfeVorabcheck` and `geldEinklagenVorabcheck`

frederike-ramin (commented):
Thank you for taking this over!

Originally, I imagined the differentiation on a different level, I think. I didn't expect multiple session IDs (and thus multiple cookies if I understand correctly?) to be created but instead have *one* session ID that references data from all forms. (just as a data object that has one key per form)
The main benefit of this to me is that we know which data belongs to which user - making debugging easier. Imagine a user having an error - could be a combination of data from different forms leading to that error - we'd have to ask the user for all of their session ids, not just one.
Of course, the invalidation is more isolated in your solution, allowing for more specific deletion of data.

Can we maybe take the best of both worlds and store *one* session ID per user in the cookie and use that same ID for *multiple* redis entries? So that we have `main_1234` and `beratungshilfe_1234`?

(Also I'm not 100% sure I understand how the sessionId ends up in the cookie in the first place tbh)

SannyNguyenHung:
> Thank you for taking this over!
> 
> Originally, I imagined the differentiation on a different level, I think. I didn't expect multiple session IDs (and thus multiple cookies if I understand correctly?) to be created but instead have _one_ session ID that references data from all forms. (just as a data object that has one key per form) The main benefit of this to me is that we know which data belongs to which user - making debugging easier. Imagine a user having an error - could be a combination of data from different forms leading to that error - we'd have to ask the user for all of their session ids, not just one. Of course, the invalidation is more isolated in your solution, allowing for more specific deletion of data.
> 
> Can we maybe take the best of both worlds and store _one_ session ID per user in the cookie and use that same ID for _multiple_ redis entries? So that we have `main_1234` and `beratungshilfe_1234`?
> 
> (Also I'm not 100% sure I understand how the sessionId ends up in the cookie in the first place tbh)

I implemented this in particular to avoid merge errors when making changes. By completely overwriting the entire object in Redis instead of individual elements, we can be sure that nothing will break when data changes. If we use a set in Redis instead, the entire data set will always be overwritten. I will try to implement your suggestion as I find it extremely useful! Thanks a lot for your feedback :)

app/services/session/index.ts
SannyNguyenHung: Unfortunately I can't do that, because geld-einklagen and beratungshilfe use the step.ts I have to follow the flowId so that I take over the right context

app/services/session/index.ts
```diff
@@ -39,34 +39,36 @@ export function createDatabaseSessionStorage({
   return createSessionStorage({
     cookie,
     async createData(data, expires) {
-      const uuid = `${context}_${crypto.randomUUID()}`;
-      await setDataForSession(uuid, data);
+      const uuid = crypto.randomUUID();
+      await setDataForSession(`${context}_${uuid}`, data);
```
frederike-ramin, line 43: Can you explain to me how that is making sure we only have one cookie? I don't think I get it.

SannyNguyenHung: because the uuid is given back and will be used to store within the cookie with the name "__session". So the uuid will be the same for each context but just the prefix will change

app/services/session/index.ts
chohner: maybe we should rename the flowIDs instead 🙈

app/services/session/index.ts
```diff
@@ -25,22 +27,46 @@ export async function setReturnToURL(request: Request, uuid: string) {
   }
 }
 
-export function createDatabaseSessionStorage({ cookie }: { cookie: Cookie }) {
+export type Context = "main" | "beratungshilfe" | "geld-einklagen";
```
chohner, line 30: I think this type name will get very confusing in the future because we also use `Context` to mean the object holding the variables. Maybe `SessionContext`?

chohner (approved):
Nice, I like the new `getSessionForContext()` interface 👍🏼

app/services/session/index.ts
SannyNguyenHung: I can't because the flow id is based on the url path

PR #300, opened 2023-09-15
chore: sort the package.json alphabetically

pgurusinga:
feel free to accept or reject this PR. 🙏


package.json
```diff
@@ -1,35 +1,36 @@
...
-    "build-storybook": "storybook build --output-dir=public/storybook",
-    "generate-opensource-licenses": "ts-node -T ./app/services/openSourceLicenses/generate.ts generateLicenseFile"
+    "typecheck": "tsc",
+    "update:URLs": "ts-node ./app/services/gerichtsfinder/checkURLs.ts checkURLs",
+    "update:gerichtsfinder": "ts-node ./app/services/gerichtsfinder/encryptedStorage.ts updateZip",
+    "verifyEmails": "ts-node ./app/services/cms/verifyExternalLinks.ts verifyEmails",
+    "verifyWebsites": "ts-node ./app/services/cms/verifyExternalLinks.ts verifyWebsites"
+  },
+  "prettier": {
+    "trailingComma": "all"
```
chohner, line 33: i think id prefer the main package settings up to, then dependencies, then tool settings. but no strong preference either way

pgurusinga: @chohner I think the trailingComma is already there since the beginning, now this PR just sorts the npm task instead... 🙏

chohner: I added it to match older prettier configs to the new defaults 😅
but in a previous iteration all keys of package.json were sorted (not just run tasks), which leads to unusual sorting imo

chohner (approved):
nice. in a next step we might want to think about more consistent naming :)

PR #431, opened 2023-11-24
Flow Navigation

chohner:
Adds styled navigation during flows

Example:
image

This also works without modifications for other flows:
image

Note: without adding a `done` method to a subflow meta it cannot receive a checkmark


app/models/flows/beratungshilfeFormular/grundvoraussetzung/context.ts
```diff
@@ -26,3 +26,13 @@ export const beratungshilfeGrundvoraussetzungenGuards = {
     context: BeratungshilfeGrundvoraussetzungen,
   ) => context.eigeninitiativeGrundvorraussetzung === "yes",
 };
+
+export const grundvoraussetzungDone = (
```
frederike-ramin, line 30: Intuitively I would have said this belongs rather into `/ guards` than `/context`?

app/models/flows/beratungshilfeFormular/grundvoraussetzung/flow.json
```diff
@@ -2,6 +2,7 @@
   "predictableActionArguments": true,
   "initial": "start",
   "id": "grundvoraussetzungen",
+  "meta": { "isUneditable": true },
```
frederike-ramin, line 5: Is this just always uneditable? Somehow, I didn't find the place where it gets set to false.

app/models/flows/beratungshilfeFormular/index.ts
```diff
@@ -34,19 +39,33 @@ export const beratungshilfeAntrag = {
...
       ),
       rechtsproblem: _.merge(_.cloneDeep(rechtsproblemFlow), {
+        meta: { done: rechtsproblemDone },
         states: {
           start: {
             on: {
               BACK: "#grundvoraussetzungen.eigeninitiativeGrundvorraussetzung",
             },
           },
+          danke: {
```
frederike-ramin, line 49: Haha, tiny, but not sure we want to call a page danke? :D

app/routes/shared/flowSpecifics.ts
```diff
@@ -74,11 +77,50 @@ export const flowSpecifics = {
     flow: _.merge(geldEinklagenFormularFlow, {
       states: {
         "persoenliche-daten": _.merge(_.cloneDeep(persoenlicheDatenFlow), {
+          meta: {
+            done: (context: GeldEinklagenFormularContext) =>
```
frederike-ramin, line 81: Two thoughts on different level of detail:
1. I think we should for sure place this function somewhere else and not have it inline
2. We could leverage the higher-level validation of zod here. Instead of just naming all fields here and seeing if they're set - the real question could also be if they're set correctly. So we could actually link to a zod object here that validates part of the full object? (Not the best explanation here, I could provide an example if you don't know what I mean, but maybe you already thought about it)

app/routes/shared/step.tsx
```diff
@@ -280,9 +284,23 @@ export function StepWithPreHeading() {
...
+          </div>
+        </div>
+        <div
+          className="ds-stack-40 w-full pb-48 pt-[14px]" // hacky alignment of pre-heading to navigation
+          style={{
+            // the padding below matches Container. look to re-use it instead
+            maxWidth: "59rem",
+            paddingLeft: "clamp(1rem, 5vw, 3rem)",
+            paddingRight: "clamp(1rem, 5vw, 3rem)",
+          }}
```
frederike-ramin, line 301: All of there look quite hacky. Can we find another way for this? maybe even just creating classes for it?

app/services/flowNavigation.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { FlowId, flowSpecifics } from "~/routes/shared/flowSpecifics";
+import { type buildFlowController } from "./flow/buildFlowController";
+
+export function navItemsFromFlowSpecifics(
+  currentStepId: string,
+  flowId: FlowId,
+  flowController: ReturnType<typeof buildFlowController>,
+): NavItem[] {
+  const currentFlow =
+    flowController.getFlow() as (typeof flowSpecifics)[keyof typeof flowSpecifics]["flow"];
```
frederike-ramin, line 11: Haha, this is so hard to read..wtf :D Should we introduce a type for this within the FlowSpecifics?

app/services/flowNavigation.ts
```diff
@@ -0,0 +1,40 @@
...
+    .filter(([_, state]) => "states" in state)
+    .map(([rootStateName, rootState]) => {
+      const destinationStepId = `${rootStateName}/${rootState.initial}`;
+      const destination = `/${flowId}/${destinationStepId}`;
+
+      let state = NavState.OpenDisabled;
+
+      if (currentStepId.startsWith(rootStateName)) {
+        state = NavState.Current;
+      } else if (flowController.isReachable(destinationStepId)) {
```
frederike-ramin, line 22: Uh..I love the usage of isReachable here!

app/services/flowNavigation.ts
```diff
@@ -0,0 +1,40 @@
+import { NavState, type NavItem } from "~/components/FlowNavigation";
+import type { FlowId, flowSpecifics } from "~/routes/shared/flowSpecifics";
+import { type buildFlowController } from "./flow/buildFlowController";
+
+export function navItemsFromFlowSpecifics(
```
frederike-ramin, line 5: This is quite some business logic here. I think this is worth some tests?

frederike-ramin (commented):
Nice approach overall. I really just have two bigger points: 1) an idea how we can approach the "done" condition differently and 2) let's add tests for the state logic.

PR #442, opened 2023-12-01
feat: use strapi translations

chohner:
Only merge after https://github.com/digitalservicebund/a2j-rechtsantragstelle-strapi/pull/184


pgurusinga:
nice

PR #485, opened 2023-12-20
feat(tracking): add custom events with distinct id

chohner:
- adds `sendCustomEvent()`
- tries to extract `distinct_id` from posthog cookie, otherwise falls back to `ENVIRONMENT`
- ~⚠️ currently sends ANY submitted step data ⚠️~

TODO
- [x] decide for which steps to fire (and how to name the event)

### DIFF
--- a/app/services/analytics/customEvent.ts
+++ b/app/services/analytics/customEvent.ts
@@ -1,0 +1,37 @@
+import { PostHog } from "posthog-node";
+import { hasTrackingConsent } from "./gdprCookie.server";
+import { config } from "../env/web";
+import { parse } from "cookie";
+
+const { POSTHOG_API_KEY, POSTHOG_API_HOST, ENVIRONMENT } = config();
+const posthogClient = POSTHOG_API_KEY
+  ? new PostHog(POSTHOG_API_KEY, { host: POSTHOG_API_HOST })
+  : undefined;
+
+function idFromCookie(request: Request) {
+  // Note: can't use cookie.parse(): https://github.com/remix-run/remix/discussions/5198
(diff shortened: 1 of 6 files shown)


app/services/analytics/customEvent.ts
```diff
@@ -0,0 +1,37 @@
+import { PostHog } from "posthog-node";
+import { hasTrackingConsent } from "./gdprCookie.server";
+import { config } from "../env/web";
+import { parse } from "cookie";
+
+const { POSTHOG_API_KEY, POSTHOG_API_HOST, ENVIRONMENT } = config();
+const posthogClient = POSTHOG_API_KEY
```
frederike-ramin, line 7: Should we have multiple instances of this? Feels like one could be enough?

app/services/analytics/customEvent.ts
```diff
@@ -0,0 +1,37 @@
...
+  ? new PostHog(POSTHOG_API_KEY, { host: POSTHOG_API_HOST })
+  : undefined;
+
+function idFromCookie(request: Request) {
+  // Note: can't use cookie.parse(): https://github.com/remix-run/remix/discussions/5198
+  // Returns ENVIRONMENT if posthog's distinct_id can't be extracted
```
frederike-ramin, line 13: Just curious - why are we returning ENVIRONMENT and not something like "not given"?

frederike-ramin (commented):
Looks fine to me. We should really see if we really want to do this on *every* page submit though.

app/services/analytics/customEvent.ts
chohner: Great idea to have one implementation for all posthog interactions (although I would hope this init call isn't too expensive)

PR #518, opened 2024-01-11
feat: move to new CI flow

chohner:
- using new Dockerfile as standard (with split app & content)
- using new CI flow as standard
- removing unused github actions


.github/workflows/build-content.yml
```diff
@@ -6,25 +6,148 @@ env:
...
+          argocd_server: ${{ secrets.ARGOCD_SERVER }}
+          argocd_sync_timeout: 300
+
+  test-production-text:
+    needs: [deploy-production]
+    permissions:
+      packages: read
+    runs-on: ubuntu-latest
+    steps:
+      - run: curl -s "https://service.justiz.de" | grep -q "Justiz-Services"
```
pgurusinga, line 153: @chohner I tried this one-liner seems won't fail if we search for a non-existing string, or am I missing something?
please see screenshot

image

chohner: It does for me 🤔  
image

PR #565, opened 2024-01-31
WIP: testing action with ref to app image

pgurusinga:


pgurusinga:
FYI. I will merge this because of the kyverno config in the Platform points to main branch, if any issue occurs I will revert this

PR #606, opened 2024-02-16
refactor(stepPages): split vorabacheck & antrag

chohner:


app/components/MigrationDataOverview.tsx
```diff
@@ -5,12 +5,13 @@ import Container from "./Container";
...
 type MigrationDataProps = {
-  readonly migrationData: Record<string, unknown>;
+  readonly migrationData?: Record<string, unknown>;
   readonly translations: Translations;
 };
 
 export default function MigrationDataOverview(props: MigrationDataProps) {
-  if (Object.keys(props.migrationData).length == 0) return false;
+  if (!props.migrationData || Object.keys(props.migrationData).length == 0)
+    return null;
```
pgurusinga, line 14: I noticed that we're using `null` in this code instead of Fragment `<></>`< like we do in other parts of the code. Would it be possible to align and use React Fragment here as well?

chohner: good catch, something I also wanted to discuss. I was also wondering about what the right return value is any found [this in the docs](https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null)

pgurusinga (approved):

SannyNguyenHung (approved):

app/routes/shared/vorabcheck.server.ts
```diff
@@ -0,0 +1,190 @@
...
+  await throw404IfFeatureFlagEnabled(request);
+
+  // get data from request
+  const { pathname } = new URL(request.url);
+  const { flowId, stepId } = parsePathname(pathname);
+  const cookieId = request.headers.get("Cookie");
+
+  // get data from redis
+  const { data, id } = await getSessionForContext(flowId).getSession(cookieId);
+  const userDataFromRedis: Context = data; // Recast for now to get type safety
```
rbrtrfl, line 46: nit(non-blocking): i'd be consistent with the naming, either `flowContext` or `userDataFromRedis`

app/routes/shared/antrag.server.ts
```diff
@@ -198,40 +174,50 @@ export const loader = async ({
...
-    commonContent,
-    nextButtonLabel: cmsContent.nextButtonLabel,
-    isFinal: flowController.isFinal(stepId),
-    configMetadata: flowController.getMeta(stepId),
-    previousStepUrl: flowController.getPrevious(stepId)?.url,
-    returnTo,
-  });
+  const buttonNavigationProps = {
+    next: flowController.isFinal(stepId)
+      ? undefined
```
rbrtrfl, line 179: question: müssen wir den case im Vorabcheck nicht beachten?

app/routes/shared/antrag.server.ts
```diff
@@ -198,40 +174,50 @@ export const loader = async ({
...
-      translations,
-      progress,
-      templateReplacements,
-      buttonNavigationProps,
       navItems,
+      postFormContent: cmsContent.postFormContent,
+      preHeading: cmsContent.preHeading,
       returnTo,
+      stepData,
+      translations: flowStrings,
```
rbrtrfl, line 220: praise: alphabetisch 😍

app/routes/shared/antrag.server.ts
```diff
@@ -119,64 +105,54 @@ export const loader = async ({
...
+    navigationStrings,
+    defaultStrings,
   ] = await Promise.all([
-    fetchSingleEntry("vorab-check-common"), // TODO replace with translations
-    fetchCollectionEntry(currentFlow.cmsSlug, lookupPath),
+    fetchCollectionEntry("form-flow-pages", lookupPath),
     fetchMeta({ filterValue: parentFromParams(pathname, params) }),
     fetchTranslations(flowId),
     fetchTranslations(`${flowId}/menu`),
+    fetchTranslations("vorabcheck"), // TODO: rename to defaultTranslations in strapi?
```
rbrtrfl, line 125: question: sollen die 'translations' von sowohl Vorabcheck als auch Antrag beide in `defaultTranslations` in Strapi liegen?

rbrtrfl (approved):
It feels a bit strange to have so much almost identical code twice. But it's always clearer!

app/routes/shared/antrag.server.ts
chohner: imo: yes

app/models/flows/beratungshilfeFormular/index.ts
```diff
@@ -142,6 +142,7 @@ export const beratungshilfeAntrag = {
   stringReplacements: (context: AllContexts) => ({
     ...getAmtsgerichtStrings(context),
     ...getStaatlicheLeistungenStrings(context),
+    // TODO: is this needed?
```
frederike-ramin, line 145: Yes. It's for showing or hiding a hint at the end of the form that you should find yourself an Anwalt.

app/routes/shared/antrag.server.ts
```diff
@@ -77,22 +75,10 @@ function stepDataFromFieldNames(
...
-    ogTitle: pageMeta.ogTitle ?? parentMeta?.ogTitle,
-    title: `${pageMeta.title} - ${parentMeta?.title ?? ""}`,
-  };
-}
-
 export const loader = async ({
   params,
   request,
   context,
-  // eslint-disable-next-line sonarjs/cognitive-complexity
```
frederike-ramin, line 95: 🙏

app/routes/shared/antrag.server.ts
```diff
@@ -198,40 +174,50 @@ export const loader = async ({
   const headers = { "Set-Cookie": await sessionContext.commitSession(session) };
 
   // get navigation destinations + labels
-  const buttonNavigationProps = getButtonNavigationProps({
```
frederike-ramin, line 201: Why not keep a function for this? Now it's just more cognitive load within the loader?

app/routes/shared/components/AntragPage.tsx
```diff
@@ -6,68 +6,50 @@ import Background from "~/components/Background";
...
 import Heading from "~/components/Heading";
 import MigrationDataOverview from "~/components/MigrationDataOverview";
 import FlowNavigation from "~/components/FlowNavigation";
 import { splatFromParams } from "~/services/params";
-import type { loader } from "../step.server";
+import type { loader } from "../antrag.server";
 import ArraySummary from "~/components/ArraySummary";
 
-export function StepWithPreHeading() {
+export function AntragPage() {
```
frederike-ramin, line 16: Can we rename the common components? So this e.g. is not the AntragPage, I think it should be FormFlowPage? And as Sanny mentioned in the Daily: let's rename antrag.server to formular.server?

frederike-ramin (commented):
- Love that we're filling the templates in the loader now. 👌 
- Thanks for removing `buttonNavigationProps`

In general - nice idea. I'll be honest though - this is not the change I wanted to see with our "reduce loader + action complexity" task. In my head, it's step 2 that we're now doing before the step 1 I had in mind - creating functions for the different tasks a loader has. Because what we now did, is duplicating so much code that I would have liked simplified before separating the two site types. I think you went a few steps in the right direction but I would have loved to see more loader tasks divided into functions, which I still think we should do, we now just have to do it twice.

So - yes, it's a good change and let's merge it, but let's not stop there, but instead make the loaders and actions more readable. 🙏

app/routes/shared/antrag.server.ts
SannyNguyenHung: We will have a look at this in the next iteration

app/routes/shared/components/ResultPage.tsx
```diff
@@ -61,11 +60,7 @@ export function Result() {
     <>
       <div className={backgrounds[cmsData.pageType]}>
         <Container paddingTop="24">
-          <ProgressBar
-            label={common.progressBarLabel}
-            progress={progress.current}
-            max={progress.total}
-          />
+          <ProgressBar label={common.progressBarLabel} progress={1} max={1} />
```
frederike-ramin, line 63: I get that it is visually the same, but I think in a accessibility-sense this might be confusing, right? Because it is spoken out like "Step 4/6", "Step 5/6", "Step 1/1"?

SannyNguyenHung: Nice hint!! Your right - we need to keep accessibility here

PR #612, opened 2024-02-19
feat(posthog): add custom event for cookie consent and unify event handling

rbrtrfl:
Actually I just wanted to quickly add the event to the cookie-consent, then I tried to standardize the event handling. Feedback appreciated 🙏😁

### DIFF
--- a/app/routes/shared/step.server.ts
+++ b/app/routes/shared/step.server.ts
@@ -37,6 +37,7 @@ import {
   deleteFromArrayInplace,
 } from "~/services/session.server/arrayDeletion";
 import type { StrapiMeta } from "~/services/cms/models/StrapiMeta";
+import { hasTrackingConsent } from "~/services/analytics/gdprCookie.server";
 
 const structureCmsContent = (
   formPageContent: z.infer<
@@ -301,8 +302,12 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   });
 
   const customEventName = flowController.getMeta(stepId)?.customEventName;
-  if (customEventName)
-    void sendCustomEvent(customEventName, validationResult.data, request);
+  if (customEventName && (await hasTrackingConsent({ request })))
+    sendCustomEvent({
+      eventName: customEventName,
+      context: validationResult.data,
+      request,
+    });
 
   const returnTo = formData.get("_returnTo");
   const destination =
--- a/app/routes/shared/step.server.ts
+++ b/app/routes/shared/step.server.ts
@@ -304,9 +304,9 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   const customEventName = flowController.getMeta(stepId)?.customEventName;
   if (customEventName && (await hasTrackingConsent({ request })))
     sendCustomEvent({
-      eventName: customEventName,
-      context: validationResult.data,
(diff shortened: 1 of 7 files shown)


app/services/analytics/gdprCookie.server.ts
```diff
@@ -37,21 +38,34 @@ async function createTrackingCookie({
...
+    updated: cookieUpdated,
+  };
 }
 
 export async function consentCookieFromRequest({
   request,
 }: {
   request: Request;
-}) {
+}): Promise<Record<string, string | never>> {
```
chohner, line 63: why not leave this to typescript?

app/services/analytics/gdprCookie.server.ts
```diff
@@ -37,21 +38,34 @@ async function createTrackingCookie({
...
   const formData = await request.formData();
   const fieldContent = formData.get(acceptCookiesFieldName);
   let consent = undefined;
   if (fieldContent === "true") consent = true;
   else if (fieldContent === "false") consent = false;
-  return createTrackingCookie({ request, consent });
+  const { cookie, updated } = await createTrackingCookie({ request, consent });
+  return updated ? { "Set-Cookie": cookie } : {};
```
chohner, line 70: I'm not sure why we handle the updated case but shouldn't this logic move into `createTrackingCookie`?

app/services/analytics/customEvent.ts
```diff
@@ -14,20 +14,25 @@ function idFromCookie(request: Request) {
...
+export function sendCustomEvent({
+  request,
+  eventName,
+  context,
+  properties,
+}: {
+  request: Request;
+  eventName: string;
+  context?: Context;
+  properties?: Record<string, string | boolean | undefined>;
```
chohner, line 26: This context isn't a flow context but exactly what `properties` is doing here (enriching the custom event). so we only need one of the two

app/routes/shared/step.server.ts
```diff
@@ -301,8 +302,12 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   });
 
   const customEventName = flowController.getMeta(stepId)?.customEventName;
-  if (customEventName)
-    void sendCustomEvent(customEventName, validationResult.data, request);
+  if (customEventName && (await hasTrackingConsent({ request })))
```
chohner, line 305: afaik we specifically do not need `hasTrackingConsent` here since we are sending a server-side event (which doesn't know about users that haven't consented)

app/services/analytics/customEvent.ts
```diff
@@ -14,20 +14,25 @@ function idFromCookie(request: Request) {
...
-) {
-  if (!(await hasTrackingConsent({ request }))) return;
-
+export function sendCustomEvent({
+  request,
+  eventName,
+  context,
+  properties,
+}: {
+  request: Request;
```
pgurusinga, line 23: we could probably pass just `request.url` rather the whole request object

app/routes/shared/step.server.ts
rbrtrfl: I just moved this condition from app/services/analytics/customEvent.ts 🧐

PR #651, opened 2024-03-07
Array — Kinder

pgurusinga:


pgurusinga:
there are 2 options to keep the commit message:

- Create a merge commit, **or**
image
- Merge and squash, and we will still have the reference in the main,
better is first option, where we can revert easily in the `main` branch

PR #654, opened 2024-03-11
Pdf refactoring

SannyNguyenHung:
Refactoring:

- Move attachment / description field into a own file
- Split pdf file into multiple sections
- Create tests for each section
- Create export for kids


app/services/pdf/beratungshilfe/beratungshilfe.pdf.ts
```diff
@@ -0,0 +1,62 @@
+import type { BeratungshilfeFormularContext } from "~/models/flows/beratungshilfeFormular";
+import { checkboxListToString } from "../checkboxListToString";
+
+export const staatlicheLeistungMapping = {
```
pgurusinga, line 4: honestly, if I didn't know the PDF context, I would not know this variable for. is it for the Attachment? Is it `staatlicheLeistungAttachmentMapping` explanatory enough?

app/services/pdf/beratungshilfe/beratungshilfe.pdf.ts
```diff
@@ -0,0 +1,62 @@
...
+  grundsicherung: "Grundsicherung",
+  asylbewerberleistungen: "Asylbewerberleistungen",
+  buergergeld: "Bürgergeld",
+  andereLeistung: "Andere Leistung",
+  keine: "Keine",
+};
+
+export const getOccupationDetails = (
+  context: BeratungshilfeFormularContext,
+  withAdditionalIncome = true,
```
pgurusinga, line 14: why do we set the default argument to `true` if later on we need to conditionally check it in line 46 which I think, it does not make sense? or am I missing something? 🙏 

```js
...
if (context.weitereseinkommen && withAdditionalIncome) {
...
```

app/services/pdf/beratungshilfe/sections/E_unterhalt.ts
```diff
@@ -0,0 +1,144 @@
...
+export function fillUnterhalt(
+  descriptionField: DescriptionField,
+  pdfFields: BeratungshilfePDF,
+  context: BeratungshilfeFormularContext,
+) {
+  const unterhaltPdfFields: UnterhaltPdfFields[] = [];
+
+  if (context.partnerschaft === "yes" && context.unterhalt === "yes") {
+    unterhaltPdfFields.push({
+      name: [context.partnerVorname ?? "", context.partnerNachname ?? ""].join(
```
pgurusinga, line 22: very interesting, if `partnerschaft` exists, under what condition it can be an empty string? user must enter partner information such as `vorname` and `nachname` right?

app/services/pdf/beratungshilfe/sections/E_unterhalt.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  descriptionField.descriptions.unshift({
+    title: "Unterhalt",
+    text: description.join("\n\n"),
+  });
+
+  const person1 = unterhaltPdfFields.pop();
+  if (!person1) return;
+  fillPerson1(pdfFields, person1);
```
pgurusinga, line 68: or is it possible to do it other way?
```js
if (person1) fillPerson1(pdfFields, person);
```

app/services/pdf/beratungshilfe/beratungshilfe.pdf.ts
pgurusinga: ah nevermind. I got it now

app/services/pdf/beratungshilfe/sections/E_unterhalt.ts
SannyNguyenHung: Yes, but typescript doesn't know that. That's why partnerFirstName and partnerLastName are optional in the context

app/services/pdf/beratungshilfe/sections/E_unterhalt.ts
SannyNguyenHung: I would like to cancel as soon as there is no more element in the list:

```
const person1 = unterhaltPdfFields.pop();
if (person1) { fillPerson1(pdfFields, person) }
else return;

const person2 = unterhaltPdfFields.pop();
if (person2) { fillPerson2(pdfFields, person) }
else return; 

...
```

I don't know how to write it better, because these functions (fillPerson1, fillPerson2 ...) are derived from the PDF and even if they actually do the same thing, they have different field IDs

app/services/pdf/beratungshilfe/beratungshilfe.pdf.ts
SannyNguyenHung: I will move it to the header section and remove the whole file :)

app/services/pdf/beratungshilfe/sections/E_unterhalt.ts
pgurusinga: probably we can leave it as it is now and we can improve in chunks later 👍

app/services/pdf/beratungshilfe/sections/E_unterhalt.ts
```diff
@@ -0,0 +1,145 @@
...
+function fillPerson4(
+  pdfFields: BeratungshilfePDF,
+  unterhaltPdfFields: UnterhaltPdfFields,
+) {
+  pdfFields.e1Person4.value = unterhaltPdfFields.name;
+  pdfFields.e3Familienverhaeltnis4.value =
+    unterhaltPdfFields.familienverhaeltnis;
+  pdfFields.e4Zahlung4.value = unterhaltPdfFields.unterhaltSumme;
+  pdfFields.e5Einnahmen4.value = unterhaltPdfFields.hatEinnahmen;
+}
```
pgurusinga, line 126: I'm not quite sure if this is doable, but I've been thinking about a possible plan to have these as single-function. but this is not blocker

```ts
function fillPerson(
  pdfFields: BeratungshilfePDF,
  unterhaltPdfFields: UnterhaltPdfFields,
  personNumber: number
) {
  const personPrefix = `e1Person${personNumber}`;
  pdfFields[personPrefix].value = unterhaltPdfFields.name;
  pdfFields[`e3Familienverhaeltnis${personNumber}`].value =
    unterhaltPdfFields.familienverhaeltnis;
  pdfFields[`e4Zahlung${personNumber}`].value = unterhaltPdfFields.unterhaltSumme;
  pdfFields[`e5Einnahmen${personNumber}`].value = unterhaltPdfFields.hatEinnahmen;
}
```

SannyNguyenHung: Is this more readable? and there is a twist: for person1 we need "pdfFields.e3Familienverhaeltnis" without the person number on the end

PR #663, opened 2024-03-14
Marital

aaschlote:
- Add two new options for the partner marital options (separated and widowed)
- Add marital status in the pdf generator
- Add unit tests
- Add note when runs E2E

### DIFF
--- a/app/services/pdf/beratungshilfe/attachment.ts
+++ b/app/services/pdf/beratungshilfe/attachment.ts
@@ -1,10 +1,16 @@
 import type { BeratungshilfeFormularContext } from "~/models/flows/beratungshilfeFormular";
+import { getMaritalDescriptionByContext } from "./sections/header";
+
+const MAXIMUM_LENGTH_MARITAL_DESCRIPTION = 10;
+export const MARTIAL_STATUS_TITLE = "Familienstand:";
 
 export function createAttachment(
   context: BeratungshilfeFormularContext,
   maxLength = 255,
 ) {
   const descriptions: Attachment["descriptions"] = [];
+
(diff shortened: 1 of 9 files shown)


app/services/pdf/beratungshilfe/sections/header.ts
```diff
@@ -52,6 +54,23 @@ export default function fillHeader(
...
+    case "no":
+      return "ledig";
+    case "separated":
+      return "getrennt";
+    case "widowed":
+      return "verwitwet";
+    default:
+      return "";
+  }
+};
```
chohner, line 72: Not a blocker: I have a slightly preference for a "simple" lookup in an enum -> string object

app/services/pdf/beratungshilfe/attachment.ts
```diff
@@ -1,11 +1,17 @@
 import type { BeratungshilfeFormularContext } from "~/models/flows/beratungshilfeFormular";
+import { getMaritalDescriptionByContext } from "./sections/header";
+
+const MAXIMUM_LENGTH_MARITAL_DESCRIPTION = 10;
+export const MARTIAL_STATUS_TITLE = "Familienstand:";
```
chohner, line 5: Not a blocker: Could this somehow go closer to/in the corresponding code? I don't feel like this is a global

chohner (approved):

app/services/pdf/beratungshilfe/attachment.ts
aaschlote: as we discuss, export it for the unit tests

PR #682, opened 2024-03-20
Add testcases for beratungshilfe antrag

pgurusinga:
## Problem:
we don't have testcases for flow logic

## Proposed changes:
- add testcases but not deep into the subflow (still can be discussed


tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcases.ts
```diff
@@ -0,0 +1,139 @@
...
+import { beratungshilfeFormular } from "~/models/flows/beratungshilfeFormular";
+import type { TestCases } from "../../TestCases";
+import type { FlowStateMachine } from "~/services/flow/server/buildFlowController";
+import type { BeratungshilfeFinanzielleAngaben } from "~/models/flows/beratungshilfeFormular/finanzielleAngaben/context";
+
+const machine: FlowStateMachine = createMachine(beratungshilfeFormular.config, {
+  guards: beratungshilfeFormular.guards,
+});
+
+const FINANZIELLE_ANGABEN = "finanzielleAngaben";
```
pgurusinga, line 11: to satisfy solarlint (repetitive text) 😓

tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcases.ts
```diff
@@ -0,0 +1,139 @@
...
+    [
+      `${FINANZIELLE_ANGABEN}/besitz/gesamtwert`,
+      `${FINANZIELLE_ANGABEN}/besitzZusammenfassung/zusammenfassung`,
+    ],
+  ],
+  [
+    {},
+    [
+      `${FINANZIELLE_ANGABEN}/besitzZusammenfassung/zusammenfassung`,
+      `${FINANZIELLE_ANGABEN}/danke`,
```
frederike-ramin, line 130: I feel like this could just be one bigger test case?

tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcases.ts
```diff
@@ -0,0 +1,139 @@
...
+      `${FINANZIELLE_ANGABEN}/kinder/kinder-frage`,
+    ],
+  ],
+  [
+    { partnerschaft: "yes" },
+    [
+      `${FINANZIELLE_ANGABEN}/partner/partnerschaft`,
+      `${FINANZIELLE_ANGABEN}/partner/zusammenleben`,
+    ],
+  ],
```
frederike-ramin, line 69: I believe partnerschaft also has later implications in the flow? the page `heirat-info` comes to mind.

frederike-ramin (commented):
Nice, thank you for taking case of this! 
Could you also add test cases for the kinder subflow? (So once you clicken on "Kind hinzufügen"? I believe that has already grown quite complex too?

tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcases.ts
aaschlote: add it

tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcases.ts
aaschlote: We didn't get what do you mean. Can you explain again please?

tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcasesBesitz.ts
```diff
@@ -0,0 +1,33 @@
...
+import { beratungshilfeFormular } from "~/models/flows/beratungshilfeFormular";
+import type { TestCases } from "../../TestCases";
+import type { FlowStateMachine } from "~/services/flow/server/buildFlowController";
+import type { BeratungshilfeFinanzielleAngaben } from "~/models/flows/beratungshilfeFormular/finanzielleAngaben/context";
+
+const machine: FlowStateMachine = createMachine(beratungshilfeFormular.config, {
+  guards: beratungshilfeFormular.guards,
+});
+
+const FINANZIELLE_ANGABEN = "finanzielleAngaben";
```
chohner, line 11: I was surprised this comes up because we specifically disable this lint rule on test files.

However I just checked https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/ac51d3666d043cd12757564db59352aefdefaaf3/.eslintrc.cjs#L43 and we match against `"**/*.test.ts"` (which this file isn't). Should we instead match against `/tests/**/*.ts`?

tests/unit/models/flows/beratungshilfeFormular/finanzielleAngaben/testcasesBesitz.ts
```diff
@@ -0,0 +1,33 @@
+import { createMachine } from "xstate";
+import { beratungshilfeFormular } from "~/models/flows/beratungshilfeFormular";
+import type { TestCases } from "../../TestCases";
+import type { FlowStateMachine } from "~/services/flow/server/buildFlowController";
+import type { BeratungshilfeFinanzielleAngaben } from "~/models/flows/beratungshilfeFormular/finanzielleAngaben/context";
+
+const machine: FlowStateMachine = createMachine(beratungshilfeFormular.config, {
+  guards: beratungshilfeFormular.guards,
+});
```
chohner, line 9: since we're re-declaring the same machine many times - maybe it makes sense to create it once in `/test/unit/models/flows/beratungshilfeFormular/` and just keep the specific testcases in each file?

chohner (approved):

frederike-ramin (approved):
Thanks for taking care of the comments! 🙏

PR #692, opened 2024-03-25
Iterate Grundeigentum

chohner:
To reach multiple array subflows, we used to configure multiple targets for a virtual `ADDITEM` event for the overview step:
```
  "ADDITEM": {
    "target": [
      "bankkonten",
      "wertgegenstaende",
      ...
    ]
  }
```

Even though this shouldn't be valid state machine configuration (as only the first valid destination should be reachable), this seemed to work for the two cases we had: a single array with multiple steps and multiple arrays with a single step each.

However, now multiple arrays with multiple steps should be reachable. Now each array declares its own `event`, which is added to the machines meta:

```
meta: {
    arrays: {
        bankkonten: {
            event: add-bankkonten,
            ...
        }
    }
}
```

These custom array events can be used as transitions from overview page (`"add-bankkonten": "bankkonten") as well as passed to `xstate` when finding all reachable steps.

This PR then uses this new configuration to update the Grundeigentum flow (inkluding pdf and tests)


app/services/flow/server/buildFlowController.ts
```diff
@@ -48,14 +46,22 @@ type Meta = {
   done: (context: Context) => boolean | undefined;
   subflowState: (context: Context, subflowId: string) => SubflowState;
   subflowDone: (context: Context, subflowId: string) => boolean | undefined;
+  arrays?: Record<string, ArrayConfig>;
```
aaschlote, line 49: can we rename it? I feel a field called ``arrays`` doesn't give so much clear what it has as values. Maybe ``arrayConfigurationsFlow``?

chohner: I feel like the location is already quite specific (its already inside the array configuration). But maybe `arrayConfiguration` if you think that is more clear?

frederike-ramin:
I don't think I understand the problem yet. 🤔 Why would the old solution not work with multiple arrays with multiple steps? There's always just one first step for each sub-flow, no matter how many come afterwards?

chohner:
I agree, the issue is not apparent from code/configuration.

However, my assumption is that we were relying on a bug or undocumented feature, since specifying multiple transition destinations are by definition not reachable by a (non-parallel) state machine (see https://xstate.js.org/docs/guides/transitions.html#multiple-targets).

This PR replaces the generic (virtual) `ADDITEM` event with a unique event for each array.

app/services/flow/server/buildFlowController.ts
aaschlote: ``arrayConfiguration`` would be fine by me :)

frederike-ramin:
Ah okay, I get it. So `getShortestPaths` probably only goes deeper into a branch if it is the first in the list. Bummer. But yes, relying on unexpected behaviour is probably a bad idea in general as you pointed out 👍

app/models/flows/beratungshilfeFormular/finanzielleAngaben/arrayConfiguration.ts
```diff
@@ -0,0 +1,52 @@
+import type { ArrayConfig } from "~/services/array";
+import type { BeratungshilfeFinanzielleAngaben } from "./context";
+
+const prefix = "/beratungshilfe/antrag/finanzielleAngaben";
+
+export const finanzielleAngabenArrayConfig = {
+  bankkonten: {
+    url: `${prefix}/besitzZusammenfassung/bankkonten`,
+    initialInputUrl: "daten",
```
frederike-ramin, line 9: I would have imagined this to come only from the flow config (aka xstate)?

app/models/flows/beratungshilfeFormular/finanzielleAngaben/arrayConfiguration.ts
```diff
@@ -0,0 +1,52 @@
...
+    questionUrl: `${prefix}/besitz/bankkonten-frage`,
```
frederike-ramin, line 10: Why did we decide to make this a special case? I would have imagined this to be defined in Strapi as it is really just alink?

frederike-ramin (approved):
Thanks for finding and fixing this. `dependabot` would be lucky to have you as a godfather 😘 
I think I only commented a few things that have nothing to do with this PR, still wanted to point them out though^^

app/models/flows/beratungshilfeFormular/finanzielleAngaben/arrayConfiguration.ts
chohner: +1, I think we might need to rethink what belongs in the `ArraySummary` component and what in the configuration (stuff like title & description).

app/models/flows/beratungshilfeFormular/finanzielleAngaben/arrayConfiguration.ts
chohner: I agree, but I don't think there is a straightforward way where this can be evaluated (yet), as this has to happen in the loader 🤔

PR #698, opened 2024-03-26
Feat: Rechtsproblem iteration

pgurusinga:
### Problem:
- Multiple steps for writing the rechtsproblem

### Proposed changes:
- Extend `TextArea` to receive description property
- Label is styled with h3 (text area should have a label for a11y)
- Update the Rechtsproblem flow
- Update data model (context)
- Unit tests
- E2E tests
- Textarea component test
- Context to PDF adjustment


app/models/flows/beratungshilfeFormular/rechtsproblem/flow.json
```diff
@@ -4,46 +4,16 @@
   "states": {
     "start": { "on": { "SUBMIT": "bereich" } },
     "bereich": {
-      "on": { "SUBMIT": "beschreibung", "BACK": "start" }
+      "on": { "SUBMIT": "beschreibungssituationen", "BACK": "start" }
```
frederike-ramin, line 7: Can we rename this to `situation-beschreibung`? I don't think that german word `beschreibungssituationen` exists. :)

app/services/pdf/beratungshilfe/attachment.ts
```diff
@@ -87,7 +94,7 @@ const addMaritalToAttachment = ({
 
   if (maritalDescription.length > MAXIMUM_LENGTH_MARITAL_DESCRIPTION) {
     descriptions.push({
-      title: MARTIAL_STATUS_TITLE,
```
frederike-ramin, line 90: Hihi what a funny typo this was. :D 🥋

tests/unit/services/pdf/beratungshilfe/attachment.test.ts
```diff
@@ -24,8 +29,64 @@ describe("createAttachment", () => {
 
     const attachment = createAttachment(context);
     const hasMaritalDescription = attachment.descriptions.some(
-      (description) => description.title === MARTIAL_STATUS_TITLE,
+      (description) => description.title === MARITIAL_STATUS_TITLE,
     );
     expect(hasMaritalDescription).toBeFalsy();
   });
+
+  it("should add bereich description in the attachment", () => {
```
frederike-ramin, line 37: Do I see it correctly that you only check if the *titles* are there, not the actual text? Maybe that's also worth an `expect()`?

frederike-ramin: Ah, interesting, I just saw that this is a different test? 🤔 Could you real quick explain why it's different? I haven't worked with the PDF at all yet and would be interested. :)

frederike-ramin (approved):
Nice, great work. You checked with Paul because of the new margin between label and textfield? 
Also: Bonus points if you add a storybook story for the new textarea configuration ✨

tests/unit/services/pdf/beratungshilfe/attachment.test.ts
pgurusinga: There is no specific reason to test only the title property; it was a decision made based on what we think gives us sufficient confidence. However, if you believe that the title is equally important, I can add it to this test.

pgurusinga:
> You checked with Paul because of the new margin between label and textfield?

@frederike-ramin yes, Paul was informed and is okay with that

tests/unit/services/pdf/beratungshilfe/attachment.test.ts
frederike-ramin: I believe you *are* already testing it in `tests/unit/services/pdf/beratungshilfe/sections/A_angelegenheit.test.ts`? Just wondering why we test a similar thing (the description and the value of the same thin) in two different test files.

aaschlote: I believe the test ``tests/unit/services/pdf/beratungshilfe/sections/A_angelegenheit.test.ts`` is checking what has been added to the ``attachment`` is saved into ``pdfFields``. In this test (``attachment``) it's checked what it added into the attachment and it doesn't care what into the ``pdfFields``. Even there are very similars, they are checking different things. And this is the idea of the unit test, checking the intput and output for a single unit (in this case a function) :)

pgurusinga: ah ok. those two are different things but look similar, one test is for testing whether `A_angelegenheit` field would get correct data and the other is for attachment data

pgurusinga: @aaschlote you are faster 🚗 💨

aaschlote: image

frederike-ramin (commented):
Love the tests 🙏 Can you maybe just add one more where you trigger a TAG event and type into the field that then the value of the field is set accordingly? Like, really testing the interaction with the component, not just the look.

PR #709, opened 2024-04-03
Feat: add new subflow "Andere Unterhaltszahlungen"

aaschlote:


app/models/flows/beratungshilfeFormular/finanzielleAngaben/context.ts
```diff
@@ -173,6 +201,8 @@ export const beratungshilfeFinanzielleAngaben = {
   apartmentCostOwnShare: buildMoneyValidationSchema(),
   apartmentCostFull: buildMoneyValidationSchema(),
   apartmentCostAlone: buildMoneyValidationSchema(),
+  hasOtherMaintenancePayments: YesNoAnswer,
```
frederike-ramin, line 204: I know it's very annoying with the english-german mix, but can you rename this variable to `hasWeitereUnterhaltszahlungen`?

PR #710, opened 2024-04-03
Remove no longer necessary cosign config

carhartl:
See https://digitalservicebund.slack.com/archives/C01JUUDHG5Q/p1712149498432839


aaschlote (approved):
Thanks for doing the changes

PR #713, opened 2024-04-04
Proposal: simplify navigation logic

chohner:
Issues:
 - mixing of view logic & flow logic
 - `done()` function of child states are called multiple times
 - hardcoded two levels of nesting
 - complex `navItemsFromFlowSpecifics()` and `getSubflowSpecifics()`

Proposal:
- split into logic layer (inside flow controller, only checks done and adds paths) and view layer (`flowNavigation.server`, adds current, labels etc)
- use recursion to traverse nested states (until `done()` function or no more children)
- each `done()` is called only once: parent state fully depends on their child state (ie current if any child is current, done if all children are done)
- removal of specialized `subflowState` and `subflowDone` functions in meta

Todo:
- [x] unit tests


app/services/flowNavigation.server.ts
```diff
@@ -1,106 +1,40 @@
...
 import { type Translations } from "./cms/index.server";
 
-export function navItemsFromFlowSpecifics(
-  currentStepId: string,
-  flowController: ReturnType<typeof buildFlowController>,
-  translation: Translations = {},
-): NavItem[] {
-  const currentFlow = flowController.getConfig();
-  const flowId = currentFlow.id ?? "";
+function isDoneStateIdCurrent(doneStateId: string, stepId: string) {
```
pgurusinga, line 5: is this function checking whether any stepId has a done state? would this name `isStepInDoneState` clearer? I was a bit confused at first

app/models/flows/beratungshilfeFormular/index.ts
```diff
@@ -57,15 +65,25 @@ export const beratungshilfeFormular = {
...
+          einkommen: { meta: { done: einkommenDone } },
+          partner: { meta: { done: partnerDone } },
+          kinder: { meta: { done: kinderDone } },
+          "andere-unterhaltszahlungen": {
+            meta: { done: andereUnterhaltszahlungenDone },
+          },
+          besitz: { meta: { done: besitzDone } },
+          besitzZusammenfassung: { meta: { done: besitzZusammenfassungDone } },
+          wohnung: { meta: { done: wohnungDone } },
+          ausgaben: { meta: { done: ausgabenDone } },
```
pgurusinga, line 78: Thanks for the refactoring, does the navigation state update only after finishing the flow? what would happen for a skippable or optional flow such as `ausgaben` flow? we probably need a kind of "visited" state where the UI is updated without going through the flow...

tests/unit/services/flow/buildFlowController.test.ts
```diff
@@ -291,4 +291,143 @@ describe("buildFlowController", () => {
...
+                initial: "start",
+                states: { start: {} },
+              },
+            },
+          },
+        },
+      },
+    }).doneStates();
+    expect(doneStates[0].isReachable).toBe(true);
+    expect(doneStates[1].isReachable).toBe(false);
```
frederike-ramin, line 431: Not sure this is really testing what you're saying in the description? parent1 is reachable just because it's the initial state. That might be something else later in the flow? so perhaps we need a parent0 in the test that makes parent1 reachable, but not parent2?

tests/unit/services/flow/buildFlowController.test.ts
```diff
@@ -291,4 +291,143 @@ describe("buildFlowController", () => {
       });
     });
   });
+
+  describe(".doneStates()", () => {
```
frederike-ramin, line 295: I would love a very simple test without nesting, just checking that it works for the simplest case?

tests/unit/services/flowNavigation.test.ts
```diff
@@ -28,80 +27,72 @@ describe("flowNavigation", () => {
...
-      const flowController = buildFlowController({
-        config: _.merge(_.cloneDeep(config), {
-          states: { step1: { meta: { done: () => true } } },
-        }),
-      });
+    const doneStatesNested = [
+      { ...parentDoneState, subStates: [childDoneState] },
+    ];
 
+    it("IsCurrent as child", () => {
```
frederike-ramin, line 50: I'm confused. What is this test for? 🤔

tests/unit/services/flowNavigation.test.ts
```diff
@@ -28,80 +27,72 @@ describe("flowNavigation", () => {
...
     });
 
-    describe("works with isUneditable", () => {
-      const flowController = buildFlowController({
-        config: _.merge(_.cloneDeep(config), {
-          states: { step1: { meta: { isUneditable: true, done: () => true } } },
-        }),
-      });
-
+    it("IsCurrent is set for parent", () => {
```
frederike-ramin, line 63: I think a negative test makes sense here? So testing for when isCurrent is *not* set.

frederike-ramin (commented):
Love the initiative! I think you're right, we're missing some unit tests and I think the naming can be improved a little bit, but it is so much more logical and clear to me than the last version 🙏

app/services/flow/server/buildFlowController.ts
```diff
@@ -127,6 +124,60 @@ function getInitial(machine: FlowStateMachine) {
...
+    stateNode.states ?? {},
+  ).filter((state) => state.meta?.done || Object.keys(state.states).length > 0);
+
+  return statesWithDoneFunctionOrSubstates.map((state) => {
+    const stepId = stateValueToStepIds(pathToStateValue(state.path))[0];
+    const meta = state.meta as Meta | undefined;
+    const isUneditable = Boolean(meta?.isUneditable);
+    const subDoneStates = doneStates(state, reachableSteps);
+
+    // Ignore subflows if empty or parent state has done function
```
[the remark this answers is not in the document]
chohner: Yes, thats exactly whats happening - its stopping recursing at the first done function

PR #750, opened 2024-04-16
refactor: Add description on Result page based distance calculation

aaschlote:


app/util/calculateDistanceBetweenAirports.ts
```diff
@@ -0,0 +1,37 @@
+import airports from "data/airports/data.json";
+import haversine from "haversine-distance";
+import { z } from "zod";
+
+const KILOMENTERS_IN_METERS = 1000;
+const TWO_FRACTION_DIGITS = 2;
+
+const LatitudeLongitudeSchema = z.object({
```
pgurusinga, line 8: `airportCoordinateSystemSchema`?

app/util/calculateDistanceBetweenAirports.ts
```diff
@@ -0,0 +1,37 @@
...
+  latitude: z.coerce.number(),
+  longitude: z.coerce.number(),
+});
+
+function getAirportByIataCode(airportIataCode: string): unknown {
```
pgurusinga, line 13: why is the return value `unknown` ? if it finds something, it will the string or when it's not it returns `undefined`

app/models/flows/fluggastrechte/guards.ts
```diff
@@ -12,6 +13,13 @@ export const guards = {
       airportAbbreviations.includes(context.endAirport ?? "")
     );
   },
+  isNotPossibleCalculatedAirportsDistance: ({ context }) => {
+    const distance = calculateDistanceBetweenAirportsInKilometers(
+      context.startAirport ?? "",
+      context.endAirport ?? "",
+    );
+    return distance < 0;
```
pgurusinga, line 21: if the distance is `0`, is it still a valid airport distance? should it be `distance <== 0` instead?

pgurusinga: `isInvalidAirportDistance` might be shorter but not a blocker

app/util/calculateDistanceBetweenAirports.ts
aaschlote: it's returning the object of the airport in the json file, I don't have a type for that. Some when it founds something I don't know what it's there. If I chose to return non type here typescript will understand it'll return an ``any``. For avoiding ``any`` returns, I just return ``unknown ``. So even when it doesn't find anything, the type is ``unknown`` and I can parse it with ``zod``

app/models/flows/fluggastrechte/guards.ts
aaschlote: Nice comments, I'll do it :)

app/util/calculateDistanceBetweenAirports.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (airportStart.success && airportEnd.success) {
+    const haversineDistance = haversine(airportStart.data, airportEnd.data);
+    const roundDistance = (haversineDistance / KILOMENTERS_IN_METERS).toFixed(
+      TWO_FRACTION_DIGITS,
+    );
+    return Number(roundDistance);
+  }
+
+  return -1;
```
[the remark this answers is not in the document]
pgurusinga: I think we use `tiny-invariant` for handling such a case e.g. `app/services/array/resolveArrayCharacter.ts`

app/util/calculateDistanceBetweenAirports.ts
pgurusinga: I tried using a type generator out of it, does this help? :D 

```ts
type Airports = {
  iata: string;
  icao: string;
  time: string;
  city_code: string;
  country_code: string;
  airport: string;
  latitude: string;
  longitude: string;
  elevation_ft: string;
  region_name: string;
  city: string;
  county: string;
  state: string;
  type: string;
  continent: string;
  iso_region: string;
  scheduled_service: string;
  wikipedia_link: string;
  home: string;
  woeid: string;
  phone: string;
  email: string;
  runway_length: string;
  flightradar24_url: string;
  radarbox_url: string;
  flightaware_url: string;
}
```

aaschlote: I might help, but we don't use in our customize airports JSON file all those fields. I guess I would keep like this and maybe we can iterate on that

app/util/calculateDistanceBetweenAirports.ts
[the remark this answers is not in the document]
aaschlote: The ``tiny-invariant`` is nice, but it throws an exception and I would need to handle the exception everywhere I'm using this function. I added the ``true-myth`` library, let me know what do you think

app/models/flows/fluggastrechte/guards.ts
aaschlote: Done

pgurusinga (approved):
true-myth is nice, but we should also add this to dev exchange agenda and showcase this new library to the engineers. what do you think?

PR #774, opened 2024-04-25
E2E Test: Suggestion Input

aaschlote:


tests/e2e/pom/Vorabcheck.ts
```diff
@@ -26,30 +26,62 @@ export class Vorabcheck {
     await this.page.locator(`label[for="${field}-${option}"]`).click();
   }
 
-  async clickNext() {
+  async clickNext(javaScriptEnabled: boolean) {
```
chohner, line 29: this is a rather surprising argument, maybe it should rather be in an named argument?

tests/e2e/pom/Vorabcheck.ts
```diff
@@ -26,30 +26,62 @@ export class Vorabcheck {
...
     await this.page.locator(`button[name="${this.nextButtonName}"]`).click();
-    await this.page.waitForNavigation(); // deprecated but URL for waitForURL is unknown
+    if (javaScriptEnabled) {
```
chohner, line 31: I'm confused how this actually does anything for JS on the Page? 😅

aaschlote: The ``waitForNavigation`` has a parameter called ``waitUntil``, where all the methods needs JS on the page, example ``load``. More info in this [link](https://playwright.dev/docs/api/class-page#page-wait-for-url)

tests/e2e/pom/Vorabcheck.ts
aaschlote: It sounds a good idea, I'll do it :)

PR #785, opened 2024-04-30
Modify tile component

aaschlote:


app/services/cms/models/StrapiTile.ts
```diff
@@ -8,5 +8,6 @@ export const StrapiTileSchema = z
     value: z.string(),
     description: z.string().nullable(),
     image: StrapiImageSchema.nullable(),
+    tagDescription: z.string().optional().nullable(),
```
chohner, line 11: Afaik strapi fields are only nullable, not optional

aaschlote: true, but the field doesn't exist yet in the content when it runs the E2E test in Github, so it returns undefined and it was crashing without ``optional`` 😞

app/components/inputs/tile/Tile.tsx
```diff
@@ -36,9 +32,12 @@ const Tile = ({
         }`}
         htmlFor={id}
       >
-        <div className="ds-label-01-bold flex flex-row space-x-8 items-center">
+        <div
+          className={`ds-label-01-bold flex flex-row space-x-8 items-center " ${tagDescription ? "justify-between" : ""}`}
```
pgurusinga, line 36: I think we can use `classnames` package so that the class won't render with empty string if tag description would return false 

```js
classnames(
  "ds-label-01-bold flex flex-row space-x-8 items-center ",
  {"justify-between" :  Boolean(tagDescription)}
)
```

app/components/inputs/tile/TileRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+import TileTag from "./TileTag";
+
+const paragraphRenderer: Partial<Renderer> = {
+  paragraph(text) {
+    return `<p class="ds-subhead">${text}</p>`;
+  },
+};
+
+const IMAGE_HEIGHT = 32;
+const IMAGE_WIDTH = 32;
```
pgurusinga, line 15: probably can be shorter with `const IMAGE_DIMENSION = 32;`

app/services/cms/components/StrapiTileGroup.tsx
```diff
@@ -23,6 +23,11 @@ const StrapiTileGroupSchema = z
       ),
     }),
     useTwoColumns: z.boolean().nullable(),
+    showRadioButtonTiles: z
+      .boolean()
+      .nullable()
+      .transform((value) => value ?? false)
+      .default(false),
```
pgurusinga, line 30: can you explain in the transform we return false and at the end, we invoke `.default(false)` ?
can it be written like this?
```js
.transform(value)
.default(false)
```
or this code above has different effect?

app/services/cms/models/StrapiTile.ts
aaschlote: sure I can do it

app/components/inputs/tile/Tile.tsx
aaschlote: true, I'll add it

app/components/inputs/tile/TileRadio.tsx
aaschlote: I would like to have a variable for the both props, because they are two different things even right now it's the same value. Or can I create an object, like:

```
const IMAGE_DIMENSION = {
height: 32, width: 32
}
```

What do u think?

app/services/cms/components/StrapiTileGroup.tsx
aaschlote: The idea of the transformation was to add the value as false when it's null or undefined. We should have some value in this field, but it's a new field in Strapi I can't make it required.

I was checking the ``.default`` just in case it doesn't have any value, but I guess with the transformation is enough. I'll remove it

app/components/inputs/tile/TileRadio.tsx
pgurusinga: all fine 👍

PR #799, opened 2024-05-07
feat(test): add bundle size limit

chohner:
Issue: We sometimes bundle large dependencies or data chunks into our bundle. this adds a size limit for the bundle that is checked after unit tests:

image


package.json
```diff
@@ -83,6 +84,7 @@
     "@remix-run/dev": "^2.9.1",
     "@remix-run/eslint-config": "^2.9.1",
     "@remix-run/testing": "^2.9.1",
+    "@size-limit/file": "^11.1.2",
```
pgurusinga, line 87: why is it not a dev dependency?

aaschlote:
Nice idea, but I don't think we should add it during the unit tests process, but in a separate test process, because this is not an unit test. 

We can create a new job during the Github action for it and have some measures:

- Under 400kb green
- Until 600kb warning, do not stop deployment process
- Above 600kb error, stop deployment process

pgurusinga:
what happens to the commiet Chris? @chohner  :O

PR #811, opened 2024-05-13
Move unit tests to app folder

aaschlote:


tests/unit/services/licenseList.test.ts
```diff
@@ -1,16 +0,0 @@
-import { allRelevantLicenses } from "~/services/openSourceLicenses/generate.server";
```
chohner, line 1: imo we should actually enable this test instead of removing. or at least run it on CI

aaschlote: The test was skipped and half of the code was commented. Would it be fine if I create a ticket and make this test later?

frederike-ramin (approved):
Works as expected on my machine and the number of tests and runtime are the same / similar -> good to go.
(I didn't have a look at the file changes though, leaving that to Chris.

app/components/__test__/ArraySummary.test.tsx
```diff
@@ -57,10 +59,10 @@ describe("ArraySummary", () => {
...
 
     expect(queryByTestId("array-summary-title")).not.toBeInTheDocument();
-    expect(queryAllByTestId("array-summary-item").length).toBeGreaterThan(0);
-    expect(
-      queryAllByTestId("array-summary-item")[0].tagName.toLowerCase(),
-    ).toBe("h2");
+    expect(queryAllByTestId(arraySummaryItem).length).toBeGreaterThan(0);
+    expect(queryAllByTestId(arraySummaryItem)[0].tagName.toLowerCase()).toBe(
+      "h2",
+    );
```
pgurusinga, line 65: what was the reason for reordering?

pgurusinga: ah it's not reordering 🤣

eliflores: Hi 👋 
There was no reordering, as far as I can see. The change is related to the literal `""array-summary-item""` being extracted into a variable to be reused in other places in this file.

PR #812, opened 2024-05-14
Security: upgrade CSP headers

chohner:
- refactors into file + add unit tests
- add stricter directives
- add fallback for older browsers
- add references

We now full green checkmarks on CSP evaluator
image

todo:
- [x] document why we need `connect-src: *` in development?


app/services/security/cspHeader.server.tsx
```diff
@@ -0,0 +1,22 @@
...
+  [
+    `default-src 'self'`,
+    `script-src 'self' 'nonce-${cspNonce}' https://*.posthog.com 'strict-dynamic' https: 'unsafe-inline'`,
+    `style-src 'self' 'unsafe-inline'`,
+    `connect-src ${
+      config().ENVIRONMENT === "development"
+        ? "*"
+        : "'self' https://*.ingest.sentry.io https://*.posthog.com"
+    }`,
+    `img-src 'self' localhost:* https://a2j-rechtsantragstelle-infra-public-assets-bucket.obs.eu-de.otc.t-systems.com https://mermaid.ink data:`,
```
pgurusinga, line 18: I think we can conditionally set `localhost` for `dev` only — as far as I remember this is used for loading images when we start local Strapi (the images are referenced to the bucket)

tests/unit/services/security/cspHeader.test.ts
```diff
@@ -0,0 +1,19 @@
+import { cspHeader } from "~/services/security/cspHeader.server";
+
+describe("cspHeader", () => {
+  const header = cspHeader("");
+  it("should contain all OWASP recommended directives", () => {
```
pgurusinga, line 5: what do you think if we add a test for checking the correct structure of the CSP directives? we might misconfigured e.g. missing semicolon or too many spaces.
```
default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; connect-src 'self'; img-src 'self' localhost:*
```

aaschlote (approved):
Nicely done!

app/services/security/cspHeader.server.tsx
chohner: I just went to the opposite direction with the `connect-src` directive. afaik including localhost on prod isn't harmful but the more I think about it, the more I agree we should strip all mentions of localhost...

tests/unit/services/security/cspHeader.test.ts
chohner: Actually, I'm not sure what to test against here. but a malformed CSP will immediately break the site so it should be very noticeable

PR #838, opened 2024-05-21
refactor(PageContent): separate StrapiFormComponents

chohner:
No form components on content pages reduces traffic on content-only pages:

- before: 1,61 mB / 600,16 kB transferred
- after: 914,44 kB / 439,86 kB transferred


app/services/cms/components/StrapiFormComponents.tsx
```diff
@@ -0,0 +1,51 @@
...
+    case "form-elements.select":
+      return <StrapiSelect {...component} />;
+    case "form-elements.dropdown":
+      return <StrapiDropdown {...component} />;
+    case "form-elements.checkbox":
+      return <StrapiCheckbox {...component} />;
+    case "form-elements.tile-group":
+      return <StrapiTileGroup {...component} />;
+    default:
+      return <></>;
```
pgurusinga, line 39: it could return null instead of fragment

app/services/cms/components/StrapiFormComponents.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const FormComponents = ({
+  components,
+}: Readonly<{ components: Array<StrapiFormComponent> }>) => (
+  <>
+    {components.map((el) => (
+      <StrapiFormComponent key={keyFromElement(el)} component={el} />
+    ))}
+  </>
+);
```
pgurusinga, line 51: I would have difficulty searching for this Strapi to component mapper. Right now, I always check PageContent where the magic happens. Could we somehow move this closer to PageContent?

chohner: Good point. I also want to rename `<PageContent>` to `<StrapiContentComponents>` and move it to `cms/components`, so they lie closer together. Do you think I should also do this in this PR?

pgurusinga: let's make it happen in a different PR or a different commit :D

pgurusinga (approved):

PR #844, opened 2024-05-23
feat(infoBox): add separator toggle and tests

pgurusinga:


app/components/__test__/InfoBox.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    buttons: [],
+  },
+];
+
+describe("InfoBox", () => {
+  it("has expected padding when the separator is enabled", () => {
+    const { container } = render(
+      <InfoBox separator={true} items={mockInfoBoxItems} />,
+    );
+    expect(container.querySelector(".ds-stack-32")).toBeInTheDocument();
```
aaschlote, line 26: you could use also the ``toHaveClass``

app/components/InfoBox.tsx
```diff
@@ -1,22 +1,32 @@
...
   heading: HeadingPropsSchema.optional(),
+  separator: z.boolean().optional(),
   items: z.array(InfoBoxItemPropsSchema),
 });
 
 type InfoBoxProps = z.infer<typeof InfoBoxPropsSchema>;
 
-const InfoBox = ({ identifier, items, heading }: InfoBoxProps) => {
+const InfoBox = ({ identifier, items, heading, separator }: InfoBoxProps) => {
+  const showSeparator = separator ?? true;
```
aaschlote, line 16: can't make it in the zod in the line 9?

app/components/InfoBox.tsx
```diff
@@ -1,22 +1,32 @@
...
-      <ul className="list-none ds-stack-32 ps-0 info-box">
+      <ul
+        className={classNames("list-none ps-0 info-box", {
+          "ds-stack-48": !showSeparator,
+          "ds-stack-32": showSeparator,
+        })}
+      >
         {items.map((item) => (
           <InfoBoxItem
+            separator={showSeparator}
```
aaschlote, line 29: ``showSeparator`` sounds a better name for your props because the value is a boolean, but if the name is like this in Strapi, you can keep as it's

aaschlote (approved):
Add a few comments, just check if the storybook is still working, besides that LGTM

app/components/__test__/InfoBox.test.tsx
pgurusinga: yeah we could but then the result of the screen.getByRole("list") will return something like this
image

and we need another logic that the expected result should be part/subset of it

aaschlote: you add the dot (`.`), you shouldn't have it, so your test would be `toHaveClass("ds-stack-48")`

PR #852, opened 2024-05-24
Refactor FlowNavigation

chohner:
Mostly replaces subflownavigation & subnavitem with nesting. This was done now to simplify the upcoming restyling

PR #856, opened 2024-05-27
restyle flow navigation

chohner:
Depends on https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/852

look:
image

Todo:
- [x] `DoneCurrent` state
- [x] restyle active parent
- [ ] potential refactor enum

### DIFF
--- a/app/components/FlowNavigation.tsx
+++ b/app/components/FlowNavigation.tsx
@@ -46,23 +46,11 @@ type FlowNavigationProps = Readonly<{
 
 const NavigationList = ({
   navItems,
-  a11yLabels,
-  isChild = false,
+  ...props
 }: FlowNavigationProps & { isChild?: boolean }) => (
-  <ul
-    className={
-      isChild
-        ? "pl-32 mr-8 min-w-fit max-w-fit md:min-w-[250px] md:max-w-[250px] break-words"
-        : ""
-    }
-  >
+  <ul className="pl-0">
     {navItems.map((navItem) => (
-      <NavItem
-        {...navItem}
-        key={navItem.destination}
-        a11yLabels={a11yLabels}
-        isChild={isChild}
-      />
+      <NavItem {...navItem} key={navItem.destination} {...props} />
     ))}
   </ul>
 );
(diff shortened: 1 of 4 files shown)


app/services/navigation/navState.ts
```diff
@@ -1,7 +1,27 @@
...
+  DoneCurrent,
   Current,
   Open,
   OpenDisabled,
 }
+
+export const stateIsCurrent = (state: NavState) =>
+  state === NavState.Current || state === NavState.DoneCurrent;
+
+export const stateIsActive = (state: NavState) =>
```
aaschlote, line 13: do we have an unit test for it?

app/services/navigation/navState.ts
```diff
@@ -1,7 +1,27 @@
...
+  [
+    NavState.DoneCurrent,
+    NavState.Current,
+    NavState.Open,
+    NavState.Done,
+  ].includes(state);
+
+export const stateIsDone = (state: NavState) =>
```
aaschlote, line 21: do we have an unit test for it?

aaschlote (commented):
Can we think about to break the FlowNavigation.tsx in small parts? I have the feeling it's very difficult to read it, because in this file has 4 react components and 3 types.

Idea: create a new folder called `flowNavigation` or `navigation` under the components, it would be something

components
    &emsp;flowNavigation or navigation
  &emsp;&emsp;StateIcon.tsx
  &emsp;&emsp;NavigationList.tsx
   &emsp;&emsp;FlowNavigation.tsx
   &emsp;&emsp;NavItem.tsx

chohner:
I had similar ideas regarding refactoring but I wanted to keep this PR close to a restyle as possible.

app/services/navigation/navState.ts
chohner: I would like to wait with unit test for two reasons: This is imo too trivial (dangerous statement, I know) and there is no way to unit test this without recreating the functionality of this helper. More importantly, I think we should refactor this to remove the `NavState` enum altogether (again, not in this PR)

app/components/FlowNavigation.tsx
```diff
@@ -78,58 +60,52 @@ export default function FlowNavigation(props: FlowNavigationProps) {
...
+
+  const liClassNames = classNames("list-none", {
+    "text-gray-600 curser-not-allowed hover:font-normal pointer-events-none":
+      isDisabled,
+    "border-b-[1px] border-b-white last:border-b-0": !isChild,
+    "border-l-[1px] border-l-blue-800": isCurrent,
+  });
+
+  const itemClassNames = classNames(
+    "w-full	ds-label-02-reg p-16 flex gap-x-16 items-center hover:underline hover:bg-blue-300 active:bg-white	focus-visible:shadow-[inset_0px_0px_0px_4px] focus:shadow-blue-800",
```
pgurusinga, line 88: why they have big whitespace? probably tab?
```suggestion
    "w-full ds-label-02-reg p-16 flex gap-x-16 items-center hover:underline hover:bg-blue-300 active:bg-white focus-visible:shadow-[inset_0px_0px_0px_4px] focus:shadow-blue-800",
```

app/components/FlowNavigation.tsx
```diff
@@ -78,58 +60,52 @@ export default function FlowNavigation(props: FlowNavigationProps) {
...
           <button
-            className={`${stateClassNames} md:pr-0 min-w-[242px] relative flex items-center w-full cursor-pointer flex gap-x-16 items-center`}
+            className={itemClassNames}
             aria-disabled={isDisabled}
             aria-expanded={collapse.isExpanded}
             {...collapse.getToggleProps()}
             aria-describedby={iconId}
           >
-            <StateIcon isDone={isDone} id={iconId} a11yLabels={a11yLabels} />
+            {isDone && <StateIcon id={iconId} a11yLabels={a11yLabels} />}
```
pgurusinga, line 108: this is a good one!

pgurusinga (approved):

PR #862, opened 2024-05-28
Refactor(pdf generation): replace quicktype, add static getter of pdf parameters

chohner:
- removes quicktype dependency, instead export our own types
- also export static `getBeratungshilfePdfFromContext()`, which cleanly returns a new object to fill PDFs with (no await, no need to parse pdf etc)

@pgurusinga I had to adapt a few pdf unit tests, mostly where the tests previously relied on `false` instead of `undefined`. I think the new behaviour is correct but please double-check 🙏🏼 

Also, I think the unit tests are now much quicker (4-5 s on my machine, instead of ~10s)

some more stats on the /pdf endpoint (using 10s autocannon):
- before: avg latency: 468 ms, avg req/sec: 21
- after: avg latency: 352 ms, avg req/sec: 28


pgurusinga:
@chohner I'm missing some context. I noticed that the getBeratungshilfeParameters function returns a mapping of field names to structured names. Do I need to manually add the mapping for future documents?
 
image

chohner:
Hopefully not, thats the output of the pdf generator script!

app/services/pdf/beratungshilfe/sections/F_besitz/__test__/F_besitz.test.ts
```diff
@@ -70,16 +70,16 @@ describe("F_besitz", () => {
...
       fillFinancialBankkonto(attachment, pdfFields, context);
 
       expect(pdfFields.f1Konten1.value).toBe(false);
       expect(pdfFields.f1Konten2.value).toBe(true);
-      expect(pdfFields.f1InhaberA.value).toBe(false);
-      expect(pdfFields.f2InhaberB.value).toBe(false);
-      expect(pdfFields.f2InhaberC.value).toBe(false);
+      expect(pdfFields.f1InhaberA.value).toBe(undefined);
+      expect(pdfFields.f2InhaberB.value).toBe(undefined);
+      expect(pdfFields.f2InhaberC.value).toBe(undefined);
```
pgurusinga, line 82: could you explain what the decision was to be undefined? were these expectations for empty value?

app/services/pdf/pdf.generator.ts
```diff
@@ -1,145 +1,67 @@
...
-      path.resolve(path.join(process.cwd(), dataDirectory, pdfFilePath)),
-    ),
-  );
-  const form = pdfDoc.getForm();
-  const fields = form.getFields();
-
-  const quickTypeJSONSchema: { [k: string]: any } = {};
+const pdfs = [
+  {
+    service: "beratungshilfe",
```
pgurusinga, line 8: what is the service key usage for?

chohner: its only there for namespacing. but looking at it know, maybe an object makes more sense anyways

app/services/pdf/beratungshilfe/sections/F_besitz/__test__/F_besitz.test.ts
chohner: So - as far as I understand, these values never actually get set in our filling logic and I think its more consistent to therefore leave them as `undefined`.

However, the object returned from `getBeratungshilfeParameters()` had `false` default values so thats why I had to change the test.

pgurusinga (approved):

PR #870, opened 2024-05-31
Correct compensationPayment-calculation

aaschlote:


app/models/flows/fluggastrechte/__test__/stringReplacaments.test.ts
```diff
@@ -1,79 +1,93 @@
...
 
-const mockedCalculateDistanceBetweenAirports =
-  calculateDistanceBetweenAirportsInKilometers as jest.Mocked<
-    typeof calculateDistanceBetweenAirportsInKilometers
+const mockedGetRouteCompensationBetweenAirports =
+  getRouteCompensationBetweenAirports as jest.Mocked<
+    typeof getRouteCompensationBetweenAirports
   >;
 
 const TRANSLATION_KEY_RECORD = {
```
pgurusinga, line 24: what is this variable for? is it the expected results?

app/services/airports/__test__/getRouteCompensationBetweenAirports.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    expect(actual).toEqual("shortDistance");
+  });
+
+  it("should return notPossibleCalculateDistance if the distance can not be calculated", () => {
+    (mockedCalculateDistanceBetweenAirports as jest.Mock).mockReturnValue(
+      Result.err("anything"),
+    );
+
+    const actual = getRouteCompensationBetweenAirports("BER", "MUN");
```
pgurusinga, line 86: why don't we use real start and end airport codes and skip the mock? It feels weird to have the same airport code in the code that behaves differently in each test case

app/services/airports/getRouteCompensationBetweenAirports.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (distanceKm.isOk) {
+    if (distanceKm.value > LONG_DISTANCE_KILOMETERS) {
+      const isStartAirportEU = isEuropeanUnionAirport(startAirport);
+      const isEndAirportEU = isEuropeanUnionAirport(endAirport);
+
+      if (isStartAirportEU.isOk && isEndAirportEU.isOk) {
+        return isStartAirportEU.value && isEndAirportEU.value
+          ? "longDistanceInsideEU"
+          : "longDistanceOutsideEU";
```
pgurusinga, line 31: what does `.isOK` represent? is it either fulfilled (EU airports) OR not fulfilled (any airports incl. EU) because logically, I would assume that the statement `isStartAirportEU.isOk && isEndAirportEU.isOk` means that the logic inside the block scope would only handle both EU in AND out flight

app/models/flows/fluggastrechte/__test__/stringReplacaments.test.ts
aaschlote: yes, I'm using it to check the expected results I need it. I'll rename it

app/services/airports/__test__/getRouteCompensationBetweenAirports.test.ts
aaschlote: I could use, the only problem if we update the airports data and it changes the values of the distance or the country the test can fail at some point. 

Another thing, this is an unit test, so I should expect what the function `getRouteCompensationBetweenAirports` returns and not the function inside, for this reason I mocked them. The other functions has their own unit tests.

app/services/airports/isEuropeanUnionAirport.ts
```diff
@@ -0,0 +1,19 @@
+import { Result } from "true-myth";
+import airports from "data/airports/data.json";
```
chohner, line 2: these are now both pulled into the frontend via the validators :/

app/services/airports/getRouteCompensationBetweenAirports.ts
aaschlote: we have a problem that the any airport code can be entered in the function `isEuropeanUnionAirport`, so the `.isOk` means `The airport is valid and found on the DB`. When it returns `.isError` it means the `Airport not found`. It also describe on the error message on the `isEuropeanUnionAirport` (line 15)

app/services/airports/isEuropeanUnionAirport.ts
aaschlote: This function is called only in the guards, is it called also in the frontend? I guess only the zod validation, right?

chohner: yes, all validations are run on front and backend

app/services/airports/getRouteCompensationBetweenAirports.ts
```diff
@@ -0,0 +1,43 @@
...
+export function getRouteCompensationBetweenAirports(
+  startAirport: string,
+  endAirport: string,
+): RouteCompensation {
+  const distanceKm = calculateDistanceBetweenAirportsInKilometers(
+    startAirport,
+    endAirport,
+  );
+
+  if (distanceKm.isOk) {
```
chohner, line 23: Might be more readable to have an early return on error and keep the rest of the function un-indented?

app/services/airports/isEuropeanUnionAirport.ts
aaschlote: I couldn't find out where the `guards` are called in the frontend, can you show to me it?

app/services/airports/getRouteCompensationBetweenAirports.ts
aaschlote: true, I changed it :)

PR #876, opened 2024-06-04
Rate limit redis

frederike-ramin:
# The problem being solved
Our rate limiting before was specific to one pod, meaning that it will be higher if we have more pods -> beside the point of rate limiting.

# The solution
Use our existing redis for rate limiting -> one point where usage data is stored.
With that I had a lot of code duplication and thus with the idea of @chohner introduced the only-tls way. (Before, we had the local setup without TLS, leading to some logic that needed to be duplicated).
I have to say, I do like that the local setup is now closer to the rest of the setup.

# Questions/ considerations
- Do you see any further data privacy issues with this?
- Do you think the code-duplication has a decent amount? (it's not nicely possible to import the same code in `server.js` and `redis.ts`)
- Does the documentation + setup work for you on your machine?
- Do you see a point where this might break in other environments?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,22 @@ First draft of implementing a platform to create requests to the Rechtsantragste
 
 ### Run Server in Development Mode
 
+#### Start a Redis instance with TLS setup
+
+1. Create TLS certificate + key: [Here](https://serverfault.com/questions/224122/what-is-crt-and-key-files-and-how-to-generate-them) is how.
+   You can pass dummy information into the certificate.
+
+2. Name the files redis-server.crt and put them in a /cert folder.
+3. Start the redis:
+
+```
+docker compose up -d
+```
+
+#### Start the development server
+
 ```sh
 npm install
-docker compose up -d
 npm run dev
 ```
 
--- a/server.js
+++ b/server.js
@@ -1,8 +1,10 @@
 import { wrapExpressCreateRequestHandler } from "@sentry/remix";
 import { createRequestHandler } from "@remix-run/express";
 import { rateLimit } from "express-rate-limit";
+import { RedisStore } from "rate-limit-redis";
 import compression from "compression";
 import express from "express";
+import RedisClient from "ioredis";
 
 const sentryCreateRequestHandler =
   wrapExpressCreateRequestHandler(createRequestHandler);
@@ -56,9 +58,15 @@ isStagingOrPreviewEnvironment
 
 app.set("trust proxy", 2);
 
+const redisUrl = () =>
+  `rediss://default:${process.env.REDIS_PASSWORD?.trim()}@${process.env.REDIS_ENDPOINT ?? "localhost:6380"}`;
+const client = new RedisClient(redisUrl(), {
+  tls: {
+    rejectUnauthorized: false,
+  },
+});
+
 // Limit calls to routes ending in /pdf or /pdf/, as they are expensive
-// Attention - this only limits to the number of requests *per pod*.
-// The more pods we have, the higher the effective limit goes up.
 app.use(
   /.*\/pdf(\/|$)/,
   rateLimit({
@@ -68,6 +76,9 @@ app.use(
     legacyHeaders: false, // Disable the `X-RateLimit-*` headers
     message:
       '<!DOCTYPE html><html lang="de"><head><meta charset="utf-8"><title>Justiz Services - Fehler aufgetreten</title><style>html{font-family:BundesSansWeb,Calibri,Verdana,Arial,Helvetica,sans-serif;font-size:1.125rem;line-height:1.75rem}body{max-width:59rem;margin:6rem auto;padding:0 2rem}h1{font-size:1.875rem;padding-bottom:1.5rem}</style></head><body><h1>Justiz Service ist vorübergehend nicht erreichbar</h1><p>Es sind zu viele Anfragen zum Herunterladen des PDFs innerhalb kurzer Zeit gestellt worden. Bitte versuchen Sie es später noch einmal.</p></body></html>',
+    store: new RedisStore({
+      sendCommand: (...args) => client.call(...args),
+    }),
   }),
 );
 
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,8 @@ lefthook-local.yml
 # Gerichtsfinder data
 **/gerichtsfinder/_data/
 
+# local Redis TLS certs
+/certs
+
 /.vscode/
 .eslintcache
--- a/app/services/session.server/redis.ts
+++ b/app/services/session.server/redis.ts
@@ -1,6 +1,5 @@
-import { Redis, type RedisOptions } from "ioredis";
+import { Redis } from "ioredis";
 import { config } from "../env/env.server";
-import { config as configWeb } from "../env/web";
 import { logError } from "../logging";
 
 declare global {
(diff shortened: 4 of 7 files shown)


server.js
```diff
@@ -68,6 +76,9 @@ app.use(
     legacyHeaders: false, // Disable the `X-RateLimit-*` headers
     message:
       '<!DOCTYPE html><html lang="de"><head><meta charset="utf-8"><title>Justiz Services - Fehler aufgetreten</title><style>html{font-family:BundesSansWeb,Calibri,Verdana,Arial,Helvetica,sans-serif;font-size:1.125rem;line-height:1.75rem}body{max-width:59rem;margin:6rem auto;padding:0 2rem}h1{font-size:1.875rem;padding-bottom:1.5rem}</style></head><body><h1>Justiz Service ist vorübergehend nicht erreichbar</h1><p>Es sind zu viele Anfragen zum Herunterladen des PDFs innerhalb kurzer Zeit gestellt worden. Bitte versuchen Sie es später noch einmal.</p></body></html>',
```
pgurusinga, line 78: FYI, unrelated to PR, does this error response include the security header such as CSP?

README.md
```diff
@@ -15,9 +15,22 @@ First draft of implementing a platform to create requests to the Rechtsantragste
 
 ### Run Server in Development Mode
 
+#### Start a Redis instance with TLS setup
```
pgurusinga, line 18: Is setting up a local development environment with TLS necessary for the local instance to work, or is it an optional step?

README.md
```diff
@@ -15,9 +15,22 @@ First draft of implementing a platform to create requests to the Rechtsantragste
...
+
+1. Create TLS certificate + key: [Here](https://serverfault.com/questions/224122/what-is-crt-and-key-files-and-how-to-generate-them) is how.
+   You can pass dummy information into the certificate.
+
+2. Name the files redis-server.crt and put them in a /cert folder.
```
pgurusinga, line 23:
```suggestion
2. Name the files redis-server.crt and put them in a /certs folder.
```

since the ignore file is using `s` at the end

doc/adr/0006-express-rate-limiting.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Decision
+
+We decided to employ a rate limiting mechanism on the specific route.
+The following options were explored:
+
+1. handling it on an infrastructure level by employing rate limiting in a WAF.
+   - downside: This was excluded due to previous problems with IP addresses in other projects.
+2. [Express rate limiting](https://express-rate-limit.mintlify.app/overview) with in-memory store.
+   - upside: Handling rate limiting before reaching the application.
```
pgurusinga, line 22: I thought express-rate-limit in `server.js` is part of the application (the image itself) or am I wrong?

doc/adr/0006-express-rate-limiting.md
```diff
@@ -0,0 +1,36 @@
...
+   - downside: Only possible to track limits _per pod_.
+3. Storing the number of requests per user in redis on every loader + action call.
+   - upside: Having one place across all replicas where the number of requests are stored.
+   - downside: This moves the responsibility to the application layer, costing more CPU power. Plus, we'll have to remember to include it in future loaders and actions
```
pgurusinga, line 26: Assuming that `server.js` is part of the application layer, would this downside also apply to the second and third points?

doc/adr/0006-express-rate-limiting.md
```diff
@@ -0,0 +1,36 @@
+# 4. Preview/Test Environment
```
aaschlote, line 1: just a reminder here to rename it before merge 😃

README.md
aaschlote: it'll be required as far I understand

frederike-ramin:
@zechmeister @aaschlote I tried to include the cal in the docker-compose. However, we would have to execute it before even starting the container and docker currently offers no way of doing this: https://github.com/docker/compose/issues/6736
That's why I included the call in the playwright config which apparently also did not work unfortunately. :/

doc/adr/0006-express-rate-limiting.md
frederike-ramin: Mh, I heard that it is part of the "middleware" which meant to me that it's not part of the application, buuuuut I might be wrong with the wording there.

aaschlote:
> @zechmeister @aaschlote I tried to include the cal in the docker-compose. However, we would have to execute it before even starting the container and docker currently offers no way of doing this: [docker/compose#6736](https://github.com/docker/compose/issues/6736) That's why I included the call in the playwright config which apparently also did not work unfortunately. :/

I moved the creation of the certificates before to run the e2e test, it seems it works, but it doesn't connect with redis 😞 

It should be some problem to move the certificates that it has been created to the redis docker image volume

aaschlote:
> > @zechmeister @aaschlote I tried to include the cal in the docker-compose. However, we would have to execute it before even starting the container and docker currently offers no way of doing this: [docker/compose#6736](https://github.com/docker/compose/issues/6736) That's why I included the call in the playwright config which apparently also did not work unfortunately. :/
> 
> I moved the creation of the certificates before to run the e2e test, it seems it works, but it doesn't connect with redis 😞
> 
> It should be some problem to move the certificates that it has been created to the redis docker image volume

I found the problem:

```
local-redis  | 1:M 07 Jun 2024 07:23:54.801 # Failed to load private key: /usr/local/etc/redis/redis-server.key: error:8000000D:system library::Permission denied
```

The permission to access the `redis-server.key` inside docker was not working in the Github action. I changed the permission of the file during the creation of the certificates from `chmod 400` to `chmod 444` and now it's working

server.js
```diff
@@ -58,9 +60,15 @@ isStagingOrPreviewEnvironment
 // we have in front of our express, which is 2 - experimentally found out.
 app.set("trust proxy", 2);
 
+const redisUrl = () =>
+  `rediss://default:${process.env.REDIS_PASSWORD?.trim()}@${process.env.REDIS_ENDPOINT ?? "localhost:6380"}`;
+const client = new RedisClient(redisUrl(), {
+  tls: {
+    rejectUnauthorized: false,
+  },
+});
```
chohner, line 69: Is there a straightforward way to share this logic between here and `app/services/session.server/redis.ts`? Might be tricky due to cjs vs esm

server.js
```diff
@@ -58,9 +60,15 @@ isStagingOrPreviewEnvironment
 // we have in front of our express, which is 2 - experimentally found out.
 app.set("trust proxy", 2);
```
chohner, line 61: We probably should make the `2` an env variable

server.js
zechmeister: Good question! The error response does not use our CSP config. Instead of a `message`, we could use a custom handler:

```
app.use(
  /.*\/pdf(\/|$)/,
  rateLimit({
    ...
    handler: (req, res) => {
      res.setHeader("Content-Security-Policy", cspHeader());
      res
        .status(429)
        .send(
          '<!DOCTYPE html><html lang="de">...',
        );
    },
    ...
  }),
);
```

However, we'd have to use something like ts-node to run the server since the `cspHeader()` is typescript (currently `node server.js`).

README.md
zechmeister: Step is handled automatically now, removed the instructions

doc/adr/0006-express-rate-limiting.md
zechmeister: Yeah I guess from a infra perspective, we can consider the `server.js` the entrypoint part of the application that orchestrates how our backend interacts with http requests (and our middleware setup is part of the `server.js`).

However, it's outside of our core _remix app_; I update the adr.

server.js
zechmeister: I guess we could introduce a js module, that we use in `server.js` and `redis.ts`. Or the other way round by compiling something before running `node server.js`... However I'd argue the code duplication is minimal and the added complexity outweighs the duplication?

server.js
zechmeister: Do we need to configure this for different env? Otherwise I'd say yagni?

server.js
frederike-ramin: This seems worth investigating more? Did you create a ticket for this by any chance?

frederike-ramin:
@zechmeister and @aaschlote Thank you for taking care of this! 🙏

PR #894, opened 2024-06-07
feat(test): move to vitest

chohner:
- replaces `jest` with `vitest`
- splits `eigentumZusammenfassungDone` and corresponding test into new files for better mocking
- fixes `omitNull()` test
- sets up coverage without routes
- new `npm run test:watch` command

This will enable usage of [vite-env-only](https://github.com/pcattori/vite-env-only) for server-side validations


pgurusinga:
50% improvement 🎉

pgurusinga (approved):

aaschlote (changes requested):
Can you give some details about the benefits of using Vitest? Maybe writing an ADR ir discuss with us during the Dev Exchange meeting.

doc/adr/0006-vitest-test-runner.md
```diff
@@ -0,0 +1,32 @@
...
+Date: 2024-06-07
+
+## Status
+
+Agreed
+
+## Context
+
+We are currently using [Jest](https://jestjs.io/) as our test runner, which was most popular testing framework when starting this project. However, there are a few downsides:
+
```
aaschlote, line 12: I guess it'll fix the issue related the unit test for remix loader that we have faced in the past. [Link](https://github.com/remix-run/remix/issues/7967#issuecomment-1806536580)

vitest.config.ts
```diff
@@ -0,0 +1,25 @@
...
+export default defineConfig({
+  test: {
+    dir: "./app",
+    include: ["**/__test__/*.test.{ts,tsx}"],
+    globals: true,
+    environment: "jsdom",
+    setupFiles: ["./tests/unit/vitest.setup.ts"],
+    coverage: {
+      include: ["app/**"],
+      exclude: ["app/**/__test__/**", "app/routes/**"],
```
aaschlote, line 15: about the `routes`, a few files we need to have in the coverage, for example the files in the `routes/shared` and the routes that are not in the flow process (beratungshilfe, fluggastrecht and etc). Maybe we can revisit this topic again

tsconfig.json
```diff
@@ -19,7 +19,7 @@
     "moduleResolution": "Bundler",
     "resolveJsonModule": true,
     "target": "ES2022",
-    "types": ["@testing-library/jest-dom", "node"],
+    "types": ["@testing-library/jest-dom", "node", "vitest/globals"],
```
aaschlote, line 22: do we still need the `@testing-library/jest-dom`?

I guess if we remove the files where the content has:

```
/**
 * @jest-environment node
 */
```

we can get rid of it, right?

aaschlote (approved):

tsconfig.json
chohner: we need `jest-dom` for our component tests but I removed the `node` env where possible

PR #897, opened 2024-06-10
refactor: TileTag component 

eliflores:
✨

### DIFF
--- a/app/components/inputs/tile/__test__/TileTag.test.tsx
+++ b/app/components/inputs/tile/__test__/TileTag.test.tsx
@@ -2,19 +2,23 @@ import { render } from "@testing-library/react";
 import TileTag from "~/components/inputs/tile/TileTag";
 
 describe("TileTag", () => {
-  it("it should render the component, in case enter a value in the props", () => {
-    const mockTagDescription = "value";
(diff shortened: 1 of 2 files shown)


app/components/inputs/tile/__test__/TileTag.test.tsx
```diff
@@ -2,19 +2,23 @@ import { render } from "@testing-library/react";
 import TileTag from "~/components/inputs/tile/TileTag";
 
 describe("TileTag", () => {
-  it("it should render the component, in case enter a value in the props", () => {
-    const mockTagDescription = "value";
+  describe("when props are passed", () => {
```
aaschlote, line 5: I don't see so much reason to have two `describes` with each one having only one `it` calls. Can you explain why have you done this change?

eliflores: It's so that the style of the description in the reporting looks more organized, according to the use cases:
```
✓ app/components/inputs/tile/__test__/TileTag.test.tsx (2)
   ✓ TileTag (2)
     ✓ when props are passed (1)
       ✓ should render properly
     ✓ when no props are passed (1)
       ✓ should render properly
```

PR #903, opened 2024-06-12
Update country codes which count as "EU" airports

aaschlote:


app/services/airports/__test__/isEuropeanUnionAirport.test.ts
```diff
@@ -20,4 +20,25 @@ describe("isEuropeanUnionAirport", () => {
 
     expect(actual.isErr).toBe(true);
   });
+
+  it("should return ok and value true if the airport is ZRH", () => {
+    const actual = isEuropeanUnionAirport("ZRH");
+
+    expect(actual.isOk).toBe(true);
+    expect(actual.isOk ? actual.value : false).toBe(true);
```
pgurusinga, line 28: what is the intention of this test? it looks weird that an expectation has a ternary logical.

aaschlote: I need to get the `value` of the result, to be able to get it  the function `isOk` must be true, for this reason I created this ternary logical. Another option is making the same implementation has been done like [here](https://github.com/true-myth/true-myth/blob/main/test/result.test.ts#L492), but I need to add an extra `if` conditional for that

pgurusinga: The link you mentioned is better in terms of readability. let's do the one you did before

pgurusinga: yes, with the `if` block I think much better

PR #927, opened 2024-06-19
New airports datasource

aaschlote:
- Refactor: Change file `airports.cjs` to `airports.ts`
- Refactor: Change data source from library `airport-data-js` to CSV file `https://davidmegginson.github.io/ourairports-data/airports.csv`
- Refactor: Use `zod` to validate data from CSV file
- Refactor: Remove library `airport-data-js`


scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+  console.log(`Error code ${res.status}. Status text: ${res.statusText}`);
+  return [];
+}
+
+async function fetchAndSaveAirports() {
+  const airports = await fetchAllAirports();
+  const data = JSON.stringify(_.uniqBy(airports, "iata"));
+  const filePath = "data/airports/data.json";
+
+  fs.writeFile(filePath, data, (error) => {
```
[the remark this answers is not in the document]
aaschlote: We are reading the file to get the airports asynchronously in the function `airports`

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
+import fs from "fs";
+import _ from "lodash";
+import countriesTranslation from "i18n-iso-countries";
+import { z } from "zod";
+
+const germanLocale = "de";
```
eliflores, line 7: Since the name of the other constants is capitalized, shall we also capitalize the name of this const:

```suggestion
const GERMAN_LOCALE = "de";
```

?

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+const CITIES_AIRPORTS_DE = "scripts/cities_airports_de.csv";
+const AIPORTS_URL_DATA_SOURCE =
```
eliflores, line 9:
```suggestion
const AIRPORTS_URL_DATA_SOURCE =
```

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+    if (airportResult.success) {
+      airports.push(airportResult.data);
+    }
+  });
+
+  return airports;
+}
+
+async function fetchAllAirports(): Promise<Aiport[]> {
+  const res = await fetch(AIPORTS_URL_DATA_SOURCE, {
```
eliflores, line 136:
```suggestion
  const res = await fetch(AIRPORTS_URL_DATA_SOURCE, {
```

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+        (airport.type === "large_airport" ||
+          airport.type === "medium_airport") &&
+        airport.scheduled_service === "yes",
+    )
+    .map((airport) => {
+      return {
+        iata: airport.iata,
+        country_code: airport.country_code,
+        country:
+          countriesTranslation.getName(airport.country_code, germanLocale) ??
```
eliflores, line 83: Related to the previous comment about capitalization of constants in this file:

```suggestion
          countriesTranslation.getName(airport.country_code, GERMAN_LOCALE) ??
```

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  if (res.status === 200) {
+    const content = await res.text();
+    const airportsDataSource = parseAiportDataSource(content);
+    return filteredLargeMediumAirports(airportsDataSource);
+  }
+
+  console.log(`Error code ${res.status}. Status text: ${res.statusText}`);
+  return [];
```
eliflores, line 149: Should we throw here instead of returning an empty array? 
I think failing fast in this case would make more sense, since if there is no data from the airports data source, no entries will be generated and written into the `data/airports/data.json` file. 
Otherwise, with only a log, if one is not paying attention to the logs (which may be the case), one may wonder why there is no data being written into the file.

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+  const filePath = "data/airports/data.json";
+
+  fs.writeFile(filePath, data, (error) => {
+    if (error) {
+      console.error(error);
+      throw error;
+    }
+  });
+  console.log(`${filePath} written correctly`);
+}
```
eliflores, line 164: I believe the comment about writing to the file synchronously is referring to [this](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/927/files#diff-5e57d179fb325d62c54ca32067b8eddc9c34a1491049541732f532a75572afa2R11-R13) part? And if we could also write to the file using `writeFileSync`: 

```suggestion
  fs.writeFileSync(filePath, data);
  console.log(`${filePath} written correctly`);
}
```

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+  longitude: z.coerce.number(),
+  city: z.string().transform(removeDoubleQuotes),
+  type: z.string().transform(removeDoubleQuotes),
+  scheduled_service: z.preprocess(
+    (val) => removeDoubleQuotes(val),
+    z.enum(["yes", "no"]),
+  ),
+});
+
+type AiportDataSource = z.infer<typeof AirportDataSourceSchema>;
```
eliflores, line 38: small typo: `AiportDataSource` -> `AirportDataSource`

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+
+type Aiport = Omit<AiportDataSource, "type" | "scheduled_service"> & {
```
eliflores, line 40: small typo: `Aiport` -> `Airport`

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+          "",
+        airport: translateAirportName(airport.airport),
+        latitude: airport.latitude,
+        longitude: airport.longitude,
+        city: translateAirportCity(airport),
+      };
+    });
+}
+
+function parseAiportDataSource(content: string): AiportDataSource[] {
```
eliflores, line 93:
```suggestion
function parseAirportDataSource(content: string): AiportDataSource[] {
```

scripts/airports.ts
```diff
@@ -0,0 +1,165 @@
...
+async function fetchAllAirports(): Promise<Aiport[]> {
+  const res = await fetch(AIPORTS_URL_DATA_SOURCE, {
+    headers: {
+      "content-type": "text/csv;charset=UTF-8",
+    },
+  });
+
+  if (res.status === 200) {
+    const content = await res.text();
+    const airportsDataSource = parseAiportDataSource(content);
```
eliflores, line 144:
```suggestion
    const airportsDataSource = parseAirportDataSource(content);
```

scripts/airports.ts
aaschlote: Ohh I got what @pgurusinga meant now. It sounds much better like this, I'll change it!

scripts/airports.ts
aaschlote: I was thinking about this possibility, although if the file doesn't have any data, I would assume the person may see it once it commit the file `data/airports/data.json`.

But I'm fine to thrown an exception, if you prefer this approach instead of have a empty file data

scripts/airports.ts
pgurusinga: sehrrrrrr schöööööön

scripts/airports.ts
eliflores: Yes, let's throw :) I think it'd make it more clear.
Thank you!

pgurusinga (approved):

eliflores (approved):
LGTM 🎉 ✨ 🚀

PR #937, opened 2024-06-21
digitalservicebund/icons v2

pgurusinga:
# Problem
Since we are ESM-project and `@digitalservicebund/icons` is a CJS-type lib, we would need Vite's `vite-plugin-cjs-interop` to make the library work in our application

# Proposed Changes 
- Apply icons v2 https://github.com/digitalservicebund/icons/blob/main/CHANGELOG.md
- Remove Vite Interop helper
- Adapt `tsconfig` to resolve path


aaschlote:
Can you check now if we update the dependency `tsx` to the version `^4.15.1`, the generation of CMS file doesn't it fail?

pgurusinga:
@aaschlote yes, i checked last week and it works without this change ;) , but it won't be part of this PR

aaschlote:
> @aaschlote yes, i checked last week and it works without this change ;) , but it won't be part of this PR

Wasn't it failing before due the update of the dependency (tsx)? 🤔

aaschlote (approved):

eliflores (approved):
🚀

chohner (approved):
nice!

PR #938, opened 2024-06-24
Remove link text logic

aaschlote:


eliflores (approved):
LGTM 👍

pgurusinga (approved):
I would also ask Rast to have a look at this PR

frederike-ramin (changes requested):
Very glad you asked for a review, because, no, that is not quite what I meant with deleting the "delete data" logic. I meant deleting the whole logic inside `app/services/flow/server/lastStep.ts` concerning `dataDeletionKey`. Could you remove that as well as this is the only place this is ever used?

aaschlote:
> dataDeletionKey

aaschlote:
> Very glad you asked for a review, because, no, that is not quite what I meant with deleting the "delete data" logic. I meant deleting the whole logic inside `app/services/flow/server/lastStep.ts` concerning `dataDeletionKey`. Could you remove that as well as this is the only place this is ever used?

Thanks, yes this is the only place in the source code where we have this `dataDeletionKey`, but I found out this URL in strapi where it calls the `deleteData` as query parameter. The others knows this functionality won't work anymore? I can remove it from Strapi.

image

aaschlote:
> That's pretty new and it seems to be on one of your pages? `/fluggastrechte/vorabcheck/ersatzflug-starten-eine-landen-zwei-abbruch` so maybe check with your designers again?

Sure, I'll talk with them about it before to merge this PR

frederike-ramin (approved):
Nice, that's what I meant.

PR #943, opened 2024-06-24
feat: make feedback component in viewport after submitting data

pgurusinga:
### With JavaScript Enabled:
- When users submit helpful ratings, the viewport remains at the bottom of the page, which is the desired behaviour.
- However, when users submit text feedback, the page reloads, causing the viewport to jump to the top. As a result, users do not see the confirmation message inside the banner.

### Without JavaScript:
- Submitting either helpful ratings or text feedback causes the page to reload.
- This reload prevents users from seeing confirmation messages or understanding the next meaningful step.

## Proposed changes
To ensure functionality in both environments (with and without JavaScript):
- Add an id attribute to the relevant component.
- On button submission, redirect users to the page with a URL fragment corresponding to this id.
- Implement an end-to-end test to verify that the component remains in the viewport after submission.


app/services/feedback/handleFeedback.ts
```diff
@@ -39,8 +39,10 @@ export const handleFeedback = async (formData: FormData, request: Request) => {
     },
   });
   session.set(bannerStateName, bannerState);
+
   const headers = { "Set-Cookie": await commitSession(session) };
-  return redirect(pathname, { headers });
+  const userFeedbackPath = pathname + `#${USER_FEEDBACK_ID}`;
```
aaschlote, line 44:
```suggestion
  const userFeedbackPath = `${pathname}#${USER_FEEDBACK_ID}`;
```

app/components/UserFeedback/index.tsx
```diff
@@ -32,7 +34,11 @@ export default function UserFeedback(props: Readonly<UserFeedbackProps>) {
         overhangingBackground
         backgroundColor="midBlue"
       >
-        <div className="ds-stack-16" data-testid="user-feedback-banner">
+        <div
+          className="ds-stack-16"
+          data-testid={USER_FEEDBACK_ID}
+          id={USER_FEEDBACK_ID}
```
aaschlote, line 40: you could add a small unit test to check if this div contains an `id` with the value `USER_FEEDBACK_ID`.

aaschlote (approved):

app/components/__test__/UserFeedback.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+      {
+        path: "",
+        Component: () => (
+          <UserFeedback bannerState={BannerState.ShowRating} {...mockedProps} />
+        ),
+      },
+    ]);
+    render(<UserFeedbackWithRemixStub />);
+    const { id } = screen.getByTestId("user-feedback-banner");
+    expect(id).toBe("user-feedback-banner");
```
aaschlote, line 36: Maybe you can use the variable `USER_FEEDBACK_ID` to compare with your id, what do you think?

aaschlote (approved):

PR #950, opened 2024-06-26
Create user feedback component from Strapi

aaschlote:


app/components/UserFeedback/UserFeedbackContext.tsx
```diff
@@ -0,0 +1,13 @@
+import { createContext, useContext } from "react";
+import { BannerState } from ".";
+
+type UserFeedbackContext = {
+  bannerState?: BannerState;
+  flowId?: string;
```
chohner, line 6: I think `flowId` is the wrong variable name here, as it describes what exactly is being rated. Originally it was named `context` but maybe `pathname` is actually the correct word?

chohner: Also, could we type this as an optional object, instead of an object with two optional properties? so if the object is there, both properties can be accessed?

app/components/resultPage/DefaultResultPage.tsx
```diff
@@ -14,16 +16,21 @@ export function DefaultResultPage() {
     courts,
   } = useLoaderData<typeof loader>();
 
+  const userFeedbackContextValue = useMemo(
```
chohner, line 19: Could this be moved into `StrapiUserFeedback`?

app/components/resultPage/FluggastrechteResultPage.tsx
```diff
@@ -13,16 +15,21 @@ export function FluggastrechteResultPage() {
     amtsgerichtCommon,
   } = useLoaderData<typeof loader>();
 
+  const userFeedbackContextValue = useMemo(
```
chohner, line 18: this could also be left untouched if the hook is moved into `StrapiUserFeedback`

app/routes/shared/feedback.server.ts
```diff
@@ -0,0 +1,10 @@
+import { ActionFunctionArgs } from "@remix-run/node";
+import {
+  handleFeedback,
+  isFeedbackForm,
+} from "~/services/feedback/handleFeedback";
+
+export const action = async ({ request }: ActionFunctionArgs) => {
+  const formData = await request.formData();
+  if (isFeedbackForm(formData)) return handleFeedback(formData, request);
+};
```
chohner, line 10: I think the existing `/actions/send-feedback` pattern is much cleaner, as we don't need to repeatedly export the same action from all routes. also, combining actions might be a hassle in the future

chohner (changes requested):

app/components/resultPage/DefaultResultPage.tsx
aaschlote: I could, but I would need to pass the `bannerState` and `flowId` as props of the `PageContent` and pass it again to  `StrapiUserFeedback`. So using the context wouldn't make so much more difference. To have a lower level of abstraction and avoiding unneeded props I created this context, so I can access only I need it

app/routes/shared/feedback.server.ts
aaschlote: I don't fully understand what you mean here. 

I created this file to be exported in a few places where we don't handle the `feedback` in the action. In this case on the content pages.

I didn't find the existing one you mentioned `/actions/send-feedback`, only the `action/send-rating`. This what do you mean?

app/components/UserFeedback/UserFeedbackContext.tsx
aaschlote: I was strangling about the naming of this variable, because on the feedback layer is called context, but on the flow layer is `flowId`. Not sure what it's the best one, but for me when I get the `flowId` and pass to the props `context` is sounds a better weird. But I can change it if you prefer this name.

About the both object to be optional, the problem I was facing it's that the BannerState is an enum and it's defined the default value of the context when it's created and I have some errors to define it, for this reason I was forced to make it as optional, but it should be always with a value.

app/routes/shared/feedback.server.ts
chohner: Sorry for being unclear - yes I meant adding a new `actions/send-feedback` endpoint that handles the feedback action. The url would then be hard-coded in the `action` of that form

app/components/resultPage/DefaultResultPage.tsx
chohner: but I thought thats what the context is doing? or alternatively move the context one level up - I thought having 3+ times the same context looks a bit fishy

aaschlote: the another level up it's the root and inside of the root I don't have the information about the `flowId` 😞

app/components/UserFeedback/UserFeedbackContext.tsx
chohner: but what is called `flow id` here is just the name of what is being rated, right? if this component can live on any page this should be exclusively obtainable from the pathname

app/components/resultPage/DefaultResultPage.tsx
chohner: see above: I don't think we need a flow id - we just need to know what page is being rated

app/components/UserFeedback/UserFeedbackContext.tsx
aaschlote: As I mentioned in my comment above, on the flow layer context is `flowID` but on this component is called `context`. I didn't name them like this, you can see in the image and the following [link](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/app/components/resultPage/ResultPage.tsx#L181)

image

The `flowId` is optional in this component, so for this reason can be used in any page. I didn't make this change, it was already implemented like this.

If you feel the name in this react context should be `context` instead of `flowId` I can change, but it sounds a bit weird for me, but no harding feeling to change it.

app/routes/shared/feedback.server.ts
aaschlote: I see what you mean, I will check it

app/components/UserFeedback/UserFeedbackContext.tsx
aaschlote: so, you are saying this props `context` in the [RatingBox.tsx](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/app/components/UserFeedback/RatingBox.tsx#L42) is unless and I can get rid of it? Because we are using already the `pathname` in the props `url`.

image

app/routes/shared/feedback.server.ts
pgurusinga: I would honestly use different PR or commit for that since it's a different scope than just strapify the component

aaschlote: I found sometime yesterday and I implemented it :)

app/components/UserFeedback/FeedbackFormBox.tsx
```diff
@@ -68,18 +67,14 @@ export const FeedbackFormBox = ({ destination }: FeedbackBoxProps) => {
     translations,
   );
 
-  const nextDestination = jsAvailable
-    ? destination
-    : `${destination}#${USER_FEEDBACK_ID}`;
-
```
pgurusinga, line 74: why is this deleted? how then are we showing the user after submitting data?

app/components/UserFeedback/FeedbackFormBox.tsx
```diff
@@ -68,18 +67,14 @@ export const FeedbackFormBox = ({ destination }: FeedbackBoxProps) => {
...
   return (
     <>
       <Heading look="ds-label-01-bold" tagName="h2" text={heading} />
       <ValidatedForm
         validator={feedbackValidator}
         subaction={feedbackFormName}
         method="post"
-        action={nextDestination}
```
pgurusinga, line 82: also related to one above

app/root.tsx
```diff
@@ -122,6 +132,8 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
     deletionLabel: deleteDataStrings["footerLinkLabel"],
     hasAnyUserData,
     feedbackTranslations,
+    bannerState:
+      getFeedbackBannerState(mainSession, pathname) ?? BannerState.ShowRating,
```
pgurusinga, line 136: if this defaults to `BannerState.ShowRating` could you explain why `useUserFeedback()` in the component has default arguments with the same value? 
`const { bannerState = BannerState.ShowRating } = useUserFeedback();`

app/components/UserFeedback/FeedbackFormBox.tsx
aaschlote: I created a new route (`action.send-feedback`), the whole logic is there :)

app/root.tsx
aaschlote: I had a problem with the context `UserFeedbackContext`. I couldn't make the context have a default value due the type is an enum and it was throwing some errors on the page. So I was forced to make it as optional. When it's used the hook `useUserFeedback()` the value of the bannerState can be `undefined`, so again I have to pass a default value.

If you want we can pair programming on it and try to find a solution.

pgurusinga (approved):

app/routes/$.tsx
chohner, line 1: I think this could be fully reverted to decrease the diff?

app/routes/action.send-feedback.ts
```diff
@@ -0,0 +1,67 @@
...
+      feedback: result.data?.feedback ?? "",
+      context,
+    },
+  });
+  session.set(bannerStateName, bannerState);
+
+  const headers = { "Set-Cookie": await commitSession(session) };
+  const userFeedbackPath = clientJavaScriptAvailable
+    ? url
+    : `${url}#${USER_FEEDBACK_ID}`;
```
chohner, line 65: what happens if we always redirect to `${url}#${USER_FEEDBACK_ID}`?

app/routes/action.send-feedback.ts
```diff
@@ -0,0 +1,67 @@
...
+  const context = getContextByUrl(url);
+
+  const cookie = request.headers.get("Cookie");
+  const { getSession, commitSession } = getSessionManager("main");
+  const session = await getSession(cookie);
+
+  const userRating =
+    (session.get(userRatingFieldname) as Record<string, boolean>) ?? {};
+  const bannerState =
+    (session.get(bannerStateName) as Record<string, BannerState>) ?? {};
```
chohner, line 44: there is probably a lot of data shared with `send-rating`... maybe for future cleanup

app/routes/action.send-feedback.ts
```diff
@@ -0,0 +1,67 @@
...
+    return flowId;
+  } catch (error: unknown) {
+    return "";
+  }
+};
+
+export const loader = () => redirect("/");
+
+export const action = async ({ request }: ActionFunctionArgs) => {
+  const formData = await request.clone().formData();
```
chohner, line 25: I think we dont need to clone here (anymore), since the data will only come from the dedicated form

app/routes/action.send-rating.ts
```diff
@@ -2,10 +2,20 @@ import type { ActionFunctionArgs } from "@remix-run/node";
 import { json, redirect } from "@remix-run/node";
 import { BannerState, USER_FEEDBACK_ID } from "~/components/UserFeedback";
 import { userRatingFieldname } from "~/components/UserFeedback/RatingBox";
+import { parsePathname } from "~/models/flows/flowIds";
 import { sendCustomAnalyticsEvent } from "~/services/analytics/customEvent";
-import { bannerStateName } from "~/services/feedback/handleFeedback";
+import { bannerStateName } from "~/services/feedback/getFeedbackBannerState";
 import { getSessionManager } from "~/services/session.server";
 
+const getContextByUrl = (url: string): string => {
```
chohner, line 10: maybe rather `parseFlowIdFromUrl`?

chohner (approved):
some nitpicks

app/routes/action.send-feedback.ts
pgurusinga: toThe `preventScrollReset` for the Y-Axis in the form will not work as intended; it is always positioned at the top. The design requires it to remain in place for JS users upon form submission.

app/routes/action.send-feedback.ts
pgurusinga: @aaschlote might be a TODO for Fixathon?

aaschlote: I'll add it

PR #951, opened 2024-06-26
Add translations user feedback

aaschlote:
### Add translations for the User feedback component

Today we have hardcode the texts/descriptions of the User feedback component. We are moving those texts/descriptions to Strapi Translations

### Proposal

- Move texts/descriptions to Strapi Translation
- Create a context with the feedback translations
- Use hook to get the feedback translations without pass them via props

### Out of scope
Later we will have this component also in Strapi, so using the context and hook it would be easier to make the component following the Product and UX requirements

### DIFF
--- a/app/components/UserFeedback/__test__/PostSubmissionBox.test.tsx
+++ b/app/components/UserFeedback/__test__/PostSubmissionBox.test.tsx
@@ -1,13 +1,17 @@
 import { render } from "@testing-library/react";
 import { FeedbackTranslationContext } from "../FeedbackTranslationContext";
-import { PostSubmissionBox } from "../PostSubmissionBox";
+import {
+  HEADING_POST_SUBMISSION_TRANSLATION_KEY,
+  PostSubmissionBox,
+  TEXT_POST_SUBMISSION_TRANSLATION_KEY,
+} from "../PostSubmissionBox";
 
 const HEADING_POST_SUBMISSION = "Heading";
 const TEXT_POST_SUBMISSION = "Text";
 
 const TRANSLATION_KEY_RECORD = {
-  "heading-post-submission": HEADING_POST_SUBMISSION,
-  "text-post-submission": TEXT_POST_SUBMISSION,
+  [HEADING_POST_SUBMISSION_TRANSLATION_KEY]: HEADING_POST_SUBMISSION,
+  [TEXT_POST_SUBMISSION_TRANSLATION_KEY]: TEXT_POST_SUBMISSION,
 };
 
 describe("PostSubmissionBox", () => {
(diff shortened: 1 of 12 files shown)


chohner:
I thought we agreed to make it a "classic" strapi page component that contains its own text. did the requirement change?

aaschlote:
> I thought we agreed to make it a "classic" strapi page component that contains its own text. did the requirement change?

Only the first question of the feedback can be changed via the page component, the other texts it should be translated, because it'll be same text. And it's a [ticket](https://app.asana.com/0/1207022303033556/1207469229502273) requirement

app/components/UserFeedback/__test__/PostSubmissionBox.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { render } from "@testing-library/react";
+import { FeedbackTranslationContext } from "../FeedbackTranslationContext";
+import { PostSubmissionBox } from "../PostSubmissionBox";
+
+const HEADING_POST_SUBMISSION = "Heading";
+const TEXT_POST_SUBMISSION = "Text";
+
+const TRANSLATION_KEY_RECORD = {
+  "heading-post-submission": HEADING_POST_SUBMISSION,
```
pgurusinga, line 9: is it possible to do like this? if not we can ignore it
```suggestion
  `[HEADING_POST_SUBMISSION_TRANSLATION_KEY]: HEADING_POST_SUBMISSION,
```

https://www.freecodecamp.org/news/how-to-set-dynamic-object-properties-using-computed-property-names/#:~:text=How%20to%20Set%20Variables%20as%20Property%20Names

app/components/UserFeedback/__test__/RatingBox.test.tsx
```diff
@@ -0,0 +1,33 @@
+import { createRemixStub } from "@remix-run/testing";
+import { render } from "@testing-library/react";
+import { FeedbackTranslationContext } from "../FeedbackTranslationContext";
+import { RatingBox } from "../RatingBox";
+
+const YES_RATING = "yes";
+const NO_RATING = "no";
+
+const TRANSLATION_KEY_RECORD = {
+  "yes-rating": YES_RATING,
```
pgurusinga, line 10: also can it be constructed with dynamic object property? 
```suggestion
  [YES_RATING_BUTTON_LABEL_TRANSLATION_KEY]: YES_RATING,
```

pgurusinga (approved):
beside my comment

app/components/UserFeedback/__test__/PostSubmissionBox.test.tsx
aaschlote: done!

PR #956, opened 2024-06-28
Docs: Write adr 0008-staging-environment-public

aaschlote:


doc/adr/0008-staging-environment-public.md
```diff
@@ -0,0 +1,24 @@
...
+
+Date: 2024-06-28
+
+## Status
+
+Agreed
+
+## Context
+
+Previously, accessing the staging environment required user and password authentication. However, this posed issues for external users who needed access, leading to the decision to remove authentication and make the staging environment public.
```
pgurusinga, line 11: I think:
1. could we make it clearer that it was HTTP basic auth?
2. can we add more details, that the issue was some users cannot even enter the credentials with HTTP basic auth?

doc/adr/0008-staging-environment-public.md
```diff
@@ -0,0 +1,24 @@
...
+
+Agreed
+
+## Context
+
+Previously, accessing the staging environment required HTTP basic authentication. However, this posed issues for external users who needed access due problems with their devices or browsers to enter the credetials with HTTP basic authentication, leading to the decision to remove authentication and make the staging environment public.
+
+### Issue
+
+With ADR 0005, we created the preview environment, eliminating the need to share the staging environment with external users. However, the staging environment remains public, making it accessible to everyone. This public access could result in interruptions due to issues with the Strapi API, potentially disrupting our development workflow.
```
chohner, line 15: I don't think I agree with the last sentence, as I think public access to staging cannot lead to disrupting our workflow. But maybe I am not following your reasoning here, are you thinking of malicious attacks?

The only issue I see is leaking non-public data

aaschlote: Yes, this was problem I mentioned during the `Dev Exchange`. As the staging environment connects direct with Strapi via API and this environment is public, the staging can fail in case we get too many requests and this might stop the workflow whom depends of this environment, like us and the UX and product.

chohner: Ok sounds fair. Maybe its easiest to list the two things that can go wrong:

- DDOS against staging will impact our ability to continue working (mitigated by OTC load balancer & DDOS protection)
- Leaking of non-public information (mitigated by generally working publicly and knowledge about staging being open)

pgurusinga (approved):

PR #957, opened 2024-07-01
Remove feedback banner component and single entry global

aaschlote:


chohner (approved):
nice. could you also add `single entry `global` into the PR title?

tests/e2e/pom/CookieSettings.ts
```diff
@@ -38,6 +38,12 @@ export class CookieSettings {
     await this.submitForm();
   }
 
+  async acceptCookieBanner() {
+    // wait 4 seconds before accept cookie banner
+    await this.page.waitForTimeout(4000);
```
chohner, line 43: I'm a bit surprised by this - why is this needed now? Especially with a 4s timeout?

aaschlote: Once I removed the feedback banner component, the cookie banner was hidden the next button when playwright was trying to click on it. So I was forced to accept the banner for every page where this error was appearing to me. 

The 4s timeout is due the `useEffect` on the cookie banner component, because when the page is loaded it's set the click button to load without javascript and later come the `useEffect` and change it. So the page sometimes is loading so fast in playwright and the click is happening to fast as well, that the action on click to accept the banner was loading the cookie banner page for `non-javascript`. I added 4s because we are having some flaking in my previous tests.

chohner: hm thats very unfortunate. Could you add this context to the code? otherwise its impossible to re-trace.
Also, it would be cool to not need this but I guess not in this PR. maybe add a `TODO`?

aaschlote: I can write a comment in the source about it.

About the changes in this PR, I need it, because since I removed the component, I got this error, without it, the pipeline fails 😞

PR #975, opened 2024-07-10
Add lazy loading for StrapiAutoSuggestion

aaschlote:
### Problem

Current our bundle is very large due the airports and airlines data. This cause performance issues on mobile devices.

### Solution

Use [React lazy loading](https://react.dev/reference/react/lazy) to load the StrapiAutoSuggestion component during runtime and avoid unneeded data in the bundle.


frederike-ramin:
> During the E2E Test, we are forced to not load the component StrapiSelect as lazy, due issues running the tests (as this component has been used many times, somehow it was failing to load it during the runtime)

Does this mean that if we use more and more input fields, the same problem might arise there?

aaschlote:
> > During the E2E Test, we are forced to not load the component StrapiSelect as lazy, due issues running the tests (as this component has been used many times, somehow it was failing to load it during the runtime)
> 
> Does this mean that if we use more and more input fields, the same problem might arise there?

I don't know, but in the Beratungshilfe formular it has a lot of input fields, but it didn't cause this issue 😞 

I would like to try this approach in staging and see if we can reproduce this error.

PR #979, opened 2024-07-10
Solution sorting airports

aaschlote:


app/components/inputs/autoSuggestInput/AutoSuggestInput.tsx
```diff
@@ -95,6 +95,20 @@ const keyDownOnInput = (
   }
 };
 
+const getSortingAirportsByCode = (
+  filteredOptions: DataListOptions[],
+  inputCode: string,
+): DataListOptions[] => {
+  if (inputCode.length === AIRPORT_CODE_LENGTH) {
```
pgurusinga, line 102: do you mean `>==` since it will evaluate `value` then `subDescription` ?

app/components/inputs/autoSuggestInput/getDataListOptions.ts
```diff
@@ -20,7 +20,7 @@ export function getDataListOptions(
     }
     case "airports": {
       return [...airports]
-        .sort((a, b) => a.iata.localeCompare(b.iata))
+        .sort((a, b) => a.city.localeCompare(b.city))
```
pgurusinga, line 23: Why do we need these changes? I think the IATA comparison is much faster because it's guaranteed to compare 3 letters.

app/components/inputs/autoSuggestInput/AutoSuggestInput.tsx
aaschlote: we just need to order by the value only the user type 3 characters, because this is the size of the IATA code, otherwise we don't need to order by value (iata code)

app/components/inputs/autoSuggestInput/getDataListOptions.ts
aaschlote: this is the default value, we should order by the city as it's describe in the ticket.

So if the user enter 3 characters, we order by the iata code, otherwise by the city.

pgurusinga: btw now I just realized we can remove this if we decide to use match-sorter, otherwise we are sorting twice

aaschlote: yeap, but if I use the `match-sorter` it'll be sorted every time I type on the component

aaschlote: From this [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/990) I moved the items of the component to the serve, so the `match-sorter` is only called once.

pgurusinga (approved):

PR #985, opened 2024-07-11
Remove role button for ahref element and tests

aaschlote:


frederike-ramin:
Could you provide context on why you did that change? Just having a test-id is *by far* not the same as having a `role=button` so I imagine there's a reasoning behind this?

aaschlote:
> Could you provide context on why you did that change? Just having a test-id is _by far_ not the same as having a `role=button` so I imagine there's a reasoning behind this?

Sonar wasn't so happy having a `ahref` element with a `role=button`. It seems, we can use but it's not recommended (the recommendation from sonar is having a `button` HTML element. The component was like that, but it has been changed to `ahref` and I don't know why 🤔 ). So I have to remove this role and add a `test-id` for a unit tests and E2E Test.

frederike-ramin:
Okay, So, we resoled this in [this thread](https://github.com/digitalservicebund/a2j-rechtsantragstelle/commit/43710c9e7d59e06f304830cc7e6b92893e7c7aa1).

Still, I'm not quite satisfied with adding the `data-testid="custom-button"` on every single button. For the buttons we need in the e2e tests, we can add a meaningful ID, I believe, and that not on the component level but on the specific button. As far as I see it's really just the cookie banner button which we can configure right in the code (which btw already has a test-id, so we can just use that.

aaschlote:
> Maybe I didn't explain correctly: With "meaningful ID" I didn't mean the attribute `id` I'm aware that this has a very different purpose. I just meant a more meaningful attribute value. Such as `cookie-banner-accept` or similar.
> 
> On your question:
> 
> > replace the test by getByRole("link")
> 
> You mean the component test? Yes, that would be one option. Or just add the `data-testid` _in_ the test, not in the component itself. In our case though, we do not need a `data-testid` for any test. We can just delete the new tests for `data-testid` and the only one you had to adapt was deleted in the meantime anyways.

Ok, I got what you meant. I'll create a PR and ask to check the changes 😃

PR #986, opened 2024-07-12
feat(datalist): add datalist inside AutoSuggestInput

chohner:
Minimal support for datalist for users without JS


aaschlote:
I'll close this MR, because for the current implementation of component, we don't have an option for non-JS users

PR #990, opened 2024-07-12
Get data list options in the server

aaschlote:


app/components/inputs/autoSuggestInput/useListOptions.tsx
```diff
@@ -0,0 +1,27 @@
...
+        const response = await fetch(`${DATA_LIST_API_PATH}/${dataListType}`);
+        const json = await response.json();
+        setDataListOptions(json);
+      } catch (error) {
+        setDataListOptions([]);
+      }
+    };
+
+    fetchData();
+  }, [dataListType]);
```
pgurusinga, line 22: Why do we need `dataListType` as a dependency? We only need the data once in the beginning and `dataListType` won't change over time on the client side, except if users go to the next route (airline), which will re-request the backend. So, I think it can be an empty array instead.

app/components/inputs/autoSuggestInput/useListOptions.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const response = await fetch(`${DATA_LIST_API_PATH}/${dataListType}`);
+        const json = await response.json();
+        setDataListOptions(json);
+      } catch (error) {
+        setDataListOptions([]);
+      }
```
pgurusinga, line 18: we might need more check for the response, if you know that it's always 200 then we can ignore this.

```suggestion
      try {
        const response = await fetch(`${DATA_LIST_API_PATH}/${dataListType}`);
        if (response.ok) {
          const json = await response.json();
          setDataListOptions(json);
        }
      } catch (error) {
        setDataListOptions([]);
      }
```

app/routes/data-list-options.$TYPE.tsx
```diff
@@ -0,0 +1,9 @@
+import { LoaderFunctionArgs, json } from "@remix-run/node";
+import { getDataListOptions } from "~/components/inputs/autoSuggestInput/getDataListOptions";
+import { DataListType } from "~/services/cms/components/StrapiAutoSuggestInput";
+
+export async function loader({ params }: LoaderFunctionArgs) {
+  const dataListType = params.TYPE as DataListType["dataList"];
```
pgurusinga, line 6: If I send a random HTTP request to this endpoint from the command line interface and I set the $TYPE as an integer, who will handle the error? I believe we need runtime validation, specifying that the TYPE can only be a string and can either be 'airport' or 'airlines'.

app/components/inputs/autoSuggestInput/useListOptions.tsx
aaschlote: it's an eslint warning, this why I added it

app/components/inputs/autoSuggestInput/useListOptions.tsx
aaschlote: The changes are just a draft for now, I'll improve the code

app/routes/data-list-options.$TYPE.tsx
aaschlote: The changes are just a draft for now, I'll improve the code

pgurusinga (approved):

PR #1026, opened 2024-07-29
feat: lookup flow pages by flowId & stepId

chohner:
following https://github.com/digitalservicebund/a2j-rechtsantragstelle-strapi/pull/293, this PR allows querying for multiple fields and refactors the internal fetch interface to filter by `stepId` and `flowId`


app/services/cms/__test__/getStrapiEntryFromFile.test.ts
```diff
@@ -64,7 +64,7 @@ describe("services/cms", () => {
         expect(
           await getStrapiEntryFromFile({
             apiId: "pages",
-            filterValue: impressumPath,
+            filters: [{ field: "slug", value: impressumPath }],
```
frederike-ramin, line 67: Is there also a test for multiple filters?

frederike-ramin: Same for the Api case?

app/services/cms/getStrapiEntryFromFile.ts
```diff
@@ -27,16 +27,29 @@ export const getStrapiEntryFromFile = async ({
...
+        if (nestedFielname === undefined) {
+          return relevantField === value;
+        }
+
+        return (
+          typeof relevantField === "object" &&
+          Array.isArray(relevantField.data) &&
+          relevantField.data.some(
+            (nestedItem) => nestedItem.attributes[nestedFielname] === value,
+          )
```
frederike-ramin, line 50: What if the relevant field does not contain an array as `data`? I feel this needs a lot more testing edge cases?

app/services/cms/index.server.ts
```diff
@@ -1,32 +1,28 @@
...
 const getStrapiEntry =
   config().CMS === "FILE" ? getStrapiEntryFromFile : getStrapiEntryFromApi;
 
-export async function fetchMeta(opts: Omit<GetStrapiEntryOpts, "apiId">) {
+export async function fetchMeta(
+  opts: Omit<GetStrapiEntryOpts, "apiId" | "filter"> & { filterValue: string },
+) {
   const populate = "meta";
-  const pageEntry = await getStrapiEntry({ ...opts, apiId: "pages", populate });
+  const filters = [{ value: opts.filterValue, field: "slug" }];
```
frederike-ramin, line 23: This still relies on the slug? Do we want to keep it afterall? I thought the plan was to remove it?

app/services/cms/index.server.ts
```diff
@@ -94,7 +95,9 @@ export async function fetchErrors() {
   const cmsErrorSlug = "/error/";
 
   const errorPagePromises = httpErrorCodes.map((errorCode) =>
-    fetchCollectionEntry("pages", `${cmsErrorSlug}${errorCode}`),
+    fetchCollectionEntry("pages", [
+      { field: "slug", value: `${cmsErrorSlug}${errorCode}` },
+    ]),
```
frederike-ramin, line 100: For pages, we want to keep the slug? Or just a stepId without a flow to keep it consistent naming-wise? Or maybe that makes it less consistent? wdyt?

chohner: I think I prefer keeping `slug` but up for discussion

app/services/cms/getStrapiEntryFromFile.ts
chohner: Then `Array.isArray(relevantField.data)` would fail, right? `data` is standard naming by Strapi, the entry list looks like this:
```typescript
{
  flow_ids: {
    data: [
      {
        attributes: {
          flowId: "foo"
        }
      }
    ]
  }
}
```

app/services/cms/__test__/getStrapiEntryFromFile.test.ts
chohner: yes, just not commited 🙈

app/services/cms/getStrapiEntryFromFile.ts
frederike-ramin: So that means it always has this structure? What about nested fields that are an array? Or multiply-nested fields for that matter?
I think my problem with this is the function *looks* generic, but it still is very tailored to the two specific cases that we have? And I'm wondering if it would maybe make our lives easier if we just accounted for those two cases instead of building a fit-all filter? Or at the least - add a comment which cases are covered and which aren't?

app/services/cms/index.server.ts
frederike-ramin: No need for discussion, okay for me.

app/services/cms/getStrapiEntryFromFile.ts
chohner: The app itself sees quite narrow filters, depending on what you want to fetch (`fetchPage(slug: string)` for example).
All those requests then are translated to either a `qs` query string for the REST API or a lookup in the content file (+1 on documenting whats supported).

One alternative to a general `Filters[]` definition would be to introduce two types of filters that we currently support (`property` & `inArray`?) I'll check whether that makes it more or less confusing

chohner: I'm quite happy with introduction a single optional property, see [`d746597` (#1026)](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1026/commits/d746597365de88a9c74a5e243e7d712deeb1a6fd)

PR #1058, opened 2024-08-07
Add Standalone Link Component

Spencer6497:
My very first contribution 🎉 🥳 

This PR adds the Standalone Link component, as well as Storybook and tests.

https://app.asana.com/0/1203259475859667/1207931106219834/f

Demo: (Visually identical 😆 )
Bildschirmfoto 2024-08-07 um 15 02 23

Demo: (Left Icon added)
image

PR #1060, opened 2024-08-08
Use StandaloneLink component

aaschlote:
### Background

In our source code, we recreate many time an `a` HTML element with the same rules and className. In the [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1058) was created the component StandaloneLink, where basically render the same `a` HTML element. We also have the [ticket](https://app.asana.com/0/1207022303033556/1207820849591133) and [ticket](https://app.asana.com/0/1207022303033556/1207867713016633) to make some changes in the `a` HTML element of the website.
 
### Solution
Replace in the source code where the `a` HTML element are rendering for the component StandaloneLink. We have one centralised component that renders the `a` HTML element and make minimal effort for future changes.


app/components/RichText.tsx
```diff
@@ -24,9 +24,9 @@ const allowedAttributes = {
 
 const defaultRenderer: Partial<Renderer> = {
   link({ href, text }) {
-    const isExternal = isExternalUrl(href);
-    const isDownload = isFileDowloadUrl(href);
-    return `<a href="${href}" class="text-link" ${isExternal ? 'target="_blank" rel="noopener noreferrer"' : ""}>${text}${isExternal || isDownload ? openInNewIconText : ""}</a>`;
+    return ReactDOMServer.renderToString(
+      <StandaloneLink text={text} url={href} />,
+    );
```
frederike-ramin, line 29: This is the only place where I think the replacement is not correct. The term "StandaloneLink" I though was created for a reason, because it might be something different than "InlineLink". Currently, it does look the same, but I think the designers are thinking about this as different concepts, so maybe checkin with Marion?

aaschlote: I see what you mean, but in the RichText we can also have `StandaloneLink`, right? Like this one below

image

I got this from chatgpt

> use inline links when the linked content is closely tied to the surrounding text and standalone links when you need to provide a list of references or additional resources without integrating them into the narrative.

Spencer6497 (approved):
🚀

app/components/RichText.tsx
frederike-ramin: Either way is fine for me as long as we align it with the designers so that we're using the same words for everything.

aaschlote: I'll align with her 😃

PR #1061, opened 2024-08-08
Fix the npm task error

pgurusinga:


aaschlote:
What about to move the `verifyExternalLinks` from the `app` folder to `script` folder? I guess we don't need this file in the app, right? Or am I missing something?

app/services/cms/verifyExternalLinks.ts
```diff
@@ -47,23 +53,23 @@ function verifyExternalLinks(allowedList: Array<string>, regexPattern: RegExp) {
...
-        throw Error(
-          `${link} is not allowed. Please verify the link and add to the allowed list`,
-        );
     }
-  }
+  });
+}
+
+function extractLinks(content: string, regexPattern: RegExp) {
+  return content.match(regexPattern) || [];
```
aaschlote, line 68: My VSCode is showing this for me

image
 Can you check it?

app/services/cms/verifyExternalLinks.ts
```diff
@@ -47,23 +53,23 @@ function verifyExternalLinks(allowedList: Array<string>, regexPattern: RegExp) {
...
+  const foundLinks = extractLinks(content, regexPattern);
 
-  if (linksFromContent) {
-    for (const link of linksFromContent) {
-      // skip this for now
-      if (link == "https") continue;
-
-      const rejectedUrl = !allowedList.some((allowedLinks) =>
-        link.includes(allowedLinks),
+  foundLinks.forEach((link) => {
```
aaschlote, line 58: You can do the iteration in this list like this:

```
foundLinks
    .filter((link) => isLinkRejected(link, allowedList))
    .forEach((link) => {
      throw new Error(
        `${link} is not allowed. Please verify the link and add it to the allowed list`,
      );
    });
```

app/services/cms/verifyExternalLinks.ts
pgurusinga: I use `matchAll` now and found more URLs!

PR #1066, opened 2024-08-09
Add A11y accessibility to feedback forms 

aaschlote:


app/components/Heading.tsx
```diff
@@ -34,7 +36,7 @@ function Heading({
   }
 
   return (
-    <Tag data-testid={dataTestid} className={cssClasses}>
+    <Tag role={role} data-testid={dataTestid} className={cssClasses}>
```
pgurusinga, line 39: I don't know if this is the right way though, when there is NO role defined in Strapi, then the DOM would look like `role={null}` right? 
can we conditionally add the attribute based on the role value?

app/components/inputs/InputLabel.tsx
```diff
@@ -1,13 +1,14 @@
...
   readonly id: string;
   readonly classname?: string;
+  role?: AriaRole;
 }>;
 
-const InputLabel = ({ id, children, classname }: InputLabelProps) => {
+const InputLabel = ({ id, children, classname, role }: InputLabelProps) => {
   return (
-    <label className={classname} htmlFor={id}>
+    <label role={role} className={classname} htmlFor={id}>
```
pgurusinga, line 11: I guess here applies same comment as before

app/components/Heading.tsx
aaschlote: actually the role when it's not set a value it would be `undefined` so the attribute doesn't render on the page. Example below:

image

app/components/inputs/InputLabel.tsx
aaschlote: same as before, value is undefined, it doesn't render on the DOM

app/components/UserFeedback/FeedbackFormBox.tsx
```diff
@@ -69,16 +68,21 @@ export const FeedbackFormBox = ({ destination }: FeedbackBoxProps) => {
...
       >
-        <div className="ds-stack-16">
-          <Textarea name={feedbackFieldname} placeholder={placeholder} />
+        <div role="status" className="ds-stack-16">
+          <Textarea
+            name={feedbackFieldname}
+            label={heading}
+            classNameLabel="ds-label-01-bold"
+            placeholder={placeholder}
+            role="status"
```
pgurusinga, line 84: why hardcode the textarea role as "status" ? 🤔

aaschlote: To show the user the value was changed, more info this [link](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22).

But I need to check with Marion if this is correct, not sure if I added in the right place.

PR #1071, opened 2024-08-12
refactor(featureFlag): Use feature flags in every route, not catch-all

frederike-ramin:
Before, we only ever used feature flags in combination with a 404. This PR does two things:

1. Separate Feature Flag logic from throwing-404 logic
2. Remove the dependency of writing down which routes are covered by a certain feature flag ([which caused problems in the past](https://github.com/digitalservicebund/a2j-rechtsantragstelle/commit/ad787ad1210e30213fb8522d276a9b3a3fa2e464))

That means, we now use the remix routing + feature flags to decide which route is shown and which is not.

Please double-check that I correctly imported the step loaders and set the correct feature flags on each route


app/services/errorPages/throw404.ts
```diff
@@ -1,37 +1,15 @@
...
-      throw new Response(null, { status: 404 });
-    }
-
-    if (
-      url.includes("/prozesskostenhilfe/antrag") &&
-      !(await client.isFeatureEnabled("showProzesskostenhilfeFlow", phCallerId))
-    ) {
-      throw new Response(null, { status: 404 });
-    }
+export const throw404IfFeatureFlagEnabled = async (
```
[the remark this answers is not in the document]
frederike-ramin: You are absolutely right. 👌

app/services/featureFlags/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { config } from "~/services/env/web";
+
+export type FeatureFlag =
+  | "showFluggastrechteFormular"
+  | "showGeldEinklagenFlow"
+  | "showProzesskostenhilfeFlow";
+
+export const isFeatureFlagEnabled = async (featureFlag: FeatureFlag) => {
+  const { POSTHOG_API_KEY, POSTHOG_API_HOST } = config();
+  const phCallerId = "backend";
```
aaschlote, line 11: can we rename with a better name?

``posthogDistinctId``

This is also could be in the global scope of this file

app/services/errorPages/throw404.ts
```diff
@@ -1,37 +1,15 @@
...
-    if (
-      url.includes("/prozesskostenhilfe/antrag") &&
-      !(await client.isFeatureEnabled("showProzesskostenhilfeFlow", phCallerId))
-    ) {
-      throw new Response(null, { status: 404 });
-    }
+export const throw404IfFeatureFlagDisabled = async (
+  featureFlag: FeatureFlag,
+) => {
+  if (!(await isFeatureFlagEnabled(featureFlag))) {
```
aaschlote, line 12: I got a bit confused to understand this ``if`` condition with ``!`` and so many parentheses, maybe you can implement something like this:

```
const isPageEnabled = await isFeatureFlagEnabled(featureFlag);
  if (!isPageEnabled) {
    throw new Response(null, { status: 404 });
  }
```

app/services/errorPages/throw404.ts
```diff
@@ -1,37 +1,15 @@
-import { PostHog } from "posthog-node";
 import { config } from "~/services/env/web";
+import { FeatureFlag, isFeatureFlagEnabled } from "~/services/featureFlags";
```
aaschlote, line 2: Can we add some unit tests on this folder?

app/routes/geld-einklagen.vorabcheck.ergebnis.$.tsx
```diff
@@ -1,2 +1,10 @@
-export { loader } from "~/routes/shared/result.server";
+import type { LoaderFunctionArgs } from "@remix-run/node";
+import { loader as stepLoader } from "~/routes/shared/result.server";
+import { throw404IfFeatureFlagDisabled } from "~/services/errorPages/throw404";
+
 export { DefaultResultPage as default } from "../components/resultPage/DefaultResultPage";
```
aaschlote, line 5:
```suggestion
export { DefaultResultPage as default } from "~/components/resultPage/DefaultResultPage";
```

aaschlote (commented):
Nice job, I reviewed the pages, it sounds everything is correct! 🚀 

I wrote a few minor comments.

app/services/errorPages/__test__/throw404.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("should throw a 404 Response in production environment", () => {
+    // @ts-ignore
```
chohner, line 29: autschi

app/services/errorPages/throw404.ts
```diff
@@ -1,37 +1,16 @@
...
-      url.includes("/prozesskostenhilfe/antrag") &&
-      !(await client.isFeatureEnabled("showProzesskostenhilfeFlow", phCallerId))
-    ) {
-      throw new Response(null, { status: 404 });
-    }
+export const throw404IfFeatureFlagDisabled = async (
+  featureFlag: FeatureFlag,
+) => {
+  const isPageEnabled = await isFeatureFlagEnabled(featureFlag);
+  if (!isPageEnabled) {
```
chohner, line 13: nit: I feel like there might be one too many logic inversions going on here

chohner (approved):

aaschlote (approved):

PR #1083, opened 2024-08-14
Add minimal-height for standalone component

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1207820849591133

### DIFF
--- a/app/components/RichText.tsx
+++ b/app/components/RichText.tsx
@@ -63,7 +63,7 @@ const RichText = ({
           allowedTags,
           allowedClasses: {
             p: ["ds-subhead", "max-w-full"],
-            a: ["text-link"],
+            a: ["text-link", "min-h-[24px]", "inline-block"],
             h: ["ds-heading-01-reg", "ds-label-01-bold", "ds-heading-02-reg"],
           },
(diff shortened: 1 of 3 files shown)


app/components/RichText.tsx
```diff
@@ -63,7 +63,7 @@ const RichText = ({
           allowedTags,
           allowedClasses: {
             p: ["ds-subhead", "max-w-full"],
-            a: ["text-link"],
+            a: ["text-link", "min-h-[24px]", "inline-block"],
```
chohner, line 66: It might make sense to just import the classes from `StandaloneLink`, since they always will be allowed, right?

aaschlote: You can add HTML on the RichtText component in Strapi, like `<a>` and this prevent someone to add no-allowed classes for this attribute. I believe this was the idea when this method was built it.

chohner: Yes, I thought it might be easier to import the default classes from `StandaloneLink` here, instead of manually keeping them in sync

aaschlote: I see, I can check it

PR #1084, opened 2024-08-14
Embedded Video Component

Spencer6497:
This PR adds the frontend Embedded Video component, to be used for the Gebaerdensprache page.
Currently only supports YouTube videos, TBD whether or not we should extend it for other media types.
Strapi PR to follow

- **add Video component with Datenschutz banner, whitelist youtube csp links**
- **add StrapiVideo schema/translation layer, passthrough datenshutz rich text**
- **add strapi layer for video component**
- **add unit tests and Story, remove translations**

---
  - https://app.asana.com/0/0/1207888263381134


app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <DatenschutzBanner
+          onCookiesAccepted={() => setCookiesAccepted(true)}
+          {...{ datenschutz }}
+        />
+      )}
+    </>
+  );
+};
+
+const DatenschutzBanner = ({
```
aaschlote, line 56: can we move this component to his own file and have his own test? We can create a folder called video, so it would be:

``video/Video
video/DatenschutzBanner``

app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  useEffect(() => {
+    if (thumbnailRef.current) {
+      setThumbnailDimensions(thumbnailRef.current.getBoundingClientRect());
+    }
+  }, []);
+
+  return (
+    <>
+      {cookiesAccepted ? video : thumbnail}
```
aaschlote, line 45: how this component should work for non-Javascript users? I didn't see any comments in the ticket about it.

Or do we iterate this later?

app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
...
+    <img
+      ref={thumbnailRef}
+      alt={`${title} Miniaturbild`}
+      className="w-full opacity-60"
+      src={`https://img.youtube.com/vi/${ytVideoId}/hqdefault.jpg`}
+    ></img>
+  );
+  const video = (
+    <iframe
+      src={`https://www.youtube-nocookie.com/embed/${ytVideoId}?cc_load_policy=1&cc_lang_pref=en`}
```
aaschlote, line 27: the query parameter `cc_lang_pref` should be not in german?

app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
...
+  datenschutz: RichTextProps;
+}) => {
+  return (
+    <section
+      className="border-2 border-blue-800 z-50 bg-blue-300 relative -mt-24"
+      aria-label="Datenschutz banner"
+      data-testid="datenschutz-banner"
+    >
+      <div className="p-16 gap-y-28 flex flex-col flex-wrap">
+        <Heading text="Hinweis zum Datenschutz" look="ds-heading-03-reg" />
```
aaschlote, line 70: Maybe add a `TODO` here to use Strapi translations.

Me and @pgurusinga we will build a context to get the translations for all the component under the root level

app/services/cms/models/StrapiVideo.ts
```diff
@@ -0,0 +1,31 @@
...
+import { HasStrapiTimestampsSchema } from "~/services/cms/models/HasStrapiTimestamps";
+
+const StrapiVideoSchema = z
+  .object({
+    title: z.string(),
+    url: z.string(),
+    datenschutz: RichTextPropsSchema,
+  })
+  .merge(HasOptionalStrapiIdSchema)
+  .merge(HasStrapiLocaleSchema)
```
aaschlote, line 15: Do we need this schema?

app/services/cms/models/StrapiVideo.ts
```diff
@@ -0,0 +1,31 @@
...
+  .merge(HasStrapiMetaSchema)
```
aaschlote, line 16: Do we need this schema?

app/services/cms/models/StrapiVideo.ts
```diff
@@ -0,0 +1,31 @@
...
+  .merge(HasStrapiTimestampsSchema);
```
aaschlote, line 17: Do we need this schema

app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
...
+  datenschutz,
+}: {
+  onCookiesAccepted: () => void;
+  datenschutz: RichTextProps;
+}) => {
+  return (
+    <section
+      className="border-2 border-blue-800 z-50 bg-blue-300 relative -mt-24"
+      aria-label="Datenschutz banner"
+      data-testid="datenschutz-banner"
```
aaschlote, line 67: are u using this `data-testid` for any test?

app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
+import { useEffect, useRef, useState } from "react";
+import Button from "~/components/Button";
+import Heading from "~/components/Heading";
+import RichText, { RichTextProps } from "~/components/RichText";
+
+type VideoProps = {
+  title: string;
+  url: string;
+  datenschutz: RichTextProps;
```
aaschlote, line 9: I would prefer this attribute in `english`, what do you think? The other fields are also in english

app/components/Video.tsx
aaschlote, line 45: I would do this condition something like this:

```
 <>
      {cookiesAccepted ? (
        video
      ) : (
        <>
          {thumbnail}
          <DatenschutzBanner
            onCookiesAccepted={() => setCookiesAccepted(true)}
            {...{ datenschutz }}
          />
        </>
      )}
    </>
```

Just to avoid two statements for the same variable. But if you wanna keep as you did, I'm fine with that.

app/components/Video.tsx
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }, []);
+
+  return (
+    <>
+      {cookiesAccepted ? video : thumbnail}
+      {!cookiesAccepted && (
+        <DatenschutzBanner
+          onCookiesAccepted={() => setCookiesAccepted(true)}
+          {...{ datenschutz }}
```
aaschlote, line 49: why are you spreading this props, can we not make like this?

``
datenschutz={datenschutz}
``

aaschlote (commented):
Nice job, just got curios why the component is not in Strapi yet and how we handle for non-Javascript users

app/components/Video.tsx
chohner: I don't think one can watch youtube videos without JS :/

chohner:
Could we somehow enforce `https://www.youtube-nocookie.com/embed`? Or rather leave this to the CMS editors?

app/components/Video.tsx
aaschlote: so this feature is only for JS users?

app/services/cms/models/StrapiVideo.ts
chohner: all the additional ones are probably there for legacy reasons. Lets discuss in larger round but I think I agree in that we should remove them everywhere

app/components/Video.tsx
Spencer6497: ah sorry, we can do that. it's a habit I picked up in my last role, one of the engineers thought it was less redundant 😅

aaschlote: Normally I avoid to use spreading if I don't need to remove or replace one attribute from a object or set very attributes individually (like we do from StrapiComponent to the real Components), because when we spread an object it's created a new position in the memory, for small objects like this it's fine, but huge objects it might have some memory leak issues.

More about it you can read in this [article](https://medium.com/@joaopaulocmarra/why-spread-operators-in-javascript-might-surprise-you-with-slowness-and-when-to-pay-attention-bbb047ce9a22).

app/components/Video/Video.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <Container>
+      <div className="flex flex-col">
+        {cookiesAccepted ? (
+          <Video />
+        ) : (
+          <>
+            <Thumbnail />
+            <DataProtectionBanner
+              onCookiesAccepted={() => setCookiesAccepted(true)}
```
aaschlote, line 66: Once we had a problem in the component `AutoSuggestionInput` because we were re-rendering a component with arrow function and I got this from chatgtp:

```
Arrow Function ((props) => CustomControl(props)): Introduces an additional layer of function call, which can be useful for adding pre-processing logic.

Direct Assignment (CustomControl): More straightforward and efficient when no additional logic is required.
```

In this case to avoid re-rendering, you can do something like this:

```
const acceptCookie = useCallback(() => {
    setCookiesAccepted(true
  }, []);

 <DataProtectionBanner
              onCookiesAccepted={acceptCookie}
/>
```

Almost sure you won't face any issue as we had in the component `AutoSuggestionInput`, so if you wanna keep like this all good for me 😃

aaschlote (approved):
LGTM

app/components/CookieBanner/CookieConsentContext.tsx
```diff
@@ -0,0 +1,13 @@
+import { createContext, useContext } from "react";
+
+type CookieConsentContext = {
+  hasTrackingConsent?: boolean;
```
chohner, line 4: nit, maybe in the future: I wonder whether we should rename this to something less aggressive like `acceptedThirdPartyCookies` or similar

app/components/Video/Video.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [cookiesAccepted, setCookiesAccepted] = useState(hasTrackingConsent);
+  const [thumbnailDimensions, setThumbnailDimensions] =
+    useState<Partial<DOMRect>>(DEFAULT_SIZE);
+  const thumbnailRef = useRef<HTMLImageElement | null>(null);
+  const ytVideoId = getYoutubeVideoId(url);
+
+  const Thumbnail = () => (
+    <img
+      ref={thumbnailRef}
+      alt={`${title} Miniaturbild`}
```
chohner, line 32: I think `Vorschau für ${title}` might be more appropriate but technically this also belongs in translations :/

app/root.tsx
```diff
@@ -176,28 +182,31 @@ function App() {
...
-        <footer>
-          <Footer
-            {...footer}
-            deletionLabel={deletionLabel}
-            showDeletionBanner={hasAnyUserData}
-          />
-        </footer>
-        <ScrollRestoration nonce={nonce} />
-        <Scripts nonce={nonce} />
+        <CookieConsentContext.Provider value={{ hasTrackingConsent }}>
```
chohner, line 185: I am slightly torn regarding the scope of `CookieConsentContext.Provider`. On the one hand I feel like we should keep it small (so ideally only `CookieBanner` & `main`) on the other hand, wrapping all children of `body` feels like a less arbitrary choice.
Imo lets leave it and see how this looks in the future

chohner (approved):
lgtm!

PR #1085, opened 2024-08-15
fetch individual translations collections for arrays

rbrtrfl:
the idea is to subdivide the /beratungshilfe/antrag Translations collection in strapi into `arrayLabels` as well as `kinder`, `kraftfahrzeuge`, `geldanlagen` etc.


app/routes/shared/formular.server.ts
```diff
@@ -91,16 +92,33 @@ export const loader = async ({
...
+
+  const arraySummaryData = getSummaryData(
+    arrayCategories,
+    arrayConfigurations,
+    userData,
+  );
+
+  const flowStrings = arrayConfigurations
+    ? await fetchTranslations(flowId)
+    : await getArrayCategoryTranslations(arrayCategories);
```
chohner, line 120: I think you could move this logic above the `Promise.all()` to keep a single await?

Also might make sense to refactor this into a function?

rbrtrfl: I think it can't be moved to the `Promise.all()`, because `arrayCategories` needs `formPageContent` to be declared, which only happens in the `Promise.all()`

app/services/array/getArrayCategoryTranslations.ts
```diff
@@ -0,0 +1,12 @@
+import { fetchTranslations, Translations } from "../cms/index.server";
+
+export async function getArrayCategoryTranslations(
```
aaschlote, line 3: Can we add test for it?

aaschlote, line 3: The name of the function is `getArrayCategoryTranslations` but you fetching translations of `arrayLabels`, so the function doesn't do what it's described on this name. 

Can we find a better name for this function?

app/routes/shared/formular.server.ts
```diff
@@ -91,16 +92,33 @@ export const loader = async ({
   const [
     formPageContent,
     parentMeta,
-    flowStrings,
+    // flowStrings,
```
aaschlote, line 95: any reason to keep this line with `//``?

app/routes/shared/formular.server.ts
```diff
@@ -91,16 +92,33 @@ export const loader = async ({
...
     navigationStrings,
     defaultStrings,
   ] = await Promise.all([
     fetchFlowPage("form-flow-pages", flowId, stepId),
     fetchMeta({ filterValue: parentFromParams(pathname, params) }),
-    fetchTranslations(flowId),
+    // fetchTranslations(flowId),
```
aaschlote, line 101: Same here

app/routes/shared/formular.server.ts
```diff
@@ -91,16 +92,33 @@ export const loader = async ({
...
+    .filter(isStrapiArraySummary)
+    .map((strapiSummary) => strapiSummary.category);
+
+  const arraySummaryData = getSummaryData(
+    arrayCategories,
+    arrayConfigurations,
+    userData,
+  );
+
+  const flowStrings = arrayConfigurations
```
aaschlote, line 118: can we rename this variable? We are always fetching translations but passing the value for a variable called ``flowStrings``. What do you think?

aaschlote (commented):
Nice work, just leave a few comments for improvements. Reach me in case you need some better clarification 😃

app/services/array/getArrayCategoryTranslations.ts
rbrtrfl: ok

app/routes/shared/formular.server.ts
rbrtrfl: wip - deleted!

PR #1092, opened 2024-08-16
Add aria-labeledby in icon

pgurusinga:
Add text on the external standalone link icon 

next steps:
- fetch data from Strapi for custom aria-labelledby text


app/components/OpenInNewTabIcon.tsx
```diff
@@ -1,8 +1,13 @@
 import OpenInNewIcon from "@digitalservicebund/icons/OpenInNew";
 
+type OpenInNewTabIcon = {
```
aaschlote, line 3: can we rename to ``OpenInNewTabIcon`` to ``OpenInNewTabIconProps``? The react component of this file is called also ``OpenInNewTabIcon``?

app/components/OpenInNewTabIcon.tsx
```diff
@@ -15,10 +20,11 @@ export const openInNewAllowedAttributes = {
   path: ["d", "fill"],
 };
 
-export const OpenInNewTabIcon = () =>
+export const OpenInNewTabIcon = ({ ariaLabelledBy }: OpenInNewTabIcon) =>
```
aaschlote, line 23: Is it possible to add a small unit test here?

app/components/StandaloneLink.tsx
```diff
@@ -34,7 +34,7 @@ export const StandaloneLink = ({
     <a {...anchorProps}>
       {icon}
       {text}
-      {shouldOpenNewTab && <OpenInNewTabIcon />}
+      {shouldOpenNewTab && <OpenInNewTabIcon ariaLabelledBy={text} />}
```
aaschlote, line 37: Should be the ``öffnet neues Fenster`` as it's described in the [ticket](https://app.asana.com/0/1207022303033556/1207867713016633)?

aaschlote (approved):
LGTM

app/components/StandaloneLink.tsx
```diff
@@ -34,7 +34,9 @@ export const StandaloneLink = ({
     <a {...anchorProps}>
       {icon}
       {text}
-      {shouldOpenNewTab && <OpenInNewTabIcon />}
+      {shouldOpenNewTab && (
+        <OpenInNewTabIcon ariaLabelledBy={"In neuem tab öffnen"} />
```
chohner, line 38: I don't think this is the right usage of `aria-labledby`, as its used to reference an existing element ID in them DOM that contains the label, see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

pgurusinga: Thanks for the hint! we changed the implementation and tested using the aria-label in the `<a>`, but unfortunately, the voice assistive cannot speak the aria-label inside SVG.

cc: @aaschlote

chohner (approved):

PR #1094, opened 2024-08-19
Refactor: Unify File & API schemas

chohner:
We had a separate model for the content file (`StrapiFileContent.ts`) and the API, even though the file is filled using the API layer.

This unifies them into `StrapiSchemas` and cleans up typing throughout.
Downside:
- typing of `fetchX()` return values is a bit ugly (since we want to unpack content of the first item)


chohner:
replaced by https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1095

PR #1074, opened 2024-08-13
feat(pruner): Add config based pruner

zechmeister:
This introduces a pruning service that _prunes_/filters/reduces unnecessary properties from userData (`Context`) by combining the flow configuration and flowPages from strapi.

It does so in three steps:
1. use userData and the flowConfiguration and follow the Paths a user has taken through the flow to collect all steps a user has seen (`function getPaths(...): StepPath[]`). The steps are organised into lists of steps (`Paths`) because subflows and arrays need to be treated differently.
2. get the associated formFields from strapi for every step from 1.
3. pick the data from userData that has representations in the formFields from 2.

## todo:
- [x] fetch the flowPages only once (relevant for performance on staging)
- [ ] ~~e2e tests~~ integration tests
- [ ] move `getSteps()` to FlowController

PR #1095, opened 2024-08-19
refactor: unify file&api schemas

chohner:
- Introduces an shared interface `GetStrapiEntry` for data request layer
- This layer now always returns an array
- Unpacking of data moved to internal fetch interface (`fetchX()`)
- Unifies schema between file (`StrapiFileContentSchema`) and API (`schemas.ts`) to 
 a single `StrapiSchemas` and cleans up typing throughout
- removes `locale` from the internal API (for now)

Downside:
- return types of `fetchX()` are a bit ugly (since we want to unpack content of the first item)

unblocks https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1074


app/services/cms/getStrapiEntry.ts
```diff
@@ -0,0 +1,12 @@
...
+import { getStrapiEntryFromApi } from "./getStrapiEntryFromApi";
+import { getStrapiEntryFromFile } from "./getStrapiEntryFromFile";
+import type { ApiId, StrapiSchemas } from "./schemas";
+import { config } from "../env/env.server";
+
+export type GetStrapiEntry = <T extends ApiId>(
+  opts: GetStrapiEntryOpts & { apiId: T },
+) => Promise<StrapiSchemas[T] | [null]>;
+
+export const getStrapiEntry: GetStrapiEntry =
```
frederike-ramin, line 11: I'm struggeling a little bit with the naming here. And not only here, but also in our types. Everything is called `Strapi`, even for the file version, which is just Strapi in the second look. So maybe we can rename all of the `Strapi` things into `Cms` things? Becuse the file can be considered a Cms, but not a Strapi?

app/services/cms/getStrapiEntryFromApi.ts
```diff
@@ -33,37 +33,29 @@ const buildUrl = ({
...
-    response = unpackResponse(await makeStrapiRequest(buildUrl(opts)));
-  }
-  return response?.attributes;
+  const stagingData =
+    opts.locale !== "all"
+      ? (await makeStrapiRequest<T>(stagingUrl)).data.data
+      : null;
+
+  const invalidStagingData =
+    !stagingData || (Array.isArray(stagingData) && stagingData.length === 0);
```
frederike-ramin, line 54: It's `stagingDataInvalid` right?

frederike-ramin: I think `_.isEmpty() ` should do exactly what you want here?

app/services/cms/getStrapiEntry.ts
chohner: Sorry, just getting to this now - since the file is just a dump of strapi content I think the name still makes sense? Because the data structure follows strapi, not our own.

If we would add another CMS we would keep the internal `getEntry` interface but could have a totally different data structure behind it

PR #1096, opened 2024-08-19
Enable no-store (no caching) in preview and staging

pgurusinga:
## Problem
We are encountering an issue where the cookie banner reappears even after it has been accepted or declined on a different page.

## Steps to Reproduce
1. The user visits [https://service.justiz.de/](https://service.justiz.de/) and doesn't accept the banner.
2. The user clicks to the next page (e.g. Fluggastrechte or Beratungshilfe).
3. The user then accepts the banner when prompted.
4. After that, the user uses the browser's back button.
5. The banner reappears.

This issue only occurs in the production build because the Dev server or Remix doesn't include the `Cache-Control` header.

## Investigation Result
The browser caches the previous page, so when the cookie banner is accepted on a new page, the previous page is retrieved from the cache, showing outdated information.

After reviewing the article from the Chrome team at [https://web.dev/articles/bfcache#minimize-no-store](https://web.dev/articles/bfcache#minimize-no-store), I've concluded that this **no-store header** causes the back button to **re-fetch outdated cached data**.

> Cache-Control: `no-store` is an HTTP header web servers can set on responses that instructs the browser not to store the response in any HTTP cache. It's used for resources containing sensitive user information, such as pages behind a login.

## Solution
- Set `Cache-Control: no-store` to the client

## Pros/Cons
**Pros:**
- No user data is cached (this is still an assumption).
- Eliminates stale page issues.

**Cons:**
- The header might impact performance, although the browser may smartly cache some JS and CSS files.

## Next Steps
- We plan to enable the header in Staging and Preview environments to observe potential issues.
- Duration of observation: 1-2 weeks.


aaschlote (approved):
LGTM

PR #1098, opened 2024-08-19
Write ADR 009 - Recording Fail Videos from E2E Tests

aaschlote:


chohner (approved):

frederike-ramin (approved):
Not sure about the `significantly reduced` as I never checked it, but if you say so^^

pgurusinga (approved):

PR #1100, opened 2024-08-20
Dynamic Hinweisblatt Content

Spencer6497:
This PR dynamically creates the Hinweisblatt for the user, conditionally checking for a printed/online application, and checking for valid Amtsgericht.

Output (left = printout, right = online)
Bildschirmfoto 2024-08-20 um 15 32 32

---
  - https://app.asana.com/0/0/1208032874339857


Spencer6497:
NOTE: probably "cleverer" than need be. I wanted to minimize code duplication and ensure ease of copy changes in the future. LMK if I should do something simpler. Also, how the heck do I test this?! I tried in [this file](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/301298282dcc0de3516c3da32627333a7d452a29/app/services/pdf/beratungshilfe/__test__/getBeratungshilfePdfFromContext.test.ts) but couldn't figure it out.

PR #1113, opened 2024-08-26
Feat: Define width options for Select component

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1208087966238690

### DIFF
--- a/app/components/inputs/Select.tsx
+++ b/app/components/inputs/Select.tsx
@@ -5,14 +5,24 @@ import { type ErrorMessageProps } from ".";
 import InputError from "./InputError";
 import InputLabel from "./InputLabel";
 
-type SelectProps = {
+export type SelectProps = {
   name: string;
   options: { value: string; text: string }[];
   label?: ReactNode;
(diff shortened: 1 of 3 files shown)


app/components/inputs/Select.tsx
```diff
@@ -5,29 +5,44 @@ import { type ErrorMessageProps } from ".";
...
 
-  const selectClassName = classNames("ds-select", {
-    "has-error": error,
-  });
+  const selectClassName = classNames(
+    "ds-select",
+    {
+      "has-error": error,
+    },
+    width && widthClass(width),
```
chohner, line 44: Not sure whether its more readable but at least it would keep the whole logic in one place
```
classnames("ds-select", {
   "w-[22ch]": width === "16",
   "w-[30ch]": width === "24",
})
```

aaschlote: I got this logic from the `Input.tsx`. If you have strong feelings that it's better this way I can change it 😃

PR #1122, opened 2024-08-28
Create hidden input component

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1208148935193352

### DIFF
--- a/app/components/inputs/__test__/HiddenInput.test.tsx
+++ b/app/components/inputs/__test__/HiddenInput.test.tsx
@@ -1,0 +1,35 @@
+import { createRemixStub } from "@remix-run/testing";
+import { render } from "@testing-library/react";
+import * as remixValidatedForm from "remix-validated-form";
+import HiddenInput from "../HiddenInput";
+
+vi.mock("remix-validated-form", () => ({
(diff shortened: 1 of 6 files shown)


app/components/inputs/__test__/HiddenInput.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    clearError: vi.fn(),
+    validate: vi.fn(),
+    touched: false,
+    setTouched: vi.fn(),
+  });
+});
+
+describe("HiddenInput", () => {
+  it("should render an input hidden", () => {
+    const RemixStub = createRemixStub([
```
pgurusinga, line 24: Can we simplify this? there's no dependency on Remix so we could use `render(<HiddenInput />)` directly for example is the `Button.test.tsx`

pgurusinga (approved):
beside my comment

app/components/inputs/__test__/HiddenInput.test.tsx
aaschlote: sure, I didn't know it would work without. Thanks!

PR #1123, opened 2024-08-28
feat(ci): Disable preview deployment + test in pipeline

frederike-ramin:
We will be running external tests on our preview environment in the coming week so it should be stable.
However, we still want to deploy to production in the meantime.

This PR therefore disables the preview deployment along with the then unneeded (and misleading) e2e tests against preview.

*Do not merge this before 30.09.2024*
*Attention: Our pipeline will be more shaky during this time, so be even more careful what you push to production and possibly double-check production more often.*
We will be reverting this straight after the test ends.


aaschlote (changes requested):
It's missing to remove the`deploy-preview` in the build.content.yml file. Or do we want to have content deployment in preview?

aaschlote:
Sorry, I made an empty commit here, but it was in the wrong branch 😆

frederike-ramin:
@aaschlote nice find, thank you!

aaschlote (approved):
LGTM

PR #1125, opened 2024-08-29
Add flugdaten done funcs

pgurusinga:
I am not sure if I am doing the correct done funcs for the flugdaten, maybe we could simplify these funcs


app/flows/fluggastrechteFormular/flugdaten/doneFunctions.ts
```diff
@@ -0,0 +1,276 @@
...
+      context.zwischenstoppAnzahl !== "no" &&
+      hasDefaultFlugdaten({ context }) &&
+      hasAnnullierungOrNichtBefoerderung({ context }) &&
+      context.tatsaechlicherFlug === "no" &&
+      context.ersatzverbindungArt === "keineAnkunft" &&
+      hasOptionalString(context.zusaetzlicheAngaben)
+    );
+  };
+
+export const flugdatenDone: FluggastrechteFlugdatenGuard = ({ context }) => {
```
aaschlote, line 231: I'm bit lost on this huge `or` conditions here, can't we simply making small scopes for every case scenario?

For example, if the flight is Verspatung, make check the Verspatung, so it would be:

```
const bereich = context.bereich;

if (bereich = "Verspatung") {
 return hasVerspatungDone();
}

return hasAnnullierungOrNichtBefoerderungDone();
```

And inside the function `hasVerspatungDone` you can create small functions and reuse, like:

```
function hasVerspatungDone(): boolean {
const zwischenstoppAnzahl = context. zwischenstoppAnzahl;

if (zwischenstoppAnzahl === "no") {
 return hasNoZwischenstoppAnzahl();
}
}
```

In the function `hasNoZwischenstoppAnzahl` you can add all the conditions for no ZwischenStopp and reuse in another functions.

You also can have separate files for every function and have unit tests for everyone as well. We can chat more about it.

pgurusinga: I updated the done logic and I manage to cut down from 270 to 170ish line of code

aaschlote: Much better!

app/flows/fluggastrechteFormular/flugdaten/__test__/doneFunctions.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    direktFlugnummer: "AB6303",
+    direktAbflugsDatum: "10.03.2024",
+    direktAbflugsZeit: "09:09",
+    zwischenstoppAnzahl: "no",
+    direktAnkunftsDatum: "10.03.2024",
+    direktAnkunftsZeit: "09:09",
+    zusaetzlicheAngaben: "nice",
+  } as const;
+
+  test("returns true for all complete Flugdaten", () => {
```
aaschlote, line 23: can we have a test case description for every expect you are doing here?

aaschlote (approved):
Just check the tests. LGTM!

PR #1129, opened 2024-08-29
PKH Subflow - Einkuenfte

Spencer6497:
This PR adds the PKH Einkuenfte subflow to prozesskostenhilfe/financial-angaben.

- **move staatlicheLeistungen to shared context, begin einkuenfte flow**
- **add buergergeld and arbeitslosengeld screens, build out einkommen**
- **complete einkuenfte subflow up to abzuege**
- **rename selbststaendige fields for specificity, complete selbststaendige flow pages**
- **move adresseSchema to common context, begin building out abzuege flow**
- **add arbeitsausgaben array**
- **add optional arrayDataModifier function to show display-only values in the ArraySummary view**
- **build out einkuenfte subflow through Leistungen**
- **add leistungen pages**
- **add empty array guards and warning pages, add further income fields**
- **Move Einkuenfte to its own subfolder**
- **test doneFunctions, finish flow test happy path, fix andereArbeitsausgaben guard, finish doneFunctions**
- **add missing information string for einkuenfte, fix done function logic to skip all finanzielle angaben if user receives grundsicherung or asylbewerberleistungen**


app/components/arraySummary/ArraySummaryDataItems.tsx
```diff
@@ -40,6 +39,8 @@ const ArraySummaryDataItems = ({
     translations[`${category}.label.heading`] ?? "",
     {
       indexPerson: (itemIndex + WEITERE_PERSONEN_START_INDEX).toString(),
+      indexArbeitsausgabe: (itemIndex + 1).toString(),
```
aaschlote, line 42: as the `indexArbeitsausgabe` and `indexEinkunft` has the same value, maybe you just need one placeholder, right? Can you also add unit test for that?

app/services/array/index.ts
```diff
@@ -13,4 +14,12 @@ export type ArrayConfig = {
   statementKey: AllContextKeys | "showAlways";
   statementValue?: boolean;
   hiddenFields?: string[];
+  /**
+   * Optional mutation function to add display-only fields to the array
+   * @param items items in the ArraySummary
+   * @returns Modified ArraySummary items
+   */
+  arrayDataModifier?: (
```
aaschlote, line 22: what's the difference from this attribute and from the attribute ` hiddenFields`?

app/services/array/getSummaryData.ts
```diff
@@ -16,8 +16,13 @@ export function getSummaryData(
       .filter((category) => category in arrayConfigurations)
       .map((category) => {
         const arrayConfiguration = arrayConfigurations[category];
+        const { arrayDataModifier: dataModifier } = arrayConfiguration;
         const possibleArray = userData[category];
-        const data = Array.isArray(possibleArray) ? possibleArray : [];
+        const data = Array.isArray(possibleArray)
```
aaschlote, line 21: do we have an unit test for that?

aaschlote (commented):
Can you check why you don't have testscases for the prozesskostenhilfeFormular/finanzielleAngaben? Is it something expected?

Small tip, if you wanna keep the PR smaller, maybe you can create a small PR for moving the `...adresseSchema` to the `shared` folder and another PR for changes in the ArraySummary.

app/services/array/index.ts
Spencer6497: afaict, `hiddenFields` exist on the strapi object, but are hidden from the summary view -- this is the opposite case: this function _adds_ calculated, read-only fields, using the strapi data

Spencer6497:
> Can you check why you don't have testscases for the prozesskostenhilfeFormular/finanzielleAngaben? Is it something expected?

[here](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/987b4d37d3a7bc3e1e47cbfe7a71f22c3488143d/app/flows/prozesskostenhilfeFormular/__test__/testcases.ts#L15) are our prozesskostenhilfe/finanzielleAngaben test cases..... currently we only have the one "happy path", but we're sure to add more

> Small tip, if you wanna keep the PR smaller, maybe you can create a small PR for moving the `...adresseSchema` to the `shared` folder and another PR for changes in the ArraySummary.
good idea, would I just cherry-pick the changes from this branch into another branch, then selectively revert them here?

app/services/array/index.ts
aaschlote: Not sure if I understand, can you gimme an example?

Spencer6497: [here's](https://www.figma.com/design/7AspL0WtCnZqYbjuxsW7LM/Prozesskostenhilfe?node-id=270-163039&t=5TXuIqNohQvx8haH-4) an example: the field "Betrag umgerechnet pro Monat" is a calculated field, which takes the entered sum and divides it/displays it per-month. Afaik there's no way to do this in Strapi, and I couldn't figure out how to add it to the data model on form submission.

aaschlote: I got. Maybe can you think about a better name? Because it's called `arrayDataModifier`, but you are calling a callback function.

Two other things, I'm missing test for the function `addMonthlyAmounts` and it has a small sonar lint error here (image below)

image

tests/fixtures/prozesskostenhilfeFormularData.ts
```diff
@@ -18,17 +25,61 @@ export const happyPathData: ProzesskostenhilfeFormularContext = {
...
-  partnerEinkommenSumme: faker.finance.amount(),
+  partnerEinkommenSumme: faker.finance.amount({ min: 1 }),
   hasKinder: YesNoAnswer.Enum.yes,
   hasWeitereUnterhaltszahlungen: YesNoAnswer.Enum.yes,
   bankkonten: [
     {
       kontoEigentuemer: Eigentuemer.Enum.myself,
       bankName: faker.finance.accountName(),
-      kontostand: faker.finance.amount(),
+      kontostand: faker.finance.amount({ min: 1 }),
```
rbrtrfl, line 82: in fact, we want to allow for negative amounts as account balance (reality shows that's often the case)

rbrtrfl (commented):
nice one!
I am still checking out how you implemented that arrayDataModifier method and will think about whether this could be a solution for a missing feature in my PR.

Let me know when you have questions regarding my comments

app/services/array/index.ts
Spencer6497: [UPDATE]: after a discussion with design, we're nuking the `dataModifier` logic altogether

app/services/array/getSummaryData.ts
Spencer6497: [UPDATE]: after a discussion with design, we're nuking the `dataModifier` logic altogether

app/components/arraySummary/ArraySummaryDataItems.tsx
```diff
@@ -40,6 +39,8 @@ const ArraySummaryDataItems = ({
...
+      indexEinkunft: (itemIndex + 1).toString(),
```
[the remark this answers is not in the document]
Spencer6497: this prints out the array item number next to its' label, e.g. "Werbungskosten 1, 2...." etc

PR #1137, opened 2024-08-30
Sanitise Referer Header

pgurusinga:


app/services/security/sanitizeReferrer.ts
```diff
@@ -0,0 +1,9 @@
+export function sanitizeReferrer(referrer: string, origin: string) {
+  const rootPath = "/";
+
+  if (referrer === "") return rootPath;
```
chohner, line 4: is this early return necessary?

app/routes/persoenliche-daten-loeschen.tsx
```diff
@@ -19,10 +20,14 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
   const isSubmitting = useNavigation().state === "submitting";
-  const [previousPage, setPreviousPage] = useState("");
+  const [previousPage, setPreviousPage] = useState<string | undefined>(
+    undefined,
+  );
 
   useEffect(() => {
-    setPreviousPage(document.referrer);
+    setPreviousPage(
+      sanitizeReferrer(document.referrer, window.location.origin),
```
chohner, line 29: maybe make this a variable?

app/services/security/sanitizeReferrer.ts
pgurusinga: we need to handle:
1. user who visits the page directly
2. user who came from a page with no-referrer set (thus in our page `document.referrer` is an empty string)

https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

chohner: but I think this should work without the early return, right?
```
export const sanitizeReferrer = (referrer: string, origin: string) =>
  (new URL(referrer)).origin === origin ? referrer : "/"
```

chohner: also, I think object argument might be very helpful here (`sanitizeReferrer({referrer, origin})`)

pgurusinga: referrer must not be a string, otherwise, we will get a URL error
image

app/routes/persoenliche-daten-loeschen.tsx
```diff
@@ -37,7 +42,7 @@ export default function Index() {
           <ButtonContainer>
             <Button
               type="button"
-              href={previousPage.length > 0 ? previousPage : "/"}
+              href={previousPage}
```
aaschlote, line 45: the `previousPage.length > 0 ? previousPage : "/"` was because the non-JS users. What behaviour with this condition?

app/services/security/sanitizeReferrer.ts
```diff
@@ -0,0 +1,9 @@
+export function sanitizeReferrer(referrer: string, origin: string) {
+  const rootPath = "/";
```
aaschlote, line 2: you can declare this variable outside the function scope, like:

```
const ROOT_PATH = "/";
```

aaschlote (approved):
besides my comments, LGTM!

chohner (approved):

PR #1149, opened 2024-09-02
ADR for no-cache [skip-ci]

pgurusinga:


aaschlote (approved):
LGTM!

doc/adr/0010-cache-control-no-store.md
```diff
@@ -0,0 +1,22 @@
...
+After investigation, we discovered that the browser caches the previous page, leading to the display of outdated information when returning to a page after the banner has been accepted on a different page.
+
+## Decision
+
+To prevent the stale page issue and ensure that user interactions with the cookie banner are accurately reflected across pages, we decided to set the `Cache-Control: no-store` header in the client-side responses.
+
+## Consequences
+
+- The solution resolves the issue of stale pages by making sure the browser always retrieves the most up-to-date content.
+- As a result, when users click the back button, the page has to be fetched from the server each time.
```
chohner, line 22: imo this doesn't mention the downsides of disabling caching?

pgurusinga: is this enough or do you see another one? https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1149/files#diff-a839393f0e8cfc101396027c2ffc6dbb758ae0b6a0f59859b94363d6b1ed8538R23

doc/adr/0010-no-caching-while-cookie-banner-visible.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Normally, any interaction with the cookie banner (accepting or declining) should hide it until cookies expire or data is actively deleted. However, we encountered a problem where the cookie banner on would reappear under certain circumstances:
+
+When the user after interacting with the banner navigates to another page and then clicks the browser's back button, the browser cache will serve the previously cached page version with the banner still visible. Because the visibility state of the banner is checked on `GET` request, the cached page would still contained the visible cookie banner.
```
pgurusinga, line 13:
```suggestion
Normally, any interaction with the cookie banner (accepting or declining) should hide it until cookies expire or data is actively deleted. However, we encountered a problem where the cookie banner would reappear under certain circumstances:

When the user after interacting with the banner navigates to another page and then clicks the browser's back button, the browser cache will serve the previously cached page version with the banner still visible. Because the visibility state of the banner is checked on `GET` request, the cached page would still contain the visible cookie banner.
```

doc/adr/0010-no-caching-while-cookie-banner-visible.md
```diff
@@ -0,0 +1,25 @@
+# 10. Cache-Control: No Store
```
pgurusinga, line 1: if it is a rewrite, why the title still the old one?

doc/adr/0010-no-caching-while-cookie-banner-visible.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Context
+
+Normally, any interaction with the cookie banner (accepting or declining) should hide it until cookies expire or data is actively deleted. However, we encountered a problem where the cookie banner on would reappear under certain circumstances:
+
+When the user after interacting with the banner navigates to another page and then clicks the browser's back button, the browser cache will serve the previously cached page version with the banner still visible. Because the visibility state of the banner is checked on `GET` request, the cached page would still contained the visible cookie banner.
+
+## Decision
+
+We decided to disable caching while the cookie banner is visible by setting the `Cache-Control: no-store` header while the cookie state is `undefined`.
```
pgurusinga, line 17: as I understood and observed the browser, we don't disable the cache as the browser still handles and decides all of the caching (memory, file caching mechanism), and that's why I didn't use the word "disable the cache", but correct me if I am wrong

if we disable the cache, there's an option in the browser e.g. Chrome
image

chohner: good point, let me rephrase!

PR #1155, opened 2024-09-04
Check if useRef variable is not null

aaschlote:
Trying to fix the issue of the image below

image

### DIFF
--- a/app/components/cookieBanner/__test__/CookieBanner.test.tsx
+++ b/app/components/cookieBanner/__test__/CookieBanner.test.tsx
@@ -1,5 +1,5 @@
+import { createRemixStub } from "@remix-run/testing";
 import { render, screen } from "@testing-library/react";
-import { RouterProvider, createMemoryRouter } from "react-router-dom";
 import { CookieBanner } from "~/components/cookieBanner/CookieBanner";
 
 describe("CookieBanner", () => {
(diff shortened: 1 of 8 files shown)


app/components/cookieBanner/__test__/CookieBanner.test.tsx
```diff
@@ -17,13 +17,13 @@ describe("CookieBanner", () => {
   };
 
   it("renders content without link when no link props given", () => {
-    const router = createMemoryRouter([
+    const CookieBannerWithRemixStub = createRemixStub([
```
pgurusinga, line 20: We can use the `render()` function, similar to the last test I mentioned.

package.json
```diff
@@ -75,6 +75,7 @@
     "react-dom": "^18.3.1",
     "react-dropzone-esm": "^15.0.1",
     "react-imask": "^7.6.1",
+    "react-router-dom": "^6.26.1",
```
pgurusinga, line 78: can you elaborate on why we added `react-router-dom` and the only place we use directly `react-router-dom` you removed it
image

aaschlote: I was checking some threads in the stackoverflow and I read that the problem might be the usage of the `react-dom` library, so I removed it from the tests. Unfortunately it didn't work, so I decided to install it (someone mentioned, to usage the latest version), and again it didn't work.

aaschlote:
Closing this PR for now, I'll check in another time

PR #1164, opened 2024-09-05
Improve cache response header

aaschlote:
## Problem

We want to avoid implementing a global no-caching mechanism.

### 1. Initial Approach

We initially decided to set the `Cache-Control: no-store` header in the `root.tsx` file. However, we faced challenges because the `HeadersFunction` overrides any headers set in the JSON response.

For reference, this is a common issue discussed here: [Remix Discussion on HeadersFunction](https://github.com/remix-run/remix/discussions/6529).

#### Example of the Issue

```js
// Headers function
export const headers: HeadersFunction = () => ({
  "X-Frame-Options": "SAMEORIGIN",
  // other headers
});
```

```js
// Client won't receive this because of the headers function
return json(rootLoaderResponse, {
  headers: {
    "Cache-Control": "no-store",
  },
});
```

Since `HeadersFunction` does not receive the `request` parameter, we cannot conditionally set response headers based on the `Cookie` value.

### 2. Revised Approach (Working Solution)

Instead, we can pass the `Cache-Control` header through the `json()` function. Then, in the `headers` function, we parse and conditionally merge the `Cache-Control` header.

#### Implementation

```js
// Set the Cache-Control header conditionally in the json function
const cacheControlHeader =
  hasTrackingCookie === undefined
    ? { "Cache-Control": "no-store" }
    : undefined;

return json(rootLoaderResponse, {
  headers: cacheControlHeader,
});
```

```js
// Headers function to merge the Cache-Control header conditionally
export const headers: HeadersFunction = ({ loaderHeaders }) => {
  const responseHeaders: Record<string, string> = {
    "X-Frame-Options": "SAMEORIGIN",
    // other headers
  };

  // Conditionally add Cache-Control header if present
  if (loaderHeaders.get("Cache-Control") === "no-store") {
    responseHeaders["Cache-Control"] = "no-store";
  }

  return responseHeaders;
};
```


app/root.tsx
```diff
@@ -95,6 +106,7 @@ export type RootLoader = typeof loader;
 export const loader = async ({ request, context }: LoaderFunctionArgs) => {
   const { pathname } = new URL(request.url);
   const cookieHeader = request.headers.get("Cookie");
+  const hasTrackingCookie = await trackingCookieValue({ request });
```
aaschlote, line 109: you can add it inside the `Promise.all` scope :)

app/root.tsx
```diff
@@ -124,27 +136,35 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
...
-  return json({
-    header: {
-      ...getPageHeaderProps(strapiHeader),
-      /**
-       * Only hide the header links if we're viewing a flow page
-       */
-      hideLinks: !headerLinksEnabled || !!flowIdFromPathname(pathname),
+  const cacheControlHeader =
+    hasTrackingCookie === undefined
+      ? { "Cache-Control": "no-store" }
```
aaschlote, line 141: what about to have our own custom header value here? It could be:

```
{ "Should-Disable-Cache-Control": "true" }
```

and the line above you can ado:

```
if (loaderHeaders.get("Should-Disable-Cache-Control") === "true") {
    responseHeaders["Cache-Control"] = "no-store";
  }
```

This makes easier to understand when we should add the cache-control

aaschlote (commented):
LGTM, but we can think to add an small E2E Test to see if the behaviour works (only to check if the response header has `Cache-Control=no-store` during the action of the click on the banner)

app/root.tsx
chohner: or even `acceptedThirdPartyCookies: "true"`? The responsibility of what to do with this information lies solely in the `header` function

chohner (approved):
feels weird to approve this myself after making a commit but I really like this approach 🎉

PR #1166, opened 2024-09-06
BerH: Bremen oder Hamburg Frage

Spencer6497:
This PR adds a guard against users from Bremen and Hamburg from applying for Beratungshilfe, as they have other resources available to them.

Copy and flow co-opted from the pre-existing Vorabcheck page.


app/flows/beratungshilfeFormular/grundvoraussetzung/context.ts
```diff
@@ -1,11 +1,12 @@
 import { z } from "zod";
 import { YesNoAnswer } from "~/services/validation/YesNoAnswer";
-import type { GenericGuard, Guards } from "../../guards.server";
+import { type GenericGuard, type Guards } from "../../guards.server";
```
chohner, line 3: was this an auto-edit or do you prefer this syntax?

chohner (approved):

app/flows/beratungshilfeFormular/grundvoraussetzung/context.ts
Spencer6497: auto-edit!

PR #1170, opened 2024-09-06
pdf pruner

rbrtrfl:
- add the pruner to the pdf service to prune the userData before it's fed into the filling mechanism
- remove duplicate logic in the pdf service

https://app.asana.com/0/1203259475859667/1208083697682762


app/routes/shared/formular.server.ts
```diff
@@ -99,9 +98,7 @@ export const loader = async ({
   context.debugId = debugId; // For showing in errors
 
   const currentFlow = flows[flowId];
-  const prunedUserData = (await isFeatureFlagEnabled("pruneUserData"))
-    ? await pruneIrrelevantData(userData, flowId)
-    : userData;
+  const prunedUserData = await pruneIrrelevantData(userData, flowId);
```
chohner, line 101: are we sure we want to do this? 🤔

rbrtrfl: I was under the assumption, that the feature flag is simply switched on, but I see its actually off 😬

Anyway: what speaks againts it? How could users be affected?

chohner: At the very least I would not hide it in this PR 😇

aaschlote: this feature toggle was already enable in production?

rbrtrfl: I thought about it once again: Maybe this _is_ a suitable and PR to remove the FF? I mean the scope of the ticket is about actual user-facing issues. So at least the FF should be switched on, but we could as well remove it. What do you think?

app/services/pdf/beratungshilfe/sections/F_besitz/util.ts
```diff
@@ -0,0 +1,6 @@
+export const eigentuemerMapping = {
+  myself: "Ich alleine",
+  partner: "Ehe-Partner:in",
+  myselfAndPartner: "Mein:e Ehe-Partner:in und ich gemeinsam",
+  myselfAndSomeoneElse: "Ich gemeinsam mit jemand anderem",
+} as const;
```
chohner, line 6: where does this come from?

rbrtrfl: I see that you did exactly the same thing on Friday and moved this to a separate file `eigentuemerMapping.ts`. I moved this in order to remove a circular dependency.

PR #1175, opened 2024-09-09
use pdfFillReducer to fill BerH pdf

rbrtrfl:
- move fillOutFunction one level up
- use pdfFillReducer to fill BerH pdf


app/services/pdf/beratungshilfe/sections/F_besitz/F_besitz.ts
```diff
@@ -1,107 +1,116 @@
...
-  attachment: AttachmentEntries,
-  pdfFields: BeratungshilfePDF,
-  context: BeratungshilfeFormularContext,
-) {
-  const { hasBankkonto, bankkonten, eigentumTotalWorth } = context;
+  return {
+    pdfValues,
+    attachment,
+  };
+};
```
Spencer6497, line 38: we should be destructuring and returning the result of `pdfFillReducer`, like so:
```
const { pdfValues, attachment } = pdfFillReducer(...)
return {
    pdfValues,
    attachment
}
```

does the way it currently is work?

app/services/pdf/beratungshilfe/sections/G_ausgaben.ts
```diff
@@ -25,53 +26,52 @@ export const ausgabenSituationMapping = {
...
-) {
-  const ausgaben = context.ausgaben ?? [];
+export const fillAusgaben: BerHPdfFillFunction = ({ userData, pdfValues }) => {
+  const attachment: AttachmentEntries = [];
+
+  const ausgaben = userData.ausgaben ?? [];
   const hasOverflowAusgaben = ausgaben.length > AUSGABEN_MAX_COUNT_FIELDS;
 
-  const isPdfFieldExceedsMaxChars = context.ausgaben?.some(
+  const isPdfFieldExceedsMaxChars = userData.ausgaben?.some(
```
Spencer6497, line 35:
```suggestion
  const isPdfFieldExceedsMaxChars = ausgaben.some(
```

Spencer6497 (changes requested):
Great work on this 🙌 though I'm fairly certain there's a logical error where noted ⬇️

app/services/pdf/beratungshilfe/sections/F_besitz/F_besitz.ts
rbrtrfl: you are absolutely right!

app/services/pdf/beratungshilfe/sections/A_angelegenheit.ts
```diff
@@ -20,20 +20,20 @@ const bereichMapping = {
...
 
-export function fillAngelegenheit(
-  attachment: AttachmentEntries,
-  pdfFields: BeratungshilfePDF,
-  context: BeratungshilfeFormularContext,
-) {
+export const fillAngelegenheit: BerHPdfFillFunction = ({
+  userData,
+  pdfValues,
+}) => {
```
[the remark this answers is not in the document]
rbrtrfl: As far as I understand, a type alias like `BerHPdfFillFunction` can only be used like this with anonymous function definitions. If we would go for a named function, explicit typing for parameters and return value would be necessary.

Spencer6497 (approved):

chohner (approved):
lgtm

PR #1176, opened 2024-09-09
Beratungshilfe Vorabcheck: Refactor flow.json to ts object

rbrtrfl:


app/flows/beratungshilfeVorabcheck/guards.ts
```diff
@@ -83,39 +38,17 @@ const anyKinderAnzahlFilled: Guard = ({ context }) =>
...
-  ...yesNoGuards("kinderKurz"),
-  ...filledGuard("einkommenKinder"),
-  ...yesNoGuards("unterhalt"),
-  ...filledGuard("unterhaltSumme"),
-  ...filledGuard("miete"),
-  weitereZahlungenSummeIncomeTooHigh,
-  weitereZahlungenSummeWithWarnings,
-  ...filledGuard("weitereZahlungenSumme"),
+  staatlicheLeistungenNo,
+  staatlicheLeistungenYes,
```
chohner, line 53: any reasons these should still be defined here, rather than inline? I think the only one thats complex enough that it would warrant tests is `isIncomeTooHigh`.

We could then even get rid of the whole `guards` object..

ekl176: The staatlicheLeistungen and anyNonCriticalWarnings seemed like deliberate abstractions of business logic. Generally I don't see an issue in moving those inline.

PR #1183, opened 2024-09-10
interpolateDeep Function Guard

Spencer6497:
This PR adds a second helper function, interpolateFlowDeep, to interpolate strings in a flow and recursively guard against passed-in objects. It also adds units tests.


app/util/__test__/fillTemplate.test.ts
```diff
@@ -24,3 +29,80 @@ describe("fillTemplate", () => {
     });
   });
 });
+
+describe("interpolateDeep", () => {
```
pgurusinga, line 33: can we add a negative spec for this e.g. when three curly brackets, then we expect an error? or `{{ whatever }}` with empty space?

app/util/__test__/fillTemplate.test.ts
```diff
@@ -24,3 +29,80 @@ describe("fillTemplate", () => {
...
+      replaceMe: "New String!",
+    });
+    expect(interpolatedFlow.prop2.prop2Child).toBe("New String!");
+  });
+
+  it("should throw an error if the flow contains a nested function", () => {
+    const flow = {
+      prop1: "bar",
+      prop2: {
+        prop2Child: "{{replaceMe}}",
```
pgurusinga, line 71: Does interpolateDeepFlow only work on 2 levels, or can it be infinitely nested? if the latter then could you add spec for the deeper one? otherwise, ignore this message

pgurusinga (approved):

PR #1184, opened 2024-09-10
Doc: add context ADR [skip ci]

chohner:
Adds and overview of data structures called `context` and suggests new naming


doc/adr/0011-userdate-context.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Context
+
+XState, our dependency that we use for (flow) state machines, calls the data object that it uses for decision inside guards [context](https://stately.ai/docs/context). This lead us to call many different things "context", such as instances of userdata and even the schema of userdata.
+
+## Decision
+
+We want to use distinct names for the following three things:
+
+### 1. Userdata Schema
```
pgurusinga, line 17: I’m thinking we could use `userInputSchema`. The word 'input' clearly shows that the data comes from the user’s actions. While things like cookie consent, lastStep and UUID are user data, they don’t belong in this schema.

doc/adr/0011-userdate-context.md
```diff
@@ -0,0 +1,64 @@
...
+
+const flowSchema = {
+  firstname: z.string(),
+  lastname: z.string(),
+};
+```
+
+Note: Because we validate on both front and back-end, all schemas must be isomorphic, ie run in both environments! Server-only validation can be achieved using the `serverOnly$` function.
+
+### 2. Userdata
```
pgurusinga, line 32: also maybe we could change this to `userInput`?

doc/adr/0011-userdate-context.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Note: We do not use xstate's [assign mechanism](https://stately.ai/docs/context#updating-context-with-assign) to update context.
+
+## Consequences
+
+Refactor:
+
+- Rename of all types, variable names, function arguments and names for 1. and 2.
+- Update of documentation
```
pgurusinga, line 64: these, to me, feel like TODOs derived from the decision, consequences could be:
- This change will enhance code readability
- Developers will more easily differentiate between the user input, user input's schema, and XState's Context

pgurusinga (commented):
nice! finally the time comes

doc/adr/0011-userdate-context.md
```diff
@@ -0,0 +1,64 @@
+# 11. Userdata and context
+
+Date: 2024-09-10
+
+## Status
+
+Accepted?
```
frederike-ramin, line 7: Why the question mark? ;)

doc/adr/0011-userdate-context.md
```diff
@@ -0,0 +1,64 @@
...
+```typescript
+import { flowSchema } from "./flowSchema"; // See above
+
+const _partialSchema = z.object(flowSchema).partial();
+type FlowUserData = z.infer<typeof _partialSchema>;
+
+const userdata = {
+  firstname: "Jane",
+} satisfies FlowUserData;
+```
```
frederike-ramin, line 45: Why is the word `flow` in everything now? O.o Can we potentially call it `NameSchema` or something like it?

doc/adr/0011-userdate-context.md
frederike-ramin: "conversations will get easier" :D

frederike-ramin (approved):
That's a very nice ADR, nicely written with all required details 👌

doc/adr/0011-userdate-context.md
chohner: the flow was standing in for `beratungshilfeVorabcheck`... do you think that would make it more clear?

doc/adr/0011-userdate-context.md
chohner: I wonder if this leads to confusing with a single input / form submission

doc/adr/0011-userdate-context.md
frederike-ramin: Yes, maybe `BeratungshilfeVorabcheckUserData` then?

doc/adr/0011-userdate-context.md
frederike-ramin: Yes, I also prefer userData and userDataSchema, I have to say.

PR #1189, opened 2024-09-12
Fix writing special chars to pdf

rbrtrfl:
- use BundesSans font for filling the PDF form fields
- safely encode filename of the exported pdf


app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -32,10 +32,13 @@ export async function pdfDownloadLoader({ request }: LoaderFunctionArgs) {
...
     pdfConfigs[flowId as keyof typeof pdfConfigs];
 
+  const filename = filenameFunction(userData);
+  const encodedFilename = encodeURIComponent(filename);
+
   return new Response(await (await pdfFunction(userData)).save(), {
     headers: {
       "Content-Type": "application/pdf",
-      "Content-Disposition": `inline; filename=${filenameFunction(userData)}`,
+      "Content-Disposition": `inline; filename*=UTF-8''${encodedFilename}`,
```
pgurusinga, line 41: thanks for taking care of it, could we add a test? if there's a change from PDF lib (for any reason), we will not catch it again. Do you think that test is unnecessary and we are confident enough?

frederike-ramin (commented):
I have to admit, I don't really get how this avoids the character encoding problem within the PDF? Or was the error in the filename all along?
But to make it more apparent - and also just because it's better and makes sure it doesn't happen again - could you add a unit test testing the encoding of special characters? 😇

rbrtrfl:
@frederike-ramin @pgurusinga this PR contains two changes. Using the BundesSans font to fill out the form fields in the pdf is what actually fixes the bug.

While fixing, I stumbled across that another bug that crashed the app when the last name contains special characters.

Sure, I can add a unit test for the filename encoding!

app/services/pdf/fillPdf.server.ts
```diff
@@ -61,6 +63,17 @@ export async function fillPdf({
 
   const form = pdfDoc.getForm();
 
+  pdfDoc.registerFontkit(fontkit);
+  const fontFilePath = path.join(
+    process.cwd(),
+    "/public/fonts/",
+    "BundesSansWeb-Regular.woff",
+  );
+  const customFontBytes = fs.readFileSync(fontFilePath);
```
chohner, line 72: I think this could be done once, rather than on every function call?

app/routes/shared/pdfDownloadLoader.ts
chohner: I think this isn't touching the pdf interface but just enabling UTF8 in the filename for download? I didn't even consider this being an issue...

app/services/pdf/fillPdf.server.ts
```diff
@@ -61,6 +63,17 @@ export async function fillPdf({
...
+  const customFont = await pdfDoc.embedFont(customFontBytes);
+  const rawUpdateFieldAppearances = form.updateFieldAppearances.bind(form);
+  form.updateFieldAppearances = () => rawUpdateFieldAppearances(customFont);
```
chohner, line 75: A regression test for this would be super

app/services/pdf/fillPdf.server.ts
```diff
@@ -61,6 +63,17 @@ export async function fillPdf({
 
   const form = pdfDoc.getForm();
 
+  pdfDoc.registerFontkit(fontkit);
+  const fontFilePath = path.join(
+    process.cwd(),
+    "/public/fonts/",
+    "BundesSansWeb-Regular.woff",
```
chohner, line 70: I wonder if there are still chars that would cause breaking the pdf. I think we might still should strip all non-supported chars from the user data

chohner (approved):
nice. left some comments ✅

app/services/pdf/fillPdf.server.ts
rbrtrfl: I agree and I've also left a comment in the ticket. I would suggest to add both FE and BE/zod validation to disallow all non latin characters (including special chars, numbers ofc). But I would suggest this is for another ticket?

PR #1190, opened 2024-09-12
fix: skip partnerinformation on separated partner in finanzielleAngaben

ekl176:
---
  - https://app.asana.com/0/0/1208269580933971


app/flows/beratungshilfeFormular/finanzielleAngaben/flow.json
```diff
@@ -116,7 +116,7 @@
                 "target": "#partner.unterhalt"
               },
               {
-                "guard": "hasPartnerschaftOrSeparated",
```
chohner, line 119: looks like we could remove (or at least remove the export) of `hasPartnerschaftOrSeparated` :)

chohner (approved):
nice, tried it ✅ we could now even do some additional cleanup I think

app/flows/beratungshilfeFormular/finanzielleAngaben/flow.json
ekl176: good idea. I removed the guard completely and changed subsequent guards that relied on the partnerschaftOrSeperated to use hasPartnerschaftYes instead.

chohner (approved):
lgtm

PR #1198, opened 2024-09-13
 refactor(xstateConfig): remove BH vorabcheck guards

chohner:
This is an extension to https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1176

Moving all trivial guards into the config leaves a single function (`isIncomeTooHigh`) which I think is a) semantically nice and allows instantiation of the state machine without carrying around a separate `guards` object


app/flows/beratungshilfeVorabcheck/xstateConfig.ts
```diff
@@ -1,7 +1,14 @@
...
 import { type BeratungshilfeVorabcheckContext } from "./context";
-import { guards } from "./guards";
+import { isIncomeTooHigh } from "./isIncomeTooHigh";
+import type { GenericGuard } from "../guards.server";
+
+const staatlicheLeistungenYes: GenericGuard<
+  BeratungshilfeVorabcheckContext
+> = ({ context }) =>
+  context.staatlicheLeistungen === "grundsicherung" ||
+  context.staatlicheLeistungen === "asylbewerberleistungen";
```
Spencer6497, line 11: iirc we recently made a shared implementation of this guard somewhere?

Spencer6497 (approved):
🚀

app/flows/beratungshilfeVorabcheck/xstateConfig.ts
rbrtrfl: Kudos für refactoring this guard. I really don't like the `.includes` syntax. 🫰

PR #1202, opened 2024-09-16
Refactor Daten-Übernehme page

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1208156212423004


app/routes/fluggastrechte.formular.grundvorraussetzungen.redirect-vorabcheck.tsx
```diff
@@ -0,0 +1,3 @@
+import { type LoaderFunction, redirect } from "@remix-run/node";
+//workaround to redirect to the vorabcheck
+export const loader: LoaderFunction = async () => redirect("/fluggastrechte");
```
pgurusinga, line 3: can you ensure that PostHog will not track this page? or PostHog only tracks client side?

tests/e2e/flowPages/fluggastrechte/formular/fluggastrechte.formular.spec.ts
```diff
@@ -12,9 +12,22 @@ test.beforeEach(async ({ page }) => {
...
+    page,
+  }) => {
+    const redirectPromise = page.waitForResponse(
+      (resp) => resp.url().includes("/fluggastrechte") && resp.status() === 200,
+    );
+
+    await page.goto(
+      `${formular.url}/grundvorraussetzungen/redirect-vorabcheck`,
+    );
+    await redirectPromise;
```
pgurusinga, line 31: We need to test if clicking the button redirects the user to the Vorab-Check page, right?

I would expect something like this:
```
WHEN I am on the daten übernahme page
THEN I click on the Vorab-Check Wiederholen button
THEN I should be on the Vorab-Check start page
```

```js
// Pseudo code
page.clickBack()
await expect(page).toHaveURL(
  new RegExp(`.+${formular.url}/fluggastrechte/vorabcheck/$`),
);
```

So, we don't care what the redirection is doing; we test based on the user's perspective.

aaschlote: The idea wasn't to test the user interaction, but the redirection, because of two things:

- We need a test to redirection in the route
- To reach this page, we need to do the whole Vorabcheck flow and this would have a long e2e test and have one doing this already (test for the klage), so I would say we don't need another one

Ideally this supposed to be an integration test, but as we don't have define how the integration test exist, I would like to add this and iterate later once we decide how we write integration test

app/routes/fluggastrechte.formular.grundvorraussetzungen.redirect-vorabcheck.tsx
pgurusinga: I had a look with @judithmh earlier and we use `posthog-node` for feature flag not server-side tracking

tests/e2e/flowPages/fluggastrechte/formular/fluggastrechte.formular.spec.ts
aaschlote: nice idea, me and @judithmh implemented your suggestion :)

PR #1204, opened 2024-09-17
Fix writing special chars to pdf

rbrtrfl:
- use BundesSans ttf to fill form fields in pdf
- embed font in pdf
- add regression test to verify the app does not crash on pdf generation when userData contains special characters


package.json
```diff
@@ -43,6 +43,7 @@
     "@digitalservice4germany/angie": "^1.1.2",
     "@digitalservice4germany/style-dictionary": "^2.0.0",
     "@digitalservicebund/icons": "^2.0.0",
+    "@pdf-lib/fontkit": "^1.1.1",
```
aaschlote, line 46: can you check if this dependency has been maintainable? Last published was 4 years ago and I don't see so much changes in their github repo

app/services/pdf/beratungshilfe/sections/A_angelegenheit.ts
```diff
@@ -32,7 +32,10 @@ export const fillAngelegenheit: BerHPdfFillFunction = ({
     .map((x) => `${x.title} ${x.text}`)
     .join("\n");
 
-  if (sachverhaltString.length > 255) {
+  if (
+    sachverhaltString.length > 255 ||
```
aaschlote, line 36: What's 255? Maybe you shouldn't define a magic number but a constant variable to say what this value means. Article to [read](https://refactoring.guru/replace-magic-number-with-symbolic-constant) to know more about it :)

app/services/pdf/beratungshilfe/__test__/beratungshilfePdfFromUserdata.test.ts
```diff
@@ -33,4 +33,13 @@ describe("beratungshilfePdfFromUserdata", () => {
 
     expect(pdfField.getText()).toEqual("nachname, vorname");
   });
+
+  it("regression: special characters without throwing", async () => {
```
aaschlote, line 37: Why do you call this a regression test? is it not a simple test? Otherwise all my tests I would need to call also regression test, right? Or am I missing something?

app/services/pdf/beratungshilfe/sections/F_besitz/fillKraftfahrzeug.ts
```diff
@@ -32,18 +32,23 @@ export const fillKraftfahrzeug: BerHPdfFillFunction = ({
...
 
-  if (kraftfahrzeuge.length == 1) {
-    const kraftfahrzeug = kraftfahrzeuge[0];
-    pdfValues.f10KraftfahrzeugeA.value = kraftfahrzeug.eigentuemer == "myself";
-    pdfValues.f10KraftfahrzeugB.value = kraftfahrzeug.eigentuemer == "partner";
+  const singleKraftfahrzeug = kraftfahrzeuge[0];
+  const singleKraftfahrzeugString =
+    fillSingleKraftfahrzeug(singleKraftfahrzeug);
+
+  if (kraftfahrzeuge.length == 1 && singleKraftfahrzeugString.length < 96) {
```
aaschlote, line 39: What's 96? Maybe you shouldn't define a magic number but a constant variable to say what this value means. Article to [read](https://refactoring.guru/replace-magic-number-with-symbolic-constant) to know more about it :)

app/services/pdf/beratungshilfe/sections/F_besitz/fillKraftfahrzeug.ts
```diff
@@ -3,7 +3,7 @@ import type { BerHPdfFillFunction } from "../..";
 import { newPageHint, type AttachmentEntries } from "../../../attachment";
 import { eigentuemerMapping } from "../../eigentuemerMapping";
 
-function singleKraftfahrzeug(
+function fillSingleKraftfahrzeug(
```
aaschlote, line 6: Maybe you can think to do later, but this function has 100 lines, normally if my function more than 30 lines I try to think to break in small pieces, to be easy to understand it.

One [article](https://martinfowler.com/bliki/FunctionLength.html?ref=morsewall.com) you can read about it

aaschlote (commented):
We really need to check the dependency, it doesn't look people are taking care well of it

app/services/pdf/beratungshilfe/__test__/beratungshilfePdfFromUserdata.test.ts
chohner: Thats probably on me: imo unit tests that validate certain bugs don't re-appear are regression tests. Not sure whether we need to specifically write it in the test description though

app/services/pdf/beratungshilfe/sections/A_angelegenheit.ts
```diff
@@ -32,7 +32,10 @@ export const fillAngelegenheit: BerHPdfFillFunction = ({
...
+    sachverhaltString.split("\n").length > 4
```
chohner, line 37: Maybe move this into its own commit (and merge to main?)

For clarity, maybe this would be enough to express intent:

```typescript
const doesOverflow = sachverhaltString.length > 255;
const newlineCount = sachverhaltString.split("\n").length > 4;

if (doesOverflow || newlineCount > 4)
```

app/services/pdf/beratungshilfe/sections/F_besitz/fillVermoegenswerte.ts
```diff
@@ -60,17 +60,23 @@ export const fillVermoegenswerte: BerHPdfFillFunction = ({
...
+  const singleVermoegenswert = geldanlagen[0] ?? wertsachen[0];
+  const singleVermoegenswertString =
+    fillSingleVermoegenswert(singleVermoegenswert);
 
-    pdfValues.f14InhaberA.value = vermoegenswert.eigentuemer == "myself";
-    pdfValues.f14InhaberB.value = vermoegenswert.eigentuemer == "partner";
+  if (
+    totalVermoegenswerteCount == 1 &&
+    singleVermoegenswertString.length < 148 &&
+    singleVermoegenswertString.split("\n").length < 3
```
chohner, line 70: see above

app/services/pdf/pdf.server.ts
```diff
@@ -1,4 +1,4 @@
-import type { PDFForm } from "pdf-lib";
+import { type PDFForm } from "pdf-lib";
```
chohner, line 1: ?

chohner (commented):
thanks for trying this! I'd love to keep this PR only about the font change - all the additional overflow logic could imo move into main already

app/services/pdf/pdf.server.ts
rbrtrfl: prettier did that, not me 😬

package.json
rbrtrfl: The main package `pdf-lib`, which we are using to fill out the forms also had its last release and commit 3 years ago. We have not found a good alternative yet, but surely might have to look into it again, when we aim to make our pdf fully accessible.

As for now, I think we can live with the fact that new package has the same poor maintenance, since it does not add any further dependencies other than the existing ones in `pdf-lib`.

app/services/pdf/beratungshilfe/__test__/beratungshilfePdfFromUserdata.test.ts
```diff
@@ -33,4 +33,13 @@ describe("beratungshilfePdfFromUserdata", () => {
...
+    const pdfDoc = await beratungshilfePdfFromUserdata({
+      vorname: "Włodzimierz",
+      nachname: "Ćwikła",
```
pgurusinga, line 40: can we add another test with emoji? will it break? 💰💰💰💰

app/services/pdf/beratungshilfe/sections/F_besitz/F_besitz.ts
```diff
@@ -52,7 +52,7 @@ export const fillFinancialBankkonto: BerHPdfFillFunction = ({
       ? "Übergreifender Hinweis zu allen Vermögenswerten:\nMein gesamtes Vermögen ist insgesamt weniger als 10.000€ wert.\n\n"
       : "";
 
-  if (bankkonten.length == 1) {
+  if (bankkonten.length == 1 && eigentumTotalWorth !== "less10000") {
```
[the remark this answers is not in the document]
rbrtrfl: If `eigentumTotalWorth === "less10000"`, we print a long note `Übergreifender Hinweis [...]` into the Bankbezeichnung form field which requires to print the bank details into the attachment. How could this be handled by the pruner?

app/services/pdf/beratungshilfe/__test__/beratungshilfePdfFromUserdata.test.ts
rbrtrfl: yes, we can 🤝

package.json
aaschlote: The only problem when this `pdf-lib` gets a vulnerability and then it'll become a problem 😞

rbrtrfl: I agree. We have to keep an eye on this.

app/services/pdf/beratungshilfe/sections/A_angelegenheit.ts
rbrtrfl: Since too many new lines would also cause the text to overflow, I would find this wording confusing. I would rather go for constant variables for the magic numbers helping understand what the condition does.

app/services/pdf/fillPdf.server.ts
```diff
@@ -39,6 +41,10 @@ async function readRelativeFileToBuffer(relativeFilepath: string) {
   }
 }
 
+const customFontBytes = fs.readFileSync(
+  path.join(process.cwd(), "/public/fonts/", "BundesSansOffice-Regular.ttf"),
+);
+
```
pgurusinga, line 47: is it possible to cache the read fonts in memory? or do you think it's already been optimised? I was thinking like below:

```ts
let cachedFont: Uint8Array | null = null;

const loadCustomFont = () => {
  if (!cachedFont) {
    cachedFont = fs.readFileSync(
  	path.join(process.cwd(), "/public/fonts/", "BundesSansOffice-Regular.ttf"),
    );
  }
  return cachedFont;
};
```

PR #1205, opened 2024-09-17
refactor: move BerH finanzelle angaben flow.json to xstateConfig.ts

ekl176:
---
  - https://app.asana.com/0/0/1208317196204181


app/services/flow/server/buildFlowController.ts
```diff
@@ -17,7 +17,10 @@ import {
 import { progressLookupForMachine, vorabcheckProgresses } from "./progress";
 
 type Event = "SUBMIT" | "BACK";
-type FlowStateMachineEvents = { type: "SUBMIT" } | { type: "BACK" };
+type FlowStateMachineEvents =
+  | { type: "SUBMIT" }
+  | { type: "BACK" }
+  | { type: ArrayConfig["event"] };
```
rbrtrfl, line 23: Could we use also use `ArrayConfig["event"]` instead of `Event` in line 64 for `const arrayEvents`?

chohner: We could potentially remove that cast - but imo we should have another look at this together outside of this PR

rbrtrfl (approved):

PR #1213, opened 2024-09-19
chore: move from sanitize-html to xss

chohner:
I realized that our dependency to sanitize rendered markdown causes a huge chunk of runtime warnings:

image

This is due to a transitive `postcss` dependency, see [this issue](https://github.com/apostrophecms/sanitize-html/issues/639).

There are two options:
- use an isomorphic library (this PR)
- render & sanatize markdown on server (preferred but larger architectural change)


aaschlote:
Can you please add tests for the `h` elements? It was added to be allowed recently but we don't have tests. And also a specific tests for removed some html tag element, like `code`? We don't have tests for that.

Besides that, the changes looks fine!

chohner:
> Can you please add tests for the `h` elements? It was added to be allowed recently but we don't have tests. And also a specific tests for removed some html tag element, like `code`? We don't have tests for that.
> 
> Besides that, the changes looks fine!

Good idea - I might do this before & outside of this PR

pgurusinga:
nice one, it seems the new package is trust worthy https://socket.dev/npm/package/xss, can we add ADR for this? 🎉

aaschlote (approved):
LGTM

PR #1214, opened 2024-09-19
ZOV: add abgabe page

pgurusinga:


app/flows/fluggastrechteFormular/index.ts
```diff
@@ -3,6 +3,7 @@ import type { AllContextKeys } from "~/flows/common";
 import type { Flow } from "~/flows/flows.server";
 import type { ArrayConfig } from "~/services/array";
 import type { FlowTransitionConfig } from "~/services/session.server/flowTransitionValidation.server";
+import abgabeFlow from "./abgabe/flow.json";
```
pgurusinga, line 6: vscode suggested this lol

app/flows/fluggastrechteFormular/abgabe/flow.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "abgabe",
+  "initial": "start",
+  "states": {
+    "start": {
```
aaschlote, line 5: can we make the url without the `start`? It's defined in the [ticket](https://app.asana.com/0/1207022303033556/1207867714962573) the url should be `/fluggastrechte/formular/abgabe`

app/flows/fluggastrechteFormular/abgabe/flow.json
```diff
@@ -0,0 +1,11 @@
+{
```
aaschlote, line 1: if you want, you can move this flow.json to the `app/flows/fluggastrechteFormular/index.ts`, so it would be something:

```
abgabe: {
        id: "abgabe",
        meta: { done: () => false },
        "on": {
         "BACK": "#zusammenfassung.start"
         }
        },
```

app/flows/fluggastrechteFormular/abgabe/flow.json
pgurusinga: That's a good question. We also have `zusammenfassung/start`, which I guess doesn't make sense since it doesn't start a process in the subflow.

pgurusinga: let's ask Kirstin about this.

aaschlote: true, in the [ticket](https://app.asana.com/0/1207022303033556/1208210353539057) for the `Zusammenfassung` wasn't defined the URL, so basically I created ending with `/start`, but it would be nice if the both page has the same structure. Let me know if you need some help to talk with Kirstin about it

pgurusinga: Kirstin said OK for adding `/zusammenfassung` and `/abgabe`

PR #1220, opened 2024-09-20
Fix input to large devices

aaschlote:
### Problem
On the Formular pages for mobile and tables, when an input is selected 54 characters, it jumps to the bottom of the page.

image

More info in the following slack [link](https://digitalservicebund.slack.com/archives/C044NUG2468/p1726160178710569)

### Solution
Change width to max-width

### DIFF
--- a/app/components/inputs/__test__/Input.test.tsx
+++ b/app/components/inputs/__test__/Input.test.tsx
@@ -27,19 +27,19 @@ afterEach(() => {
 
 describe("Input", () => {
   const cases = [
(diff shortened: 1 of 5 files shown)


chohner:
can't we change `w-[x]` to `max-w`?

chohner (approved):
lgtm, with a little comment

aaschlote:
> can't we change `w-[x]` to `max-w`?

It doesn't work so well, it breaks for small responsive sizes

image

PR #1223, opened 2024-09-24
Feat/update turkish airline and airport data

judithmh:
Added Turkish Airlines and corrected the spelling of AFlughafen to Flughafen for Istanbul:

I had to change this line from lower to upper case `const airportIndex = airportName.toUpperCase().indexOf("AIRPORT");` due to the Turkish İ. With lower case it was converted to an i and therefore a wrong index was returned


aaschlote (approved):
LGTM

PR #1231, opened 2024-09-26
feat/add-input-buchungsnummer: add passengerNameRecord (PNR) as field

judithmh:


app/services/validation/passengerNameRecord.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "zod";
+
+export const passengerNameRecordSchema = z
```
aaschlote, line 3: can we add test for this function?

aaschlote, line 3: I also got curious about the name `passengerNameRecord` because I never have heard about it and I did a small research and I found it:

```
From a technical point of view, there are five parts of a PNR required before the booking can be completed. They are:

The name of the passenger
Contact details for the travel agent or airline office.
Ticketing details, either a ticket number or a ticketing time limit.
Itinerary of at least one segment, which must be the same for all passengers listed.
Name of the person providing the information or making the booking.
```

It seems the booking number is a part of the passengerNameRecord (PNR), but in this validation we are doing the validation for the booking number and nothing else. Could we rename to bookingNumberFlight or something similar?

app/services/validation/passengerNameRecord.ts
```diff
@@ -0,0 +1,18 @@
...
+    // .trim() won't remove the inbetween space
+    // e.g. " LH 234 " becomes "LH 234"
+    return passengerNameRecord.split(" ").join("");
+  })
+  .refine(
+    (passengerNameRecord) => {
+      return /^[A-Z0-9]{6}$/.test(passengerNameRecord);
+    },
+    {
+      message: "invalid_passenger_name_record_format",
```
aaschlote, line 16: Did you create this in Strapi staging? I haven't found it

aaschlote (commented):
Your E2E test is failing due the missing input in content data file. We need to do a content release with this input. Ping me if you need some help to do it!

app/services/validation/__test__/bookingNumberFlight.test.ts
```diff
@@ -0,0 +1,54 @@
+// @vitest-environment node
```
aaschlote, line 1: I guess you don't need this line here

app/services/validation/passengerNameRecord.ts
judithmh: You are right - it would have been the PNR code :) I changed it to bookingNumberFlight as suggested

judithmh: yes you are right - added tests

aaschlote (approved):
LGTM, just need to fix the E2E test 😃 

Nice tests btw!

PR #1235, opened 2024-09-26
feat: pkh persoenlicheDaten subflow

ekl176:
---
  - https://app.asana.com/0/0/1208259193432695


app/flows/prozesskostenhilfeFormular/persoenlicheDaten/xstateConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+}): Config<ProzesskostenhilfePersoenlicheDaten> {
+  return {
+    id: "persoenliche-daten",
+    initial: "start",
+    meta: { done: beratungshilfePersoenlicheDatenDone },
+    states: {
+      start: {
+        on: {
+          SUBMIT: "name",
+          BACK: transitions?.backToCallingFlow,
```
Spencer6497, line 21: what happens if `transitions` is undefined? will it ever be undefined?

Spencer6497 (approved):
LGTM! 🚢 it 🫡

app/flows/prozesskostenhilfeFormular/persoenlicheDaten/xstateConfig.ts
ekl176: At first I made the transitions mandatory.
I think the idea was that by making the transitions optional, we make the subflow truly independant from others. 
A single transition to step into or out of the flow might happen when it moves to the first or last step in a flow.

PR #1241, opened 2024-09-30
refactor(Video Component sonarlint complaints)

Spencer6497:
This PR fixes the sonarlint complaints regarding Video.tsx, namely in moving inline-components to discrete children with prop passing.

### DIFF
--- a/app/components/video/__test__/Video.test.tsx
+++ b/app/components/video/__test__/Video.test.tsx
@@ -35,7 +35,7 @@ describe("Video Component", () => {
     expect(queryByRole("img")).not.toBeInTheDocument();
   });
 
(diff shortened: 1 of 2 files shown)


app/components/video/__test__/Video.test.tsx
```diff
@@ -34,4 +34,15 @@ describe("Video Component", () => {
     expect(queryByTitle("Test Video")).toBeInTheDocument();
     expect(queryByRole("img")).not.toBeInTheDocument();
   });
+
+  it("should still render the thumbnail, given accepted cookies and a broken link", () => {
```
pgurusinga, line 38: I am wondering why do you write `given accepted cookies`? in the test, we don't fire event to accept it 🤔

PR #1244, opened 2024-09-30
refactor: unify translation contexts

chohner:
- new /translation folder


app/services/translations/translationsContext.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext, useContext } from "react";
```
aaschlote, line 1: Thinking if it's the correct path (`service`) for this hook to be, because it's a react component, I would expect to be in the `components` folder and not in the `service`. 

And this file should be `.tsx` and not `.ts`?

app/components/userFeedback/feedbackTranslations.ts
```diff
@@ -0,0 +1,38 @@
...
+  "text-post-submission",
+] as const;
+export type PostSubmissionTranslationKeys =
+  (typeof postSubmissionTranslationKeys)[number];
+
+const ratingBoxTranslationsKeys = ["yes-rating", "no-rating"] as const;
+export type RatingBoxTranslationKeys =
+  (typeof ratingBoxTranslationsKeys)[number];
+
+export function useFeedbackTranslations() {
```
aaschlote, line 23: The file is called `feedbackTranslations` but inside export a hook with a name `useFeedbackTranslations`, can we make this file a `.tsx` with the name `useFeedbackTranslations`?

app/components/arraySummary/ArraySummaryDataItems.tsx
```diff
@@ -4,7 +4,7 @@ import { WEITERE_PERSONEN_START_INDEX } from "~/flows/fluggastrechteFormular/str
 import type { ArrayConfig } from "~/services/array";
 import type { Translations } from "~/services/cms/index.server";
 import { interpolateSerializableObject } from "~/util/fillTemplate";
-import { getTranslationByKey } from "~/util/getTranslationByKey";
+import { getTranslationByKey } from "~/services/translations/getTranslationByKey";
```
aaschlote, line 7: can we fix this `warning` lint here?

aaschlote (commented):
Can we think also to move the `fetchTranslations("amtsgericht")` to the root file? It's used in 5 different places, maybe it worth to move there as well.

PR #1245, opened 2024-09-30
feat/add-translations-for-header: add translations

judithmh:
added translations for leichte Sprache and Gebärdensprache


app/components/PageHeader.tsx
```diff
@@ -31,14 +34,14 @@ export default function Header({
             {/* TODO: Add strapi translations */}
```
aaschlote, line 34: can we remove this TODO?

app/components/__test__/PageHeader.test.tsx
```diff
@@ -8,6 +8,10 @@ describe("PageHeader", () => {
         title={"Justiz-Services"}
         linkLabel={"Justiz-Services"}
         hideLinks={false}
+        translations={{
```
chohner, line 11: looking at the many repeated definitions of the same object - would it make sense to add something like `defaultTranslations` here? To be used as a fallback and across component tests and storybook? 🤔

pgurusinga: It seems like this approach makes sense for tests and stories, but I'm not sure if it's suitable for fallback. How can we sync the string between Strapi and the fallback? Let's consider this scenario:
1. The Content Manager changes the title or button label.
2. The content is deployed.
3. Developers are unaware of when and where the changes were made.
4. In the app:
- When data is successfully fetched, the button displays the text "test" as a "test" button.
- When there's a failure in fetching the data, the button displays "trial" as the button text.

chohner: Imo we shouldn't try keep it in sync, that would defeat the point of the CMS.

The trade-off is what you would rather have shown in case of a missed lookup (which already would be a failure):
 - the outdated label (for example `Leichte Sprache` instead of `Leichte Sprache für alle`)
- the raw text ID (`leichtesprache`)
 
I would strongly vote for the first one...

Either way - not blocking at all

aaschlote: I would recommend to use the function `getTranslationByKey`, so in case the translation it doesn't exist:

- It sends a warning message to sentry about the missing of the key
- It fallbacks the key as translation

The problem of the fallback, we will never know if the key is missing in the CMS and this information must come from the CMS.

But I'm fine with the current implementation, we can merge it!

PR #1246, opened 2024-09-30
Fixathon improve feedback block api

aaschlote:


aaschlote:
@frederike-ramin @ekl176 

I added a few unit tests and some small improvements, can you please take a look when. you have time?

frederike-ramin (approved):
Nice! I also really love the additional tests. I think it's ready to merge 👍 Thanks for taking care of this even after the Fixathon!

app/services/feedback/getRedirectForNonRelativeUrl.ts
```diff
@@ -0,0 +1,7 @@
+import { json } from "@remix-run/node";
+
+export function getRedirectForNonRelativeUrl(searchParameterUrl: string) {
```
chohner, line 3: I think there could be some improvements here:
- the naming is off, the function doesn't really `get` anything and when it returns something, its not a redirect
- imo the route handling should be in the decision domain of the loader, rather than this function. Maybe instead we could have a more general `isRelativeUrl()`?

aaschlote: - The idea of the team group was to have a function that it returns an `redirect` object in case the URL is not relative, for this reason `getRedirectForNonRelativeUrl`
- We thought about it, but we decided to return a `redirect` object
- `Thrown new Response` or return `json` basically does the same thing, so I keep the change as it's

PR #1261, opened 2024-10-04
fix(List links bullet point styling)

Spencer6497:
This PR fixes a styling bug where link list items with a newline would display the bullet point at the bottom, instead of at the top, of the list item.

---
  - https://app.asana.com/0/0/1208311653825044


Spencer6497:
@aaschlote afaict removing this doesn't affect other StandaloneLinks in the app, but lmk if you see a problem with it

aaschlote:
> @aaschlote afaict removing this doesn't affect other StandaloneLinks in the app, but lmk if you see a problem with it

I added this `inline-block` because of this [ticket](https://app.asana.com/0/1207022303033556/1207820849591133). Can you check if it doesn't break anything mentioned in the ticket?

PR #1263, opened 2024-10-04
feat(PKH Abgabe)

Spencer6497:
This PR updates the PKH Abgabe flow to have the correct steps listed, depending on user-entered data.

- **replace json with ts object xStateConfig, define guards inline**
- **remove now-unused abgabeArt/abgabeContext, extract verfahren type to discrete guards, simplify abgabe flow, add abgabe string replacements**
- **rename ueberpruefung guard, add erstantrag guard, filter out empty list items**
- **update e2e test**

---
  - https://app.asana.com/0/0/1208350416730394


app/flows/prozesskostenhilfeFormular/abgabe/xStateConfig.ts
```diff
@@ -0,0 +1,26 @@
+import type { Context } from "~/flows/contexts";
+import type { GenericGuard } from "~/flows/guards.server";
+import type { Config } from "~/services/flow/server/buildFlowController";
+
+export function getAbgabeXstateConfig(transitions?: {
+  readyForAbgabe?: GenericGuard<Context>;
```
chohner, line 6: why is this a function and not simply defined in line? we aren't using `abgabeXstateConfig` anywhere else, right?

Spencer6497: ah you're totally right, I guess I was blindly following the pattern we'd set for reusable/shared configs 😅

PR #1264, opened 2024-10-04
Fit connect

zechmeister:


app/services/flow/submission.server.ts
```diff
@@ -0,0 +1,24 @@
+import axios from "axios";
+import { configDotenv } from "dotenv";
+import { config } from "../env/env.server";
+
+type SubmitRequest = {
+  message: string;
+};
+
+export async function submit() {
+  configDotenv();
```
chohner, line 10: I'm surprised this is necessary here?

PR #1267, opened 2024-10-07
Feat/add payment details plaintiff

judithmh:
Add input and validation for iban and kontoinhaber 
ticket: https://app.asana.com/0/1207022303033556/1208071078791946/f


app/services/validation/__test__/iban.test.ts
```diff
@@ -0,0 +1,55 @@
+// @vitest-environment node
```
aaschlote, line 1: I think we don't need this line, right?

app/services/validation/iban.ts
```diff
@@ -0,0 +1,19 @@
+import { isValidIBAN } from "ibantools";
+import { z } from "zod";
+
+export const ibanSchema = z
+  .string()
+  .toUpperCase()
+  .transform((ibanInput) => {
+    // .trim() won't remove the inbetween space
```
aaschlote, line 8: copy&pasta? 😅

app/services/validation/iban.ts
```diff
@@ -0,0 +1,19 @@
+import { isValidIBAN } from "ibantools";
+import { z } from "zod";
+
+export const ibanSchema = z
+  .string()
+  .toUpperCase()
```
aaschlote, line 6: do we need to do`upperCase`?

package.json
aaschlote, line 1: did you change the lines from the file or was the prettier? Maybe you can revert it, just to don't have unneeded changes with your name here

aaschlote (approved):
LGTM, nice tests 👌

app/services/validation/iban.ts
judithmh: you are right - I removed it

app/services/validation/__test__/iban.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  describe("failing cases", () => {
+    const cases = [
+      { input: "", errorMessage: "invalid_iban_format" },
+      { input: "DE", errorMessage: "invalid_iban_format" },
+      { input: "123DE", errorMessage: "invalid_iban_format" },
+      { input: " :) ", errorMessage: "invalid_iban_format" },
+      { input: "foobar ", errorMessage: "invalid_iban_format" },
+      { input: "DE9110000000012345678", errorMessage: "invalid_iban_format" },
+      { input: "DE911000000001234567891", errorMessage: "invalid_iban_format" },
```
pgurusinga, line 36: why these IBANs are invalid? other countries have different lengths as well.

aaschlote: From this website: https://www.ibancalculator.com/iban_validieren.html

```

This IBAN cannot be correct because of its length. A German IBAN always contains exactly 22 digits and letters ("DE", a 2-digit checksum, the 8-digit bank code and 10 digits for the account number, possibly padded with zeros). The IBAN you entered is 23 characters long.
--
  | Account number : Account checksum not checked.

```

app/services/validation/iban.ts
```diff
@@ -0,0 +1,16 @@
+import { isValidIBAN } from "ibantools";
```
pgurusinga, line 1: is this package trustworthy? do we have a security risk here?

app/services/validation/__test__/iban.test.ts
pgurusinga: yeah, that's for Germany. but are you sure we only have people come from Germany? please read here https://www.iban.com/structure

aaschlote: the library validate for every country, as the initial is "DE", it validates as german IBAN.

app/services/validation/iban.ts
judithmh: It does depend on other packages, but its size is relatively small, it has over 200 000 weekly downloads and is frequently updated, repo has depandabot enabled and it has 200+ stars on github and `npm audit` does not find any warnings/vulnerabilites. Thus, I would say that the security risk is rather low.

app/services/validation/iban.ts
```diff
@@ -0,0 +1,16 @@
...
+import { z } from "zod";
+
+export const ibanSchema = z
+  .string()
+  .transform((ibanInput) => {
+    return ibanInput.split(" ").join("");
```
chohner, line 7: I think `ibanInput.replaceAll(" ", "")` is more readable here

app/services/validation/iban.ts
```diff
@@ -0,0 +1,16 @@
...
+  })
+  .refine(
+    (ibanInput) => {
+      return isValidIBAN(ibanInput);
+    },
+    {
+      message: "invalid_iban_format",
+    },
+  );
```
chohner, line 16: I think this can just be:
```
.refine(isValidIBAN, { message: "invalid_iban_format" }
```

app/services/validation/iban.ts
chohner: I support using a package here - I once looked into validating IBANs and its a mess 😅

app/services/validation/__test__/iban.test.ts
judithmh: I added some more test with extra long IBAN/ extra short IBAN and mixed letter/number IBAN from other countries :)

PR #1269, opened 2024-10-07
feat(Textarea textHint Accordion)

Spencer6497:
This PR edits the Textarea component to display an optional textHint accordion above the input.

Bildschirmfoto 2024-10-07 um 11 58 00

---
  - https://app.asana.com/0/0/1208269012872152


app/components/inputs/Textarea.tsx
```diff
@@ -1,16 +1,19 @@
...
 import { type ErrorMessageProps } from ".";
 import InputError from "./InputError";
-import InputLabel from "./InputLabel";
 import RichText from "../RichText";
 
 type TextareaProps = Readonly<{
   name: string;
   description?: string;
   label?: ReactNode;
+  textHint?: DetailsSummaryProps;
```
chohner, line 16: Two thoughts:
- `textHint` doesn't communicate that a details component will appear. Maybe something more explicit like `detailsTop`? Or just `details`? 
- can we make the type a bit more strict? I think if the prop is there, it should have content and an optional headline (and no identifier)

Note: maybe the last point should be fixed right inside `DetailsSummaryProps` but I don't know what else depends on those types

Spencer6497: * yeah I was sort of unsure about the name; in Strapi there's a `description` field that serves more or less the same purpose (without collapsing), so I figured a name like `details` might be confusing. for the sake of convenience, could we alias it in the code and leave the strapi name, or should we change it across the board?
* since this is a sub-component, wouldn't it be more descriptive/accurate to include _all_ of the component's props?

app/flows/beratungshilfeFormular/rechtsproblem/stringReplacements.ts
```diff
@@ -0,0 +1,17 @@
+import type { BeratungshilfeFormularContext } from "~/flows/beratungshilfeFormular";
+
+export const getRechtsproblemStrings = (
+  context: BeratungshilfeFormularContext,
+) => {
```
chohner, line 5: this probably belongs in another branch? 😅

app/components/inputs/Textarea.tsx
chohner: 1. The label in strapi is easily changed, so I would prefer to have good naming in the app. Having the same naming across the stack is helpul of course. But there might be differences between internal (app) and external (strapi label) names
2. Imo restricting to only valid combination is usually helpful. Combining multiple flexible components might not result in all permutations being valid (ie a details sub-component without content)

app/components/DetailsSummary.tsx
```diff
@@ -7,7 +7,7 @@ export const DetailsSummarySchema = z.object({
   content: z.string().optional(),
 });
 
-type DetailsSummaryProps = z.infer<typeof DetailsSummarySchema>;
+export type DetailsSummaryProps = z.infer<typeof DetailsSummarySchema>;
```
aaschlote, line 10: Do u mind to transform this from a zod to a normal typescript type? I believe we don't need to parse it in the source code

app/components/inputs/__test__/Textarea.test.tsx
```diff
@@ -79,16 +71,35 @@ describe("Textarea component", () => {
...
+        ),
+      },
+    ]);
 
+    render(<RemixStub />);
+    const accordion = screen.getByText("Text-Beispiel");
+    expect(accordion).toBeInTheDocument();
+    expect(accordion).toHaveClass("summary-content");
+    expect(screen.getByText("Lorem ipsum")).not.toBeVisible();
+    accordion.click();
```
aaschlote, line 97: I believe this test is not necessary here, because you should test only if the component `DetailsSummary` render on the page. To see clickable works, the test should be done in `DetailsSummary` component, this gives the single responsibility for the component. What I would do here, mock the `DetailsSummary` and expect if the component is rendering from your mock.

aaschlote (approved):
LGTM, please check how it renders on mobile devices!

app/components/DetailsSummary.tsx
chohner: I was about to write something similar but it requires a big change due to the schema being nested inside multiple other schema. I think we should rather just create one PR to move away from app-side zod schemas

PR #1270, opened 2024-10-07
doc: Add ADR for domain-oriented architecture

frederike-ramin:
As discussed in the Dev Exchange.


doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
...
+We want to make an effort to separate domain logic (i.e. configurations) even more from the application logic (e.g. how to interpret flow configurations).
+The goal is: If you want to add a new domain/form to the platform, you only have to touch _one_ place (maybe two, including `routes`), and it is very clear which one that is and what you have to/can configure there.
+
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe`and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
+- Things that are now specific to one service should reside in the `domains` folder. That includes:
```
aaschlote, line 30: A few things mentioned here are related react components, for those things I would like to keep maybe in the `components` folder, the others things for non-react components I feel it's ok to move to the domain folder.

frederike-ramin: You mean these two?

>   - `WEITERE_PERSONEN_START_INDEX` in the ArraySummaryDataItems
>   - `FluggastrechteResultPage` in components

The first one, I do think we need a domain-agnostic way of handling. We might not only want a different indexing for the one specific "personen" use case but for others as well.

The second one is debatable, yes. However, the easiest solution would just be to move the component into the according `route` and we'd be done. :)

aaschlote: I was thinking also to maybe inside the `domain` a subfolder called `components`, only contain the react components for the domain.

For the `WEITERE_PERSONEN_START_INDEX` it seems a new version of the `stringReplacements` for `ArraySummary`, it could be new function inside the `arrayConfig`. But we can discuss it also.

My strong feeling is having the react components inside a specific folder and the most common (and I believe everybody also does the same) is the `components` one.

frederike-ramin: Yes. I left the "todos" without specific solutions yet so that we had exactly those conversations afterwards. I'm also not 100% clear on how to solve the `WEITERE_PERSONEN_START_INDEX` issue. Your suggestion sounds interesting, let's explore that. 

> My strong feeling is having the react components inside a specific folder

Not arguing with you there at all. I think so too.

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
+# 11. Splitting domain configurations and application logic
```
aaschlote, line 1: should be 12, not?

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
...
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe`and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
+- Things that are now specific to one service should reside in the `domains` folder. That includes:
+  - airport data + logic
+  - pdf filling logic + handout generation
+  - Strapi "extra" flowIDs
+  - Vorabcheck progress configuration
```
chohner, line 34: what is progress configuration? I don't think that exists anymore 🤔

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
...
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe`and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
+- Things that are now specific to one service should reside in the `domains` folder. That includes:
+  - airport data + logic
+  - pdf filling logic + handout generation
+  - Strapi "extra" flowIDs
```
chohner, line 33: What this? 🤔

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
...
+The goal is: If you want to add a new domain/form to the platform, you only have to touch _one_ place (maybe two, including `routes`), and it is very clear which one that is and what you have to/can configure there.
+
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe`and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
+- Things that are now specific to one service should reside in the `domains` folder. That includes:
+  - airport data + logic
```
chohner, line 31: Maybe rather `domain specific data, for example airports` ?

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
...
+
+We want to make an effort to separate domain logic (i.e. configurations) even more from the application logic (e.g. how to interpret flow configurations).
+The goal is: If you want to add a new domain/form to the platform, you only have to touch _one_ place (maybe two, including `routes`), and it is very clear which one that is and what you have to/can configure there.
+
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe`and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
```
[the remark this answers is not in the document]
frederike-ramin: Yes, probably. I'd say that's up for the "domain owners" then though?

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: I knew the naming was confusing. Do you have a better name for this?
https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/c3a7230f73a6c85d96ff12f1275e40ec7e8d61d1/app/services/flow/server/progress.ts#L56

(Only found this because I searched for "beratungshilfe" ;)

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: I was also surprised to find this^^: https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/app/services/cms/models/StrapiFlowId.ts

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Context
+
+We develop an application that is thought to host a lot of different forms.
+They all are in the justice space, but still vary in the specific domain within justice.
+For example, the domain knowledge you need is very different from a Beratungshilfe form to a Fluggastrechte form.
+They all have in common that they need forms, content, data validation etc. (from now on called "application logic").
+The only thing that is really different is the configuration of those.
+Currently, we are mostly separating the domains' configurations from the application logic withing the folder "flows".
+However, some parts of the domain configuration still reside in other modules, e.g. the PDF filling logic.
```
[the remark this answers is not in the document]
frederike-ramin: I added a sentence but omitted the "uncertainty" part, sounded a little too self-rightous to me^^

doc/adr/0012-domain-oriented-architecture.md
chohner: aha! But this isn't configuration and shouldn't be defined inside a flow imo

But interesting point, we don't really have a mechanism to do something for a subset of flows, ie all vorabchecks.

However, this might be addressed by having a root configuration:

```typescript
export const vorabcheckIds = ["a", "b"] as const; // Used inside the progress service
export const formFlowIds = ["c", "d"] as const;
export const flowIds = [...vorabcheckIds, ...formFlowIds] as const;
export type FlowIds = typeof flowIds;
```

(I don't think this needs to be part of this ADR)

doc/adr/0012-domain-oriented-architecture.md
chohner: WHICH IDI... oh. 😇

This is a weird one but I would rather say this is not a "normal" configuration property.

Similar to above, this may also disappear by having some root configuration

doc/adr/0012-domain-oriented-architecture.md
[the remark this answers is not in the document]
chohner: Absolutely. But I think describing this pattern here explicitly might be helpful

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -20,6 +21,11 @@ However, some parts of the domain configuration still reside in other modules, e
 
 We want to make an effort to separate domain logic (i.e. configurations) even more from the application logic (e.g. how to interpret flow configurations).
 The goal is: If you want to add a new domain/form to the platform, you only have to touch _one_ place (maybe two, including `routes`), and it is very clear which one that is and what you have to/can configure there.
+We follow the following naming:
+
+- domain: An area in the justice space that needs specific knowledge, e.g. `beratungshilfe`
+- flow: A sequence of form steps to be completed within a domain, e.g. `beratungshilfe vorabcheck`
+- flowId: The identifier to differentiate between different flows, e.g. `/beratungshilfe/vorabcheck`
```
chohner, line 28: I think actually there is a different concept that we never spoke about: Flow type (?), ie `vorabcheck` and `flow` (differ on progress bar, side navigation, funnel logic) 🤔

doc/adr/0012-domain-oriented-architecture.md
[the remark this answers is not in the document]
frederike-ramin: Okay, could you prepare a sentence? Not sure how I would phrase that.

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: I still kind of like the idea to handle "/beratungshilfe/zustaendiges-gericht" as another flow you mentioned in a call.

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: You're right. Can we make this another ADR though?

doc/adr/0012-domain-oriented-architecture.md
chohner: I think my only challenge is that there is nothing to configure, because its fundamentally based on the "type" of the flow (vorabcheck vs flow). So for now I wouldn't include it in the canonical list of things that should be configurable

doc/adr/0012-domain-oriented-architecture.md
chohner: see above - then lets remove it from the list of configurable things and rather add it as a tech debt ticket to refactor?

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: Mh..how do you define then which flow gets a progress and which doesn't? just because of the name "vorabcheck"? I'd think that's quite dangerous?

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: Aahhh..now I see the point of your comment(s)! To you, it's about sorting out only the configuration. So "What isn't 'configuration' is 'core'. To me, I'm always thinking the other way around: "What isn't 'core', has to be configuration. But I see your point and we can just have it as a refactoring ticket and I'll remove it here.
It also seems to align neatly with  [this ask](https://digitalservicebund.slack.com/archives/C031NV6FMQA/p1728464861234759) by Marion, I'd say.

ekl176 (approved):

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Context
+
+We started this project developing only one flow for one domain (Vorabcheck for Beratungshilfe).
+Since then, we shifted towards developing an application that is thought to host a lot of different forms.
+They all are in the justice space, but still vary in the specific domain within justice.
+For example, the domain knowledge you need is very different from a Beratungshilfe form to a Fluggastrechte form.
+They all have in common that they need forms, content, data validation etc. (from now on called "application logic").
+The only thing that is really different is the configuration of those.
+Currently, we are mostly separating the domains' configurations from the application logic withing the folder "flows".
```
manuelpuchta, line 17: Nit, little typo:
```suggestion
Currently, we are mostly separating the domains' configurations from the application logic within the folder "flows".
```

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+However, some parts of the domain configuration still reside in other modules, e.g. the PDF filling logic.
+
+## Decision
+
+We want to make an effort to separate domain logic (i.e. configurations) even more from the application logic (e.g. how to interpret flow configurations).
+The goal is: If you want to add a new domain/form to the platform, you only have to touch _one_ place (maybe two, including `routes`), and it is very clear which one that is and what you have to/can configure there.
+We follow the following naming:
```
manuelpuchta, line 24: Nit:
```suggestion
We strive for the following naming:
```

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+- domain: An area in the justice space that needs specific knowledge, e.g. `beratungshilfe`
+- flow: A sequence of form steps to be completed within a domain, e.g. `beratungshilfe vorabcheck`
+- flowId: The identifier to differentiate between different flows, e.g. `/beratungshilfe/vorabcheck`
+
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe`and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
```
manuelpuchta, line 35: Nit:
```suggestion
- There might be a domain-overarching folder called `shared` for domains very similar to each other, like `beratungshilfe` and `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`.
```

And in regards to your discussion, without detailed knowledge about that part, maybe something like that is already understandable enough:
> Logic that is useful for multiple domains or flows, like `beratungshilfe`, `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`, shall be moved into a dedicated `shared` folder

manuelpuchta (approved):

zechmeister (approved):

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+  - `WEITERE_PERSONEN_START_INDEX` in the ArraySummaryDataItems
+  - `FluggastrechteResultPage` in components
+  - probably some more
+- A Readme on "how to add a new domain/form" would be beneficial.
+
+Expected Benefits:
+
+- It is clear where to add a new domain/form to the application.
+- It is very clear when I as a developer edit domain-specific logic and when I add things to the overarching application logic.
+- Possibly: Domains do not "interfere" with each other quite that much as they are kept separate.
```
frederike-ramin, line 50: @aaschlote you asked for more specific advantages on the day-to-day basis. I'm struggeling a bit with this, as these pretty much already sum it up for me. Do you have more thoughts on this where I could improve the benefits?

doc/adr/0012-domain-oriented-architecture.md
chohner: Currently the only thing that is specifically different between vorabcheck and flow pages is `cmsSlug` (I think). The progress calculation depending on "type" is mostly a performance optimization..

How about each configuration exposing a `type FlowType = "vorabcheck" | "full-flow"`, which services can consume?
- the content lookup service can configure the fitting api endpoint
- the progress can filter by `type === "vorabcheck"`
- ....

frederike-ramin: Yes, love it. Great solution. Just to be clear: I'll leave it as a Todo here then. The concrete solution comes when solving it. 👍

chohner: I still believe that it does not belong in this list of explicit things that are in the configuration domain of a flow and would rather remove it here

frederike-ramin: Misunderstanding resolved in: https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1270/commits/70c776707aa5435520c8f76938b6135918a72fea

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: Added in: https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1270/commits/70c776707aa5435520c8f76938b6135918a72fea

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,55 @@
...
+- flow: A sequence of form steps to be completed within a domain, e.g. `beratungshilfe vorabcheck`
+- flowId: The identifier to differentiate between different flows, e.g. `/beratungshilfe/vorabcheck`
+- flowType: Deciding factor what purpose the flow serves, currently `vorabcheck` or `formular`
+
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- Configurations that are useful for multiple domains or flows, like `beratungshilfe`, `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`, shall be moved into a dedicated `shared` folder.
```
chohner, line 36: Could we replace these two hard-to-read bullet points with a diagram?


- In terms of folder structure, flow configurations are scoped by their domain. Shared code should be placed in a `shared` folder that may exist on any level:
	```
	.
	└── domains
	    ├── fluggastrecht
	    │    ├── vorabcheck
	    │    │    ├── index.ts      # Exports full flow configuration
	    │    │    ├── schemas
	    │    │    ├── stringReplacements
	    │    │    └── guards
	    │    ├── formular
	    │    │    ├── index.ts
	    │    │    └── pdf           # flow specific pdf logic
	    │    └── shared             # Shared code inside domain
	    ├── beratungshilfe
	    │    ├── vorabcheck
	    │    └── antrag
	    └── shared                  # Shared code across cross-domain, ie:
	         ├── persoenlicheDaten	# Common sub-flow configuration
	         └── schemas            # Useful schemas
	```

rbrtrfl (commented):
nice initiative. I've left some minor comments/questions.

doc/adr/0012-domain-oriented-architecture.md
chohner: Not all but only domain/flow specific ones (which we probably will only have very very few)

doc/adr/0012-domain-oriented-architecture.md
pgurusinga: Yes, I would love to see the difference; the one before the change and the one after it.

doc/adr/0012-domain-oriented-architecture.md
```diff
@@ -0,0 +1,54 @@
...
+- domain: An area in the justice space that needs specific knowledge, e.g. `beratungshilfe`
+- flow: A sequence of form steps to be completed within a domain, e.g. `beratungshilfe vorabcheck`
+- flowId: The identifier to differentiate between different flows, e.g. `/beratungshilfe/vorabcheck`
+
+## Consequences
+
+"To Dos:"
+
+- A new folder is created called `domains`. In that we have a folder for each domain, e.g. `beratungshilfe`, `fluggastrechte`. In those we have more folders depicting the specific flows of the domain, e.g. `vorabcheck` and `formular`.
+- Configurations that are useful for multiple domains or flows, like `beratungshilfe`, `prozesskostenhilfe` or domain-irrelevant things like `persoenlicheDaten`, shall be moved into a dedicated `shared` folder.
```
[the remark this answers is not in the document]
aaschlote: it should be inside, check this [link](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1270#discussion_r1795332044)

doc/adr/0012-domain-oriented-architecture.md
aaschlote: I'm having a bit of trouble fully grasping the value of the proposal compared to our current implementation. It might be because I'm so used to the existing setup, or perhaps I'm not seeing the complete picture of the new approach.

That's why I suggested including some references or examples from other sources, to help me view the proposal from a different perspective 😃.

My opinion could definitely change once we have everything in place and I see it in action.

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: Wait, I understood Alex' suggestion was to *not* have components outside the `/app/components` folder? Because atm the only one we have, is the FluggastrechteResultPage and that can just reside in the routes ✅ ?

frederike-ramin: If that is the case (please confirm, @aaschlote) then we can for the sake of this PR remove the mention of `FluggastrechteResultPage` here and instead make it a refactoring task. :)

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: Committed it as you suggested. Thank you, that makes it very clear. 👍

frederike-ramin: @pgurusinga if you want the current state you can probably check the repo, right? No need to specify it here or would you disagree?

doc/adr/0012-domain-oriented-architecture.md
aaschlote: I meant what @rbrtrfl and @chohner said, we can have react components inside of the domain with a subfolder `components`, for specific react components of the domain or flow, like the `FluggastrechteResultPage`.

I won't like to see the react components inside of non `components` folder.

doc/adr/0012-domain-oriented-architecture.md
frederike-ramin: Okay, I see. In my head it goes into the way of Configuration-drive Architecture: https://www.bro-code.in/blog/configuration-driven-architecture
Only that the configuration is done by code and not JSON/YAML (therefor needing a deployment for changes).
Does that link help you? Then I can include it in the ADR.

aaschlote: yeap, this helps me :)

rbrtrfl (approved):

pgurusinga (approved):

m0dh4x (changes requested):
i think this applies to both worlds, so no drawback here in particular. I even think it's a bit clearer when you're working on something that affects both of the domains or all upcoming ones

"Expected drawbacks:

There might be more communication efforts necessary when changing application logic as it affects other/all domains."

frederike-ramin:
@m0dh4x yes, you're right. Are you suggesting to remove the drawbacks section in total?

aaschlote (approved):

m0dh4x (approved):

PR #1271, opened 2024-10-07
Fluggastrechte: First draft PDF version hardcode

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1208435805621982

- Use [PDF-Kit](https://github.com/foliojs/pdfkit) dependency to generate PDF
- Use hardcode data to generate PDF
- Tagged the text with the correct structure a11y
- Modify `pdfDownloadLoader` to call `pdfFunction` to return a Buffer


package.json
```diff
@@ -132,9 +134,9 @@
     "tsx": "~4.19.1",
     "typescript-eslint": "^8.8.1",
     "typescript": "^5.6.2",
+    "vite": "^5.4.8",
```
chohner, line 137: i always wonder about this resorting... what plugin is doing it? 🤔

aaschlote: prettier I believe it.

app/services/pdf/fluggastrechte/__test__/setPdfMetadata.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    const document = new PDFDocument();
+
+    setPdfMetadata(document);
+
+    expect(document.version).toBe(VERSION);
+    expect(document.info.Title).toBe(TITLE);
+    expect(document.info.Author).toBe(AUTHOR);
+    expect(document.info.Subject).toBe(SUBJECT);
+    expect(document.info.Keywords).toBe(KEYWORDS);
+    expect(document.info.CreationDate).not.toBeNull();
```
pgurusinga, line 23: could we expect this to be an instance of Date instead?

app/services/pdf/fluggastrechte/sections/createPageNumber.ts
```diff
@@ -0,0 +1,18 @@
...
+} from "../createPdfKitDocument";
+
+export const createPageNumber = (
+  doc: typeof PDFDocument,
+  { pageNumber }: { pageNumber: number },
+) => {
+  doc
+    .fontSize(7)
+    .font(FONTS_BUNDESSANS_REGULAR)
+    .text(`${pageNumber}/3`, 585 - PDF_MARGIN, PDF_HEIGHT_SEIZE, {
```
pgurusinga, line 15: @aaschlote I just found out about this API https://github.com/foliojs/pdfkit/issues/742#issuecomment-340451968 
```js
const range = doc.bufferedPageRange();

return range.start + range.count;
```

so we can dynamically add the page number. what do u think?

aaschlote: works perfect!

app/services/pdf/fluggastrechte/__test__/setPdfMetadata.test.ts
aaschlote: The creation date metadata should be either "Date" or "undefined", so I changed the test to check if it's not undefined, otherwise it'll be instance of Date.

image

pgurusinga: hmm... but when is this gonna be undefined? 🤔 it's the first thing of the initialisation I guess, but not a blocker

aaschlote: So I expect the `CreationDate` must be set any value, in case it's not, the value if undefined

PR #1274, opened 2024-10-08
Generate pdfFill function to return a buffer

aaschlote:


app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -46,7 +47,11 @@ export async function pdfDownloadLoader({ request }: LoaderFunctionArgs) {
...
   if (_.isEmpty(userData)) return redirect(flowId);
 
-  return new Response(await (await pdfFunction(userData)).save(), {
-    headers: createHeaders(filenameFunction()),
+  const fileContent = await pdfFunction(userData);
+  const fileSize = fileContent.length;
+  const filename = filenameFunction();
+
+  return new Response(fileContent, {
+    headers: createHeaders(filename, fileSize),
```
chohner, line 55: nice job :)

app/services/pdf/beratungshilfe/index.ts
```diff
@@ -67,5 +67,5 @@ export async function beratungshilfePdfFromUserdata(
     filledPdf,
     await pdfFromReact(Handout(userData, "Merkblatt")),
   );
-  return filledPdf;
+  return await filledPdf.save();
```
chohner, line 70: without known your constraints, this looks like a API change for the worse:
- tests are more complicated
- tests are slower
- concerns are more mixed

Could you motivate why the call-site shouldn't be responsible for further processing the returned `PdfDocument`?

app/routes/shared/pdfDownloadLoader.ts
aaschlote: kudos to @pgurusinga he added it, I just copied&pasta 😆

app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -29,6 +29,7 @@ export function createHeaders(filename: string) {
   return {
     "Content-Type": "application/pdf",
     "Content-Disposition": `inline; filename=${encodeURIComponent(filename)}`,
+    "Content-Length": fileContentLength.toString(), // Add content length
```
chohner, line 32: remove the comment?

chohner: also - add a test for the header property? 😇

app/services/pdf/beratungshilfe/index.ts
chohner: aha! ok that makes sense. Could we wrap the existing functions in a `pdfDocToArrayBuffer()` function, so we can keep the tests untouched?

aaschlote: do you mean do something like this:

```
pdfFunction: (userData) => pdfDocToArrayBuffer(beratungshilfePdfFromUserdata(userData)
```

chohner (approved):
nice

app/services/pdf/beratungshilfe/index.ts
chohner: yes. although looking at the function we could almost also just do it inline 😅
you call

app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -46,7 +41,11 @@ export async function pdfDownloadLoader({ request }: LoaderFunctionArgs) {
...
   if (_.isEmpty(userData)) return redirect(flowId);
 
-  return new Response(await (await pdfFunction(userData)).save(), {
-    headers: createHeaders(filenameFunction()),
+  const fileContent = await pdfFunction(userData);
+  const fileSize = fileContent.length;
+  const filename = filenameFunction();
+
+  return new Response(fileContent, {
+    headers: createPdfResponseHeaders(filename, fileSize),
```
pgurusinga, line 49: much better naming 👍

pgurusinga (approved):

PR #1286, opened 2024-10-11
infra(buildAppImage pipeline failure slack message)

Spencer6497:
- **add alert-pipeline-failure step to buildAppImage**
- **(revertme) write failing test for CI**


.github/workflows/ci-pipeline.yml
```diff
@@ -198,3 +198,24 @@ jobs:
...
+        code-quality,
+        get-content-file,
+        verify-local-e2e,
+        build-push-app-image,
+        deploy-staging,
+        deploy-preview,
+        verify-preview-e2e,
+        deploy-production,
+        test-production-text,
+      ]
```
chohner, line 215: Wouldn't referencing just `test-production-text` be enough?

chohner (approved):
lgtm, little comment

.github/workflows/ci-pipeline.yml
Spencer6497: unfortunately not; if any of the referenced dependencies fail or are skipped. the referrers don't execute. i.e. if `verify-local-e2e` fails, none of the dependent jobs execute, including the notif

PR #1290, opened 2024-10-11
Add pilot court on airport data source and improve functions to get airport data

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1207980403231510

### Changes

- Add pilot course on airport data source based on the [ticket](https://app.asana.com/0/1207022303033556/1208381228755863)
- Create function getAirportByIataCode to get airport type and re-use in existing functions
- Fix typos


app/services/airports/getAirportByIataCode.ts
```diff
@@ -0,0 +1,14 @@
+import airports from "data/airports/data.json";
+import type { Airport } from "./type";
+
+export function getAirportByIataCode(
+  airportIataCode: string,
+): Airport | undefined {
```
pgurusinga, line 6: when would this be `undefined`?

app/services/airports/getAirportByIataCode.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (airportIataCode.length === 0) {
+    return undefined;
+  }
```
pgurusinga, line 9: if missing IATA code, we should throw or who handles this undefined?

app/services/airports/isEuropeanUnionAirport.ts
```diff
@@ -37,14 +37,10 @@ const EU_COUNTRIES = [
...
 
-function getCountryCodeByIata(airportIata: string | undefined) {
-  return airports.find((airport) => airport.iata === airportIata)?.country_code;
-}
-
 export function isEuropeanUnionAirport(
   airportCode: string | undefined,
 ): Result<boolean, string> {
-  const airportCountry = getCountryCodeByIata(airportCode);
+  const airportCountry = getAirportByIataCode(airportCode ?? "")?.country_code;
```
pgurusinga, line 43: if this param always defaults to an empty string, should we move the fallback to its function definition?

app/services/airports/getAirportByIataCode.ts
aaschlote: When it gives an unknown iata code, like `XXX`

app/services/airports/getAirportByIataCode.ts
aaschlote: I would prefer undefined, otherwise I need to handle the exception for every function how uses this function

app/services/airports/isEuropeanUnionAirport.ts
aaschlote: Do you mean, in the line 41?

pgurusinga: Why did you include an empty string here? It seems strange to have a fallback when the function `getAirportByIataCode` can handle an invalid parameter

aaschlote: The airport data comes from context of the Vorabcheck, so the value can be undefined. I created the function `getAirportByIataCode` to accept only string, so I need to fallback to empty string here, where it always going to return undefined. What I can do?

1. Change the function `getAirportByIataCode` to accept also `undefined` values, but it'll return always `undefined`
2. In the function `isEuropeanUnionAirport` change to in case the airportCode is undefined, return `Result.err("Airport not found")`

pgurusinga (approved):

app/services/airports/isEuropeanUnionAirport.ts
aaschlote: I implemented the second option

PR #1291, opened 2024-10-14
fix: fetchAllFormFields 

ekl176:
fetchAllFormFields takes staging and prod locales into account on staging and dev

---
  - https://app.asana.com/0/0/1208349447113947


Spencer6497 (approved):
🚀

app/services/cms/fetchAllFormFields.ts
```diff
@@ -1,23 +1,49 @@
...
+    filters,
+    populate,
+    locale: "all",
+  });
 
+  const nonEmptyEntries = strapiEntries
+    .filter((formFlowPage) => formFlowPage !== null)
+    .filter(({ attributes: { stepId, form } }) => form.length > 0 && stepId);
+
+  const [staging, result] = _.partition(
```
chohner, line 27: I feel like `result` is not very descriptive

app/services/cms/fetchAllFormFields.ts
```diff
@@ -1,23 +1,49 @@
...
+
+  if (environment !== "production") {
+    // inject staging flowpages
+    _.merge(result, staging);
+  }
+
+  return result;
+}
+
+function fromStrapiSchemasToFormFields(
```
chohner, line 40: `formFieldsFromSchema`?

app/services/cms/fetchAllFormFields.ts
```diff
@@ -1,23 +1,49 @@
...
+    .filter((formFlowPage) => formFlowPage !== null)
+    .filter(({ attributes: { stepId, form } }) => form.length > 0 && stepId);
+
+  const [staging, result] = _.partition(
+    nonEmptyEntries,
+    ({ attributes: { locale } }) => locale === "sg",
+  ).map(fromStrapiSchemasToFormFields);
+
+  if (environment !== "production") {
+    // inject staging flowpages
```
chohner, line 33: nit: probably can be deleted?

app/services/cms/fetchAllFormFields.ts
```diff
@@ -1,23 +1,49 @@
...
+    locale: "all",
+  });
 
+  const nonEmptyEntries = strapiEntries
+    .filter((formFlowPage) => formFlowPage !== null)
+    .filter(({ attributes: { stepId, form } }) => form.length > 0 && stepId);
+
+  const [staging, result] = _.partition(
+    nonEmptyEntries,
+    ({ attributes: { locale } }) => locale === "sg",
```
chohner, line 29: nitpick: we could invert the check, so the first outcome is the non-staging data?

app/services/cms/__test__/index.test.ts
```diff
@@ -73,6 +73,98 @@ describe("services/cms", () => {
       });
     });
 
+    test("overwrites formfields from staging", async () => {
```
chohner, line 76: nice!

chohner (approved):
lgtm, few nitpicks

app/services/cms/fetchAllFormFields.ts
ekl176: had the feeling this is the safer option. We'd need to touch this everytime we introduce a new prod locale

app/services/cms/fetchAllFormFields.ts
ekl176: I wasn't sure if the merge itself is descriptive enough to explain why we do this

app/services/cms/fetchAllFormFields.ts
ekl176: at first the result was named prod but found it confusing when we inject the staging fields in case we are not in prod.

chohner: Maybe `[NonStagingEntries, StagingEntries]`? or `DefaultEntries`?

PR #1295, opened 2024-10-14
refactor(PKH Eigenes Einkommen Unterhalt Frage)

Spencer6497:
This PR removes the "Beziehen Sie Unterhalt" question from the Einkuenfte subflow (as we're asking the user during Antragstellende Person the same question), but leaves the question in for Einkuenfte - Partner:in.

- **rename flow.ts -> xStateConfig.ts**
- **skip unterhalt question in eigenes einkommen flow entirely**
- **use unterhalt context from grundvoraussetzungen instead of eigenes Einkommen, rename supportDone -> partnerSupportDone and fix logic/tests**
- **remove receivesSupport and supportAmount entirely, finish plumbing for partner-receivesSupport & partner-supportAmount**
- **Fix e2e test, update unit tests, update PDF logic**


chohner (approved):
brief look but lgtm

PR #1312, opened 2024-10-17
add setCacheControlHeader for no-cache if trackconsent is set or is f…

m0dh4x:
…lowid

### DIFF
--- a/app/__test__/headers.test.ts
+++ b/app/__test__/headers.test.ts
@@ -1,3 +1,5 @@
+import { random } from "lodash";
+import { flowIds } from "~/flows/flowIds";
 import { headers } from "~/root";
 
 const mockHeaders = (headers: Record<string, string>): Headers => {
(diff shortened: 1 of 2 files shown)


app/__test__/headers.test.ts
```diff
@@ -12,28 +13,38 @@ const defaultHeaders = {
     "accelerometer=(),ambient-light-sensor=(),autoplay=(),battery=(),camera=(),display-capture=(),document-domain=(),encrypted-media=(),fullscreen=(),gamepad=(),geolocation=(),gyroscope=(),layout-animations=(self),legacy-image-formats=(self),magnetometer=(),microphone=(),midi=(),oversized-images=(self),payment=(),picture-in-picture=(),publickey-credentials-get=(),speaker-selection=(),sync-xhr=(self),unoptimized-images=(self),unsized-media=(self),usb=(),screen-wake-lock=(),web-share=(),xr-spatial-tracking=()",
 };
 
+vi.mock("~/util/shouldSetCacheControlHeader", () => ({
+  shouldSetCacheControlHeader: vi.fn(),
+}));
+
 describe("Headers", () => {
+  beforeEach(() => {
```
aaschlote, line 21: do we need this beforeEach?

aaschlote (approved):
LGTM

app/__test__/headers.test.ts
m0dh4x: hmm case of cleaning the mock states? i would say yes. whats your opinion?

aaschlote: but you are cleaning before start every test, in the last test you are not cleaning, so you are keeping mocked. And if you remove this block and run all the test it won't change anything, so it makes this block redundant.

Either you remove it (where it doesn't affect any test at all) or you change it to afterEach, so you clean the mock after every test runs.

m0dh4x: as already discussed. i agree :)

pgurusinga (approved):

app/util/shouldSetCacheControlHeader.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+const isTrackingConsentSet = (loaderHeaders: Headers): boolean => {
+  return loaderHeaders.get("trackingConsentSet") === "true";
+};
+
+export const shouldSetCacheControlHeader = (
+  loaderHeaders: Headers,
+): boolean => {
+  const pathname = loaderHeaders.get("pathname") || "";
```
chohner, line 15: I think we mostly us nullish coalescing (`??`) here. However, I think removing the fallback for `if (!pathname) return false` is more precise

app/root.tsx
```diff
@@ -145,7 +146,12 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
       bannerState:
         getFeedbackBannerState(mainSession, pathname) ?? BannerState.ShowRating,
     },
-    { headers: { trackingConsentSet: String(trackingConsent === undefined) } },
+    {
+      headers: {
+        trackingConsentSet: String(trackingConsent === undefined),
+        pathname,
+      },
+    },
```
chohner, line 154: Using the loader headers as a function interface seems very cumbersome, especially in testing.

Maybe simply calling `shouldSetCacheControlHeader` in the loader and set a single `shouldCache` might be easier?

m0dh4x: fair point :) i will refactor this

m0dh4x: i have pushed the changes

app/root.tsx
```diff
@@ -145,7 +148,7 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
       bannerState:
         getFeedbackBannerState(mainSession, pathname) ?? BannerState.ShowRating,
     },
-    { headers: { trackingConsentSet: String(trackingConsent === undefined) } },
+    { headers: { shouldCache: String(shouldCache) } },
```
chohner, line 151: I think the variable is inverted and this should be `shouldNotCache`? 😅

chohner (approved):
lgtm, just variable naming

app/root.tsx
m0dh4x: true

PR #1315, opened 2024-10-17
refactor: Flow transition 

frederike-ramin:
I did four changes:
- targetFlowId is not necessary as the config only exists on the flow you want it - no need to check.
- the flowController had a fluggastrechte-specific name, leading me to think it was domain-specific and including it in [this PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1270) however, jsut the variable was specifically named
- move redirect to the top - let's redirect as soon as we can
- "flowTransisiotnValidation" very much does not sound like a "session" thing to do, so I moved it into the "flow" service.

Feedabck welcome.


pgurusinga (approved):
it makes sense that the `targetFlowId` is removed since it's exclusive to the flow configuration. thanks for taking care of this refactoring

app/routes/shared/formular.server.ts
```diff
@@ -115,6 +115,19 @@ export const loader = async ({
   if (!flowController.isReachable(stepId))
     return redirectDocument(flowController.getInitial());
 
+  const flowTransitionConfig = getFlowTransitionConfig(currentFlow);
```
aaschlote, line 118: nice one, I was thinking to move this logic to the top of the loader. Well done 👏

aaschlote (approved):
LGTM

PR #1322, opened 2024-10-18
docs(adr): prevent caching on FlowId pages

m0dh4x:
- maintain user input state


doc/adr/0013-no-caching-on-routes-with-flowid.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Context
+
+Related to ADR 10, we have still encountered problems on the Formular and Vorabcheck pages.
+
+When the user interacts with our Formular or Vorabcheck and goes through the pages, and then the user clicks the browser's back button, the browser cache will serve the previously cached page version, and we sometimes lose the state of the user's inputs, especially select fields.
+
+## Decision
+
+We decided to ask the browser to not cache pages also if the user is on one of hour FlowId Pages and we setting the `Cache-Control: no-store` header.
```
pgurusinga, line 17:
```suggestion
We decided to ask the browser to not cache pages also if the user is on one of our FlowId Pages and we set the `Cache-Control: no-store` header.
```

pgurusinga (approved):

doc/adr/0013-no-caching-on-routes-with-flowid.md
```diff
@@ -0,0 +1,25 @@
+# 13. No Caching on routes with FlowId
+
+Date: 2024-10-18
+
+## Status
+
+proposed
```
aaschlote, line 7: Accepted

doc/adr/0013-no-caching-on-routes-with-flowid.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Context
+
+Related to ADR 10, we have still encountered problems on the Formular and Vorabcheck pages.
```
aaschlote, line 11: can you link the ADR 10?

doc/adr/0013-no-caching-on-routes-with-flowid.md
aaschlote: small change: ``in the response header``

aaschlote (approved):
LGTM

chohner (approved):
content lgtm

doc/adr/0013-no-caching-on-routes-with-flowid.md
m0dh4x: sure :)

doc/adr/0013-no-caching-on-routes-with-flowid.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Related to [ADR 10](./0010-no-caching-while-cookie-banner-visible.md), we have still encountered problems on the Formular and Vorabcheck pages.
+
+When the user interacts with our Formular or Vorabcheck and goes through the pages, and then the user clicks the browser's back button, the browser cache will serve the previously cached page version, and we sometimes lose the state of the user's inputs, especially select fields.
```
chohner, line 13: Actually, the only reliable reproduction for me was the combobox, not any native input. Which makes some sense as it contains JS state that probable isn't re-hydrated/refreshed when pulled from cache? maybe worth adding...

aaschlote: @chohner this is not correctly, we made a meeting on 30.09 and I showed to you that the error happens not because the re-hydrated or refreshed of the page, but the cache approach for the browser. We saw the error happening together for the `Select` component and I entered the information in the following [ticket](https://app.asana.com/0/1207022303033556/1208416708375009).

aaschlote: I recall you mentioned this issue during a Dev Exchange, which is why we had the meeting on 30.09. During the meeting, I demonstrated that the problem wasn’t related to the React components, but rather to the browser’s caching mechanism. Specifically, when you visit a page without any data, the browser caches that state. When you return to the page using the back button, it shows the cached version, as described in the ADR.

I recommend opening a ticket and following the instructions outlined there on how to reproduce the issue.

If you'd like, I'm happy to schedule a follow-up meeting to go over the error again and explain further if needed.

m0dh4x: Hey, thank you for sharing your thoughts. It seems there are differing perspectives on the issue. It's important to note that "this is a strict intervention into the bfcache behavior and it compounds over time." It could be beneficial to explore this further, as understanding the root cause might help us find a more effective solution. From my perspective it feels a bit like "Kanonen auf Spatzen" to disable the cache. If time permits, diving deeper into the problem could provide valuable insights.

aaschlote: To be honest, this topic has been a long time withe team ZOV and has been over and over explained and discussed with other developers.

I'm quite upset and disappointed we still can't agree with the implementation for this problem, even many times discussed about it 😞

m0dh4x: At the moment I think that's fine and we use these quick fix, but we should bring it up again from time to time. Deal?

PR #1324, opened 2024-10-18
pkh pdf fill antragstellende person

rbrtrfl:


app/services/pdf/fillPdfFieldOrMoveToAttachment.ts
```diff
@@ -0,0 +1,32 @@
...
+  pdfFieldName: keyof TGeneratedPDFType;
+  pdfFieldValue: string | undefined;
+  pdfFieldMaxChars: number;
+  attachmentTitle: string;
+  pdfValues: TGeneratedPDFType;
+  attachment: AttachmentEntries;
+  attachmentPageHint?: string;
+};
+
+export function fillPdfFieldOrMoveToAttachment<TGeneratedPDFType>({
```
rbrtrfl, line 13: Since we are repeatedly doing the same thing: Filling form fields which have a character limit and potentially move the content to the attachment page - I thought we could move this logic into a versatile function.

wdyt @chohner @Spencer6497 

future improvement: add number of allowed line breaks for multiline form fields

app/services/pdf/fillPdfFieldOrMoveToAttachment.ts
```diff
@@ -0,0 +1,32 @@
...
+  attachmentTitle,
+  pdfValues,
+  attachment,
+  attachmentPageHint = newPageHint,
+}: FillPdfFieldOrMoveToAttachment<TGeneratedPDFType>): {
+  pdfValues: TGeneratedPDFType;
+  attachment: AttachmentEntries;
+} {
+  if (pdfFieldValue && pdfFieldValue.length > pdfFieldMaxChars) {
+    (pdfValues[pdfFieldName] as { value: string }).value = attachmentPageHint;
```
Spencer6497, line 26: why do we need to typecast here?

app/services/pdf/fillPdfFieldOrMoveToAttachment.ts
Spencer6497: seems sensible to me; though, isn't another big use case when there are >1 entries of something and we append it to the attachment _regardless_ of field size? maybe we should include that case here as a one-stop-shop

Spencer6497 (approved):

PR #1327, opened 2024-10-21
Unify pdf field filling incl auto-size

rbrtrfl:
Since we are repeatedly doing the same thing: Filling form fields which have a character limit and potentially move the content to the attachment page - I thought we could move this logic into a versatile function.

possible iterations:
- add number of allowed line breaks for multiline form fields (e.g. Bankkonten)
- content for fields that allow multiple entries (also e.g. Bankkonten) is moved to attachment when `entries > 1`

questionable cases:
where there are multiple form fields for the same content e.g.:
- section H 6.
- section I (Sonstige Zahlungsverpflichtungen) & J (Besondere Belastungen)


app/services/pdf/prozesskostenhilfe/A_person.ts
```diff
@@ -40,66 +41,56 @@ export const fillPerson: PkhPdfFillFunction = ({ userData, pdfValues }) => {
...
-  if (nameVornameString.length > NAME_VORNAME_FIELD_MAX_CHARS) {
-    attachment.push({
-      title: "Name, Vorname, ggf. Geburtsname",
-      text: nameVornameString,
-    });
-    pdfValues.nameVornameggfGeburtsname.value = newPageHint;
-  } else {
-    pdfValues.nameVornameggfGeburtsname.value = nameVornameString;
-  }
+  fillPdfFieldOrMoveToAttachment<ProzesskostenhilfePDF>({
```
chohner, line 44: The explicit `ProzesskostenhilfePDF` type shouldn't be necessary, as it can be inferred from the passed `pdfValues` prop

PR #1333, opened 2024-10-22
Both streams: Remove accessibility tests of forms flow pages

aaschlote:
### Changes

- Remove accessibility tests of all the Formular pages
- Keep the accessibility tests on start, abgabe and übersicht pages
- Remove accessibility tests of Vorabcheck Fluggasterechte and Geldeinklage


aaschlote:
> Amazing, look at the speed ⚡ image
> 
> I wonder if we should add the promised "in-depth" e2e test for form pages (basically testing all the things we didnt until now: validation, back button, navigation, ...) in this PR or add a ticket?

I believe we can make it in a new ticket

PR #1339, opened 2024-10-22
Doc(adr):  move formFields from CMS into app

chohner:
I recommend reading the ADR with rendered markdown to display diagrams:

https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/adr_pageConfig_inside_app/doc/adr/0013-pageConfig-inside-application.md


doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Pros:
+
+- Full decoupling of domains (flow configuration vs content)
+- Follows mental model of domain driven architecture
+- App works without CMS (faster prototyping...)
+
+### Cons:
+
+- Fieldnames and definition in the CMS still need to match the ones in the app
```
rbrtrfl, line 110: Maybe add an explanation what happens when they don't match?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+
+1. Add form field to new page
+2. Trigger content release (the new page wouldn't change as the app is in charge of whats shown)
+3. Field is moved to its new location in the `PageConfig`
+4. Trigger app release (now the form field will appear on the new page)
+5. Remove form field from the old page in CMS
+
+#### Syncronizing form fields
+
+The downside of doubling field names & definition could be mitigated by synchronizing the CMS with the app. This could be either done by:
```
rbrtrfl, line 127: Could you describe what is meant by 'definition'?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Publishing the fieldnames via an endpoint that the CMS consumes
+- Pushing changes to the CMS on deploy (probably via a simple POST request)
```
rbrtrfl, line 130: Does that mean that field names could not be written into the CMS manually, but the codebase would be ssot?

rbrtrfl (commented):
interesting concept. I'm not sure I'd be in favour of implementing it without the “synchronize form fields” feature 🤷

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+  ```
+- `SchemaStore`: A collection that can return a `stepSchema` based on a list of `fieldnames`. Our current function is called `validatorForFieldnames(fieldnames)`.
+
+#### Diagram
+
+```mermaid
+sequenceDiagram
+
+Loader->>CMS: GET(slug)
+CMS-->>Loader: content, fieldnames, formLabels
```
frederike-ramin, line 32: Why is it called formLabel? Isn't it a fieldLabel?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+Loader->>Frontend: content, fieldnames, formLabels
```
frederike-ramin, line 33: Is there a better word than "Frontend"? It's very generic. Is this the ValidatedForm maybe?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+
+- `fieldname`: the name of a field in the form, e.g. `hasAnwalt`
+- `fieldSchema`: Combination of fieldname and its schema, e.g. `hasAnwalt: z.enum(["yes", "no"]) }`
+- `stepSchema`: All schemas required for a single page, e.g.:
+  ```typescript
+  {
+    hasAnwalt: z.enum(["yes", "no"]) },
+    hasContacted: z.enum(["yes", "no"]) },
+  }
+  ```
```
frederike-ramin, line 23: I don't think that concept currently exists? The schemaStore merely returns a list of validations?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+#### Diagram
+
+```mermaid
+sequenceDiagram
+
+Loader->>CMS: GET(slug)
+CMS-->>Loader: content, fieldnames, formLabels
+Loader->>Frontend: content, fieldnames, formLabels
+Frontend->>SchemaStore: fieldnames
+SchemaStore->>Frontend: stepSchema
```
frederike-ramin, line 35: Maybe to make it more clear a self-loop with "validation" would help? To be clear what the schemas are even used for in the frontend?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+Frontend->>Backend: POST(formData)
+Backend->>SchemaStore: fieldnames
+SchemaStore->>Backend: stepSchema
```
frederike-ramin, line 38: Why don't you call this "Backend" the Action?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+Note that the `fieldnames` are passed from the CMS via the loader and front/back-end into the `SchemaStore`.
+
+#### Downsides
+
+- Its hard to understand where which fields are filled out on what page (requires manually checking the CMS)
+- The app **cannot** know why fields where seen when traversing a flow (pruner needs to request the CMS)
```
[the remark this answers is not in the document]
frederike-ramin: I guess it's more "*if* fields are shown"?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+- The app **cannot** work without the CMS
```
frederike-ramin, line 47: Is this really the only point that keeps us from showing pages without content? If we do the change, this problem doesn't exist anymore?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+
+```mermaid
+sequenceDiagram
+
+Loader->>PageConfig: slug
+PageConfig->>Loader: fieldnames
+Loader->>CMS: GET(slug, fieldnames)
+CMS-->>Loader: content, formLabels
+Loader->>Frontend:  content, formLabels
+Frontend->>PageConfig: slug
```
frederike-ramin, line 79: I know it's not for this ADR but I think it's pretty weird that the frontend has to pull this if it could just as well be pushed from the loader.

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+  },
+};
+```
+
+## Consequences
+
+### Pros:
+
+- Full decoupling of domains (flow configuration vs content)
+- Follows mental model of domain driven architecture
```
frederike-ramin, line 105: Is that so? How?

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Context
+
+Currently, the information about what field appears on what page is **solely** encoded in the CMS. In order to render a page, the app has to fetch the content for a given URL from the CMS, which may contain a list of fields and their labels (next to the page content).
+
+#### Terms
+
+- `fieldname`: the name of a field in the form, e.g. `hasAnwalt`
+- `fieldSchema`: Combination of fieldname and its schema, e.g. `hasAnwalt: z.enum(["yes", "no"]) }`
+- `stepSchema`: All schemas required for a single page, e.g.:
```
frederike-ramin, line 17: I'm still wondering if we could go one step further and not only represent pages this way but pages within whole subflows. Such as:

  ```typescript
  anwaltlicheVertretung: {
    anwaltEntscheidung: {
      hasAnwalt: z.enum(["yes", "no"]) },
      hasContacted: z.enum(["yes", "no"]) },
    }
  }
  ```

That would mean that we could validate whole sub-areas of the flow by checking only for `anwaltlicheVertretung` and gathering all fields in there. 
I don't know, probably a thought for the future as this is not that well thought-out, still wanted to leave it here.

frederike-ramin (commented):
Probably too tired for doing a PR review, but here you go. ;)

doc/adr/0013-pageConfig-inside-application.md
chohner: I thought staying generic might be less confusing than specifying the exact consumer of each (between the nested page/form/field components)...

doc/adr/0013-pageConfig-inside-application.md
chohner: The word does not exist but we do build a validation for exactly the formData object (not a list)

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,130 @@
...
+Now, any component can receive a `stepSchema` directly from the `PageConfig` by passing the `slug`:
+
+```mermaid
+sequenceDiagram
+
+Loader->>PageConfig: slug
+PageConfig->>Loader: fieldnames
+Loader->>CMS: GET(slug, fieldnames)
+CMS-->>Loader: content, formLabels
+Loader->>Frontend:  content, formLabels
```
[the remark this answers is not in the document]
chohner: frontend component can get the pathname via `useLocation()`, no need to pass it imo

doc/adr/0013-pageConfig-inside-application.md
chohner: yes, for both cases

doc/adr/0013-pageConfig-inside-application.md
chohner: I.. think so? can't think of anything else 🤔

doc/adr/0013-pageConfig-inside-application.md
chohner: I'm not sure what exactly "this" is referring to but the `stepSchema` is not serializable. the only thing the loader is pushing is text content, the component is fetching the `stepSchema` from the `PageConfig` (using the `pathname`)

doc/adr/0013-pageConfig-inside-application.md
chohner: rephrased, does it make more sense?

doc/adr/0013-pageConfig-inside-application.md
chohner: I'm not sure I follow - but I have to admit, nested schemas (and arrays) are still not prototyped using this approach.

Right now I think we could either encode them the way you showed (which matches our current `userInputSchemas`) or do it like in the CMS (eg `anwaltEntscheidung.hasAnwalt: z.enum(["yes", "no"]) }`).

aaschlote (commented):
I'm bit confusing about the two parts:

1. How the associate the fieldNames from the app with the CMS
2. How these changes will improve the developer experience. The changes sounds huge and it's not clear if it worth to do it. I would need some examples before to approve it.

Another thing we need to think, it's about how to move all the contexts to `stepSchema` (please think about a better name), because it exists a lot of fields in the current flows, specially for beratungshilfe and flugggastrechte. How we will need to do all the change without break the application.

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,137 @@
...
+A new `PageConfig` abstraction should be added, that links the `stepId` and `stepSchema` of each page and would replace our current `SchemaStore` (aka `context`/`userInputSchema`):
+
+```typescript
+type StepSchema = Record<string, z.ZodType>;
+type PageConfig = Record<string, { stepId: string; stepSchema: StepSchema }>;
+
+const pages = {
+  anwalt: {
+    stepId: "/anwalt",
+    stepSchema: { hasAnwalt: z.enum(["yes", "no"]) },
```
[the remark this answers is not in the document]
chohner: Its the collection of all `fieldSchemas` for specific step. For Example:

```typescript
const airportStepSchema = {
	startAirport: z.string(),
	endAirport: z.string(),
}
```

`startAirport` and `endAirport` would each be a `fieldSchema` and `airportsSchema` would be a `stepSchema`. Maybe `pageSchema` is a better name?

Note: This construct already exists as the return value of `validatorFromFieldnames` but is a bit hidden

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,137 @@
...
+### Pros:
+
+- Domain driven architecture: Full decoupling of flow configuration from content
+- Reduced mental load when reasoning about form fields and their position inside the flow
+- App works without CMS (pruner, prototyping, ...)
+
+### Cons:
+
+- Field names and definition in the CMS still need to match the ones in the app. However, on a mismatch we would "just" have missing labels, rather than full app breakage
+- Cleanly moving a field from between existing pages might become more complicated (See below)
```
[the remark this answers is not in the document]
chohner: The app would be in charge of what fields are rendered, not the CMS. The CMS only provides content (labels)

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,137 @@
...
+    [pages.anwalt.stepId]: {
+      meta: { stepSchema: pages.anwalt.stepSchema },
+    },
+  },
+};
+```
+
+## Consequences
+
+### Pros:
```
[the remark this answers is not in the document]
chohner: I think having an understanding of which page fills out which field is going to be useful when reasoning about flow logic. And creating new flows should be much faster.

doc/adr/0013-pageConfig-inside-application.md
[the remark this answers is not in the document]
chohner: Generally I think our data model in strapi also needs to be refactored, but I didn't want to include this in this ADR (and also not create another ADR, we already have a lot...).

So for now I propose to leave as is because each flow page in Strapi already contains the correct labels etc.

But I could be convinced to park this ADR in favor of restructuring our CMS first

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,137 @@
...
+
+- Domain driven architecture: Full decoupling of flow configuration from content
+- Reduced mental load when reasoning about form fields and their position inside the flow
```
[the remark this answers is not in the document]
chohner: When writing business logic, we often need to argue about what variable is filled out on what page. but that information is not present in the app and often requires a check in the CMS

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,137 @@
...
+#### Diagram
+
+Now, any component can receive a `stepSchema` directly from the `PageConfig` by passing the `pathname`:
+
+```mermaid
+sequenceDiagram
+
+Loader->>PageConfig: pathname
+PageConfig->>Loader: fieldNames
+Loader->>CMS: GET(pathname, fieldNames)
```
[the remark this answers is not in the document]
chohner: I agree this is confusing because we probably wouldn't send the fieldnames right now (as they already exist on each flow page). The schemas continue to live in the app just like they do now

doc/adr/0013-pageConfig-inside-application.md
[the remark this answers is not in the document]
aaschlote: Do you mean about the `flow logic`, only if I use the .ts instead of .json for the configuration of the xstate?

doc/adr/0013-pageConfig-inside-application.md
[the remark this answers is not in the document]
aaschlote: So if the app request a field that it doesn't exist in the CMS, it doesn't break? And if it gets a response of a field that it doesn't exist in the configuration, also doesn't it break?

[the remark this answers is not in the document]
chohner: If the app requests a field that doesn't exist it should log a warning and show the ids (instead of the real labels from the cms). if the cms specifies a field that doesn't exist in the app nothing happens

doc/adr/0013-pageConfig-inside-application.md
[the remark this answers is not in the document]
chohner: I dont have a strong opinion on the order, however I have the feeling restructuring the CMS is probably going to be a bigger task and I think the benefits of this are useful already.

But I would be fine with either order

doc/adr/0013-pageConfig-inside-application.md
frederike-ramin: I think that's essentially the same thing? With the first option just giving us more variety (because you can define things on the object-level)?
But yes, this is probably just one step further and not part of this ADR necessarily

doc/adr/0013-pageConfig-inside-application.md
```diff
@@ -0,0 +1,137 @@
...
+};
+```
+
+## Consequences
+
+### Pros:
+
+- Domain driven architecture: Full decoupling of flow configuration from content
+- Reduced mental load when reasoning about form fields and their position inside the flow
+- App works without CMS (pruner, prototyping, ...)
```
pgurusinga, line 113: I am unsure if "App works without CMS" is a good thing. If the app can function without the CMS, does that mean it might not alert us to issues with missing content? Since content is essential to delivering value to users, wouldn’t it be better for the app to indicate when the CMS isn’t working or content is missing? This way, we’d catch potential issues during testing or production instead of relying on users to report them. or maybe I understood it differently?

chohner: Of course altering of missing content should still happen. but we shouldn't need to crash the whole application, especially when prototyping. Also the pruner shouldn't need to rely on the CMS

doc/adr/0013-pageConfig-inside-application.md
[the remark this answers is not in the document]
chohner: renamed to `pageSchema`

doc/adr/0022-pageConfig-inside-application.md
```diff
@@ -0,0 +1,148 @@
...
+```typescript
+const pages = {
+  anwaltForAdult: {
+    stepId: "/anwalt",
+    pageSchema: z
+      .object(pick(userDataSchemas, [hasLawyer, isAdult]))
+      .refine(
+        ({ hasLawyer, isAdult }) =>
+          !(isAdult === "yes" && hasLawyer === "no",
+          { message: "Under age applicants need a lawyer" }),
```
Spencer6497, line 90: how often are we actually validating an entire page, instead of individual fields?

doc/adr/0022-pageConfig-inside-application.md
```diff
@@ -0,0 +1,148 @@
...
+
+Note that `fieldNames` are passed from the CMS via the loader + front/back-end into the `SchemaStore`.
+
+#### Downsides
+
+- Resilience: The app **cannot** work without the CMS providing the `fieldNames`
+- Slow prototyping of new flows, as it requires addition of pages to the CMS even without content
+- Mental load: Understanding where a field is filled out requires manually checking in the CMS
+- Additional traffic: When the app needs to know about fields on pages, it needs to make an additional request to the CMS (for example in the pruner)
+- Additional complexity: Unchecked checkboxes and radio groups are not sent in html forms. To still check for them, we inject a hidden input to allow the backend to verify that such an input _should_ be there (currently affects our `Checkbox`, `RadioGroup` and `TileGroup` components). See this [StackOverflow article](https://stackoverflow.com/a/1992745) for more context.
```
Spencer6497, line 56: wouldn't this still need to happen?

Spencer6497 (approved):
I like it! Though, since this is a massive architectural change, maybe we could add a rough outline of the migration plan in the "Proposal" section

doc/adr/0022-pageConfig-inside-application.md
chohner: there are quite a few already in fluggastrechte (check `multiFieldsValidation`) and I expect there might be more cases that we haven't really thought about because it wasn't easily possible

doc/adr/0022-pageConfig-inside-application.md
chohner: I don't believe so, since the app can rely on the pageSchema to check whether a required checkbox is missing in the form data.

doc/adr/0022-pageConfig-inside-application.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Domain driven architecture: Full decoupling of flow configuration from content
+- Reduced mental load when reasoning about form fields and their position inside the flow
+- App works without CMS (pruner, prototyping, ...)
+- Faster staging performance due to reduced lookups
+- Simplified per-page validation
+
+### Cons:
+
+- Field names and definition in the CMS still need to match the ones in the app. However, a mismatch just results in missing labels, not breaking the full app
```
pgurusinga, line 129: Could we add a safety net that automatically flags or fills in missing labels so users never see incomplete information? For example, what if we used a default label when the CMS and app definitions don't match? 
I prefer to fail the app so the developers can catch it and fix it asap rather than having a missing label for weeks and no one notices about it. (happened with the inline notice went missing for days)

chohner: the existing integration test should warn about this, yes

pgurusinga (approved):
🎉

PR #1354, opened 2024-10-28
feat(CSPHeaders): strengthen headers

chohner:
- adds `form-action` & `upgrade-insecure-requests`
- adds optional `CSP_REPORT_URI` env & sets up CSP reporting
- removes `unsafe-inline` -> pre-CSP3 browsers will block JS (which we support)
- auto-generated trusted domains from env variables
- minor refactoring in `cspHeader()` & tests


app/entry.server.tsx
```diff
@@ -97,9 +98,20 @@ function handleBrowserRequest(
   return new Promise((resolve, reject) => {
     let didError = false;
     const cspNonce = generateNonce();
+    const trustedDomains = [
+      configWeb().POSTHOG_API_HOST,
+      configWeb().SENTRY_DSN,
+    ]
+      .filter((url) => url !== undefined)
+      .map((url) => new URL(url).origin);
```
aaschlote, line 106: as the values are coming from env variables, in case the URL is wrong, the declaration `new URL(url)` will thrown an exception. Should we handle it or crash the system in case the environment variables are wrong?

app/entry.server.tsx
```diff
@@ -97,9 +98,20 @@ function handleBrowserRequest(
   return new Promise((resolve, reject) => {
     let didError = false;
     const cspNonce = generateNonce();
+    const trustedDomains = [
```
aaschlote, line 101: we can move this logic outside of the return Promise (above to the line 98)

app/services/security/__test__/cspHeader.test.ts
```diff
@@ -1,27 +1,48 @@
...
+      reportUri: "https://reportCsp.com",
+    });
+    expect(reportCspHeaders).toContain("report-to");
+    expect(reportCspHeaders).toContain("https://reportCsp.com");
   });
 
   it("only adds localhost in development environment", () => {
-    expect(defaultHeader).not.toContain("localhost");
-    expect(cspHeader({ environment: "development" })).toContain("localhost");
+    expect(cspHeader(defaultArgs)).not.toContain("localhost");
```
aaschlote, line 39: I would add this line for a specific test, because this test is to check the development environment, not this condition

app/services/security/cspHeader.server.tsx
```diff
@@ -3,60 +3,48 @@
...
+  const directives: Record<string, string[]> = {
     "default-src": ["'self'"],
     "script-src": [
       "'self'",
-      "https://*.posthog.com",
       "https:",
-      "'unsafe-inline'",
+      `'nonce-${args.nonce}'`,
+      "'strict-dynamic'",
+      ...args.trustedDomains,
```
aaschlote, line 19: before we just have the posthog url, but with this change we have posthog and sentry. is this fine?

app/services/security/cspHeader.server.tsx
```diff
@@ -3,60 +3,48 @@
...
-    | "object-src"
-    | "base-uri"
-    | "frame-ancestors"]: string[];
-} & { "font-src"?: string[] };
-
-export const cspHeader = (args?: { nonce?: string; environment?: string }) => {
-  const directives: Directives = {
+export const cspHeader = (args: {
+  nonce: string;
+  trustedDomains: string[];
```
aaschlote, line 8: this parameter can't be a list of `URL`? So we only make sure this parameter doesn't receive any string value that is not an `URL`.

app/services/security/__test__/cspHeader.test.ts
```diff
@@ -1,27 +1,48 @@
 import { cspHeader } from "~/services/security/cspHeader.server";
 
 describe("cspHeader", () => {
```
aaschlote, line 3: can we improve this test and check if the following values has been add to the header?

- frame-src
- style-src
- img-src

app/services/security/__test__/cspHeader.test.ts
```diff
@@ -1,27 +1,48 @@
...
+      .find((directive) => directive.startsWith("script-src"));
+    expect(scriptDirective).toContain("nonce-r4nd0mN0nc3");
+    expect(scriptDirective).toContain("'strict-dynamic'");
+    expect(scriptDirective).toContain("https://trusted.com");
+    expect(scriptDirective).not.toContain("unsafe-inline");
   });
 
-  it("disables all framing", () => {
-    expect(defaultHeader).toContain("frame-ancestors 'none'");
+  it("adds report URI if passed", () => {
```
aaschlote, line 29:
```suggestion
  it("should add report-to and report-uri to the csp header given a reportUri parameter", () => {
```

aaschlote (commented):
I left some comments

app/services/env/env.server.ts
```diff
@@ -9,6 +9,7 @@ interface Config {
   COOKIE_SESSION_SECRET: string;
   CONTENT_FILE_PATH: string;
   ENVIRONMENT: string;
+  CSP_REPORT_URI?: string;
```
pgurusinga, line 12: why setting the report uri as optional?

chohner: we dont have one right now...

app/services/security/cspHeader.server.tsx
chohner: good hint. I did some digging and reduced the connect and script source to only specific URLs.

aaschlote (approved):
LGTM

PR #1355, opened 2024-10-28
refactor(flows): replace cmsSlug with flowType

chohner:
- replaces `cmsSlug` with `flowType`
- replaces hard-coded progress with filtering by `flowType`


frederike-ramin (approved):
Love it.

app/flows/flows.server.ts
```diff
@@ -19,8 +18,11 @@ export type FlowMigration = {
   buttonUrl?: string;
 };
 
+const flowTypes = ["vorabCheck", "formFlow"] as const;
```
aaschlote, line 21: shouldn't it be only `vorabcheck` without capital `C`? I believe it's only one word, right?

app/flows/flows.server.ts
```diff
@@ -19,8 +18,11 @@ export type FlowMigration = {
...
+export type FlowType = (typeof flowTypes)[number];
```
aaschlote, line 22: doing only:
```
export type FlowType = "vorabcheck" | "formFlow"
```

is it not enough? Or do we have other problems?

chohner: I just forgot to push

app/services/cms/apiFromFlowType.ts
```diff
@@ -0,0 +1,10 @@
+import type { FlowType } from "~/domains/flows.server";
+import type { FlowPageId } from "./schemas";
+
+const apiMapping = {
+  formFlow: "form-flow-pages",
+  vorabCheck: "vorab-check-pages",
+} satisfies Record<FlowType, FlowPageId>;
```
pgurusinga, line 7: Why `result-pages` is not part of the mapping?

chohner: There is no `FlowType: result-page`

PR #1361, opened 2024-10-29
ZOV: Fluggastechte pdf fill reason pages

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1208488510017838

### DIFF
--- a/app/services/pdf/fluggastrechte/fluggastrechtePdfFromUserdata.ts
+++ b/app/services/pdf/fluggastrechte/fluggastrechtePdfFromUserdata.ts
@@ -22,7 +22,7 @@ function buildDocument(
   setPdfMetadata(doc);
   createFirstPage(doc, documentStruct, userData);
   doc.addPage();
-  createReasonPage(doc, documentStruct);
+  createReasonPage(doc, documentStruct, userData);
   createFooter(doc, documentStruct, userData);
 }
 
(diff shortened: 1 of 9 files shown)


app/services/pdf/fluggastrechte/sections/getFullPlaintiffName.ts
```diff
@@ -0,0 +1,13 @@
+import _ from "lodash";
+import type { FluggastrechtContext } from "~/domains/fluggastrechte/formular/context";
+
+export const getFullPlaintiffName = (userData: FluggastrechtContext) => {
+  const { anrede, title, vorname, nachname } = userData;
+
+  const mappedTitle = title === "dr" ? "Dr." : title;
+  const capitalizedVorname = _.capitalize(vorname);
```
judithmh, line 8: Would it maybe make sense to do this via a schema already at the input? Then the data is already correctly displayed at the Zusammenfassung.

aaschlote: I don't think so, the `vorname` is a shared schema used in the whole `domains`. Add this schema would change their side. And capitalize this element is only need in the PDF. Not sure if we need in other places. And minor thing, the UX mentioned once change values after the user enter the data, it might be a weird behaviour.

PR #1364, opened 2024-10-31
docs(adr): should we use pnpm instead npm

m0dh4x:
its more like a RFC in my mind but don't know how to handle it yet. let me know in the comments or lets discuss on the dev exchange

### DIFF
--- a/doc/adr/0015-evaluation-npm-vs-pnpm.md
+++ b/doc/adr/0015-evaluation-npm-vs-pnpm.md
@@ -1,0 +1,52 @@
+# 15. Should we use pnpm instead pnpm
+
+Date: 2024-10-28
+
+## Status
+
+Proposed
+
+## Context
+
+First, in my opinion it is more a RFC than an ADR.
+
+Actually we use `npm` in our development process, but this ends up in several challanges. One of the biggest challanges is the installation speed and the disk space usage.
+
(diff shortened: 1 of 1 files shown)


m0dh4x:
> I like the idea to do a research to move to `pnpm` and it sounds promising.
> 
> Although, not sure if we need to create a documentation for every time we want to make a research, but we can discuss it in another opportunity.
> 
> LGTM.

Hey @aaschlote, thanks a lot. I totally agree with you, we don't need documentation for every research. But we should talk in general for every bigger research "it is worth to spent time on it every?". Maybe in the Dev Exchange with a short context and wat is the estimated time to finish the research, so we have a basis to take a decision?

doc/adr/0015-evaluation-npm-vs-pnpm.md
```diff
@@ -0,0 +1,52 @@
...
+Additional benefits:
+
+- you can install dependencies while offline mode `pnpm i --offline`
+- CI/CD integrate fetching from cache during installation process
+
+## Decision
+
+We will create a test pipeline to evaluate the performance and disk space usage of `pnpm` vs `npm`.
+
+My guess is that `pnpm` will perform better up to 50% on installaton speed and use less disk space up to 10%.
```
pgurusinga, line 48: I guess this sentence is the consequence of the decision, should not part of decision

pgurusinga:
@m0dh4x any update here? this has been 2 weeks already

PR #1391, opened 2024-11-06
chore(enable typescript-eslint type awareness)

Spencer6497:
This PR enables typescript-eslint linting with type information. It also fixes a number of resulting lint errors, adding FIXME's to the rest, and overall reorganizes the overridden lint rules in a more sensible way.

- **enable typed linting**
- **disable all erring rules**
- **remove unused "camelcase" eslint-ignores**
- **remove unused no-unsafe-argument ignores**
- **remove duplicate/redundant rules, reorganize ignores for clarity**
- **remove unnecessary comments from eslint rule overrides**


Spencer6497:
Making this a draft PR before it gets way too bloated. Will need to fix (most) of the FIXME's and reevaluate/debate the style rules. Unfortunately tsc errors are not caught by this mode, meaning we can't get rid of the `typecheck` step :(

app/components/navigation/FlowNavigation.tsx
```diff
@@ -1,4 +1,4 @@
-import { NavigationList, type FlowNavigationProps } from "./NavigationList";
+import { NavigationList, type FlowNavigationProps } from "./NavItem";
```
chohner, line 1: wait, why is navigation list imported from navitem?

app/services/gerichtsfinder/checkURLs.ts
```diff
@@ -55,30 +55,37 @@ function checkUrl(urlString: string) {
...
-                  resolve([urlString, response.responseUrl]);
-                },
-              )
-              .on("error", () => {
-                if (urlString)
-                  console.error(`No valid redirect found for ${urlString}`);
-                resolve([urlString, ""]);
-              });
-          });
+        attemptHttp(urlString, resolve);
```
chohner, line 58: this change I find less readable. we could maybe discuss finding a better solution (or just add an exception since this is a script rather than runtime code)

app/components/navigation/NavItem.tsx
```diff
@@ -113,3 +112,30 @@ export function NavItem({
...
+export const NavigationList = ({
+  navItems,
+  ...props
+}: FlowNavigationProps & { isChild?: boolean }) => (
+  <ul className="pl-0">
+    {navItems.map((navItem) => (
+      <NavItem {...navItem} key={navItem.destination} {...props} />
+    ))}
+  </ul>
+);
```
chohner, line 125: I understand this is to clean up the cyclic import but:
- cyclic imports are valid (and they might be in this case)
- `NavigationList` is the outer element and should be its own file
- exporting multiple components from one file is considered bad practice afaik


tldr: because the solution isn't so clear imo, maybe we can move this refactor out of this PR?

eslint.config.js
```diff
@@ -86,40 +90,86 @@ export default tseslint.config(
...
+      "sonarjs/sonar-prefer-optional-chain": "off",
+
+      // to be enabled later
+      "sonarjs/no-unstable-nested-components": "warn", // TODO: enable later
+
+      // fix/reevaluate
+      "sonarjs/new-cap": "off", // FIXME: turn on
+      "sonarjs/deprecation": "off", // FIXME: turn on
+      "sonarjs/no-empty-test-file": "off", // FIXME: turn on
+      "sonarjs/function-return-type": "off", // FIXME: turn on
```
chohner, line 131: maybe we can replace all the TODOs/ FIXMEs we either a single one OR even better, make an asana ticket to track? no need to pollute our logs with this imo

Spencer6497: https://app.asana.com/0/1203259475859667/1208709688709647/f

chohner (approved):

app/components/navigation/FlowNavigation.tsx
```diff
@@ -1,4 +1,5 @@
-import { NavigationList, type FlowNavigationProps } from "./NavigationList";
+import type { FlowNavigationProps } from "~/components/navigation/NavigationList";
+import { NavigationList } from "~/components/navigation/NavigationList";
```
chohner, line 2: I guess these can stay as relative import?

PR #1392, opened 2024-11-06
WIP: Use PNPM

pgurusinga:


Dockerfile
```diff
@@ -4,8 +4,8 @@ ARG APP_IMAGE=app
 
 FROM node:20-alpine AS app-base
 WORKDIR /a2j
-COPY package.json package-lock.json ./
-RUN npm ci --omit=dev --omit=optional
+COPY package.json pnpm-lock.yaml ./
+RUN pnpm install --shamefully-hoist --frozen-lockfile --prod
```
chohner, line 8: `shamefully-hoist`? 😅

.github/workflows/pnpm-checks.yml
```diff
@@ -1,4 +1,4 @@
-name: NPMChecks
+name: PNPMChecks
```
chohner, line 1: we should probably rename this whole flow to not rely on the tool name

Dockerfile
pgurusinga: When we ran `pnpm install`, and then `pnpm typecheck`, we got about 400 type errors. I think the problem is how `pnpm` organizes dependencies. The `node_modules` folder that `pnpm` creates is more compact, but if we install dependencies with `--shamefully-hoist`, all the errors go away. The `--shamefully-hoist` option makes `pnpm` install dependencies like `npm` does, placing everything at the top level.

Could you try this on your machine and let me know if you see the same results?

 https://github.com/pnpm/pnpm/issues/2317

pgurusinga:
I'm closing this pull request for now. We can revisit it when we have more time.

PR #1393, opened 2024-11-07
refactor: move domain specific pdf logic for berH and PKH into their …

ekl176:
…domains


rbrtrfl (commented):
1. I think` /pdf` should not directly reside in the domain folders like but inside of a subfolder `/services` e.g. `/app/domains/beratunsghilfe/services/pdf`.
2. To my understanding, `/app/services/pdf/pdfs.ts` is also domain logic and should be moved to the domains folder.
3. Maybe all content of `/data/pdf` could also move to `/app/domains`? At least the `*.generated.ts` files.

4. What is it about those removed eslint comments in this PR?

PR #1398, opened 2024-11-11
a11y(Skip to Content link)

Spencer6497:
This PR creates a Skip to Content link button, allowing keyboard/screenreader users to skip the navigation of our site and jump directly to the content.

- **turn off sonarlint no-dead-store rule as duplicate, add accessibility Strapi translations, add SkipToContentLink component, update tests**
- **add useEffect hook to correctly focus main element, relocate isFlowIdInPathname utility function, add reusable useIsUserInFlow() hook**


Spencer6497:
Open question (see commented out code): do we want to skip to the specific form flow page when the user is inside of a form flow?

app/root.tsx
```diff
@@ -205,23 +232,24 @@ function App() {
       </head>
       <body className="flex flex-col min-h-screen">
         <CookieConsentContext.Provider value={hasTrackingConsent}>
-          <CookieBanner content={getCookieBannerProps(cookieBannerContent)} />
-          <Header {...header} translations={pageHeaderTranslations} />
-          <Breadcrumbs breadcrumbs={breadcrumbs} />
           <TranslationContext.Provider value={translationMemo}>
-            <main className="flex-grow">
+            <SkipToContentLink />
```
chohner, line 236: I would (strongly) prefer fetching the skipContent translations in the loader and then just passing them into `<SkipToContentLink/>` and leaving the `TranslationContext` to be an escape hatch for deeply nested components

app/root.tsx
```diff
@@ -173,17 +178,39 @@ function App() {
...
   const matches = useMatches();
   const { breadcrumbs, title, ogTitle, description } = metaFromMatches(matches);
   const nonce = useNonce();
+  const userInFlow = useIsUserInFlow();
+
+  // Needed to correctly trap focus on the main content after skipping to content
+  useEffect(() => {
+    const focusMainContent = () => {
+      document
+        .getElementById(/* userInFlow ? "form-flow-page-content" : */ "main")
```
chohner, line 192: I think we could use `document.querySelector('main')` without adding the `id` on `<main>`

app/root.tsx
```diff
@@ -205,23 +230,24 @@ function App() {
       </head>
       <body className="flex flex-col min-h-screen">
         <CookieConsentContext.Provider value={hasTrackingConsent}>
-          <CookieBanner content={getCookieBannerProps(cookieBannerContent)} />
-          <Header {...header} translations={pageHeaderTranslations} />
-          <Breadcrumbs breadcrumbs={breadcrumbs} />
           <TranslationContext.Provider value={translationMemo}>
-            <main className="flex-grow">
+            <SkipToContentLink translations={accessibilityTranslations} />
```
chohner, line 234: I think having a "dumb" `SkipToContentLink` component with a strict interface is preferable to passing in translations.

app/root.tsx
```diff
@@ -205,23 +230,24 @@ function App() {
...
             </main>
+            <footer>
+              <Footer
+                {...footer}
+                deletionLabel={deletionLabel}
+                showDeletionBanner={hasAnyUserData}
+              />
+            </footer>
+            <ScrollRestoration nonce={nonce} />
+            <Scripts nonce={nonce} />
```
chohner, line 249: why did all of this move?

app/root.tsx
```diff
@@ -173,16 +177,37 @@ function App() {
     feedbackTranslations,
     pageHeaderTranslations,
     videoTranslations,
+    accessibilityTranslations,
   } = useLoaderData<RootLoader>();
   const matches = useMatches();
   const { breadcrumbs, title, ogTitle, description } = metaFromMatches(matches);
   const nonce = useNonce();
 
+  // Needed to correctly trap focus on the main content after skipping to content
```
chohner, line 186: this seems like an surprising amount of work for jumping to an anchor somewhere in the DOM. could you give some context or external reference?

app/components/navigation/SkipToContentLink.tsx
```diff
@@ -0,0 +1,34 @@
...
+  translations: Translations;
+}>) {
+  const [target, setTarget] = useState("main");
+
+  // useLayoutEffect is necessary here, as we can't use `document` in an SSR environemnt
+  useLayoutEffect(() => {
+    if (document.getElementById("form-flow-page-content")) {
+      setTarget("form-flow-page-content");
+    }
+  }, []);
```
chohner, line 22: wild idea - would it be simpler if we pass the destination into the component?

app/styles.css
```diff
@@ -47,6 +47,15 @@ svg {
...
 }
 
+#skip-to-content-link {
+  position: absolute;
+  transform: translateY(-100%);
+}
+#skip-to-content-link:focus {
+  position: relative;
+  transform: translateY(0);
+}
```
chohner, line 57: could this be styled inline?

chohner (approved):

app/root.tsx
```diff
@@ -194,11 +219,15 @@ function App() {
       </head>
       <body className="flex flex-col min-h-screen">
         <CookieConsentContext.Provider value={hasTrackingConsent}>
+          <SkipToContentLink
+            label={`↓ ${getTranslationByKey(SKIP_TO_CONTENT_TRANSLATION_KEY, accessibilityTranslations)}`}
```
pgurusinga, line 223: what is this arrow-down character? is that necessary?

Spencer6497: it's [in the design](https://www.figma.com/design/LbkGUR1Kqf6ISqODlSxnMg/%F0%9F%A4%9D-A2J?node-id=5884-7151&t=DA3F7oHaVhrQnGp4-4), but I don't believe it's necessary from an accessibility standpoint

app/styles.css
pgurusinga: I guess what Chris thought inline with tailwind instead, but correct me if I'm wrong

```jsx
<a 
  id="skip-to-content-link" 
  class="absolute transform -translate-y-full focus:relative focus:translate-y-0"
  href="#form-flow-page-content"
>
  Skip to content
</a>
```

Spencer6497: ah completely forgot about tailwind, thanks for this!

PR #1411, opened 2024-11-12
chore(enable sonarlint deprecation rule + fixes)

Spencer6497:
This enables the no-alphabetical-sort sonarjs eslint rule, and turns off the sonarjs/deprecation rule (for now, until remix upgrades to react router v7 and we turn Single Fetch on - see issue listed below


Spencer6497:
it seems that there are issues with appending the accepted/refused cookie's header... from what I can tell, this seems to be some issue with remix's implementation of the `data()` helper.

app/services/flow/server/buildFlowController.ts
```diff
@@ -140,7 +135,7 @@ const metaFromStepId = (machine: FlowStateMachine, currentStepId: string) => {
 
 function getInitial(machine: FlowStateMachine) {
   // The initial state might be nested and needs to be resolved
-  const initialSnapshot = getInitialSnapshot(machine);
+  const [initialSnapshot] = initialTransition(machine);
```
pgurusinga, line 138: What problem did we face that makes us need to change this?

Spencer6497: `getInitialSnapshot` is actually deprecated

pgurusinga (approved):

PR #1412, opened 2024-11-12
feat: graceful shutdown, re-use existing redis client

chohner:
Context
- move most of the express server logic into typescript (`expressApp.ts`)
- express app is still instatiated from `server.js` (via `build.entry.module` magic 🥲)
- re-uses single redis client (rate limiting & user data)
- enables graceful shutdown of redis, sentry & posthog

Co-Authored-By: ekl176 <169838059+ekl176@users.noreply.github.com>


app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  app.all("*", remixHandler);
+
+  return {
+    app,
+    cleanup: async () => {
+      const shutdownTimoutMs = 1000;
+      return Promise.all([
+        redisClient.quit().then(() => {
+          console.log("✅ reddis client");
```
frederike-ramin, line 98: Can we add "Quit" to the messages? (and fix the reddis typo) I'm imagining searching for this in Grafana and that would just make it easier to find.

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+import type { ServerBuild } from "@remix-run/node";
+import * as Sentry from "@sentry/remix";
+import compression from "compression";
+import express from "express";
+import { rateLimit } from "express-rate-limit";
+import Redis from "ioredis";
+import { RedisStore, type RedisReply } from "rate-limit-redis";
+import type { ViteDevServer } from "vite";
+import { setRedisClient } from "./services/session.server/redis";
+import { getPosthogClient } from "./services/analytics/posthogClient.server";
```
frederike-ramin, line 13: Can you handle the lint import order warning?

app/services/session.server/redis.ts
```diff
@@ -1,31 +1,13 @@
...
+
+let redisClient: Redis;
 
-const redisUrl = () =>
-  `rediss://default:${config().REDIS_PASSWORD}@${config().REDIS_ENDPOINT}`;
+export function setRedisClient(newRedisClient: Redis) {
+  if (!redisClient) redisClient = newRedisClient;
+}
 
 async function getRedisInstance() {
```
frederike-ramin, line 9: This is then the function to check on the /health route, right?

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+
+const redisUrl = () =>
+  `rediss://default:${process.env.REDIS_PASSWORD?.trim()}@${process.env.REDIS_ENDPOINT ?? "localhost:6380"}`;
```
frederike-ramin, line 16: Not 100% sure with this mention of localhost here? It's just a convenience so we don't have to set it in the .env of our development envirnonments, right? I'm wondering if just adding it to the `.env.example` would be a better solution? (I know you didn't really introduce it, just challenging here)

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+  } else {
+    // Vite fingerprints its assets so we can cache forever.
+    app.use(
+      "/assets",
+      express.static("build/client/assets", { immutable: true, maxAge: "1y" }),
+    );
+  }
+
+  // Everything else (like favicon.ico) is cached for an hour. You may want to be
+  // more aggressive with this caching.
```
frederike-ramin, line 53: For my understanding, what is "everything else"? Everything that is not "asset requests"? Is the request to `favicon.ico` not an asset request? Is there a definition I don't know about?

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+  const mountPathWithoutStorybook = [
+    /^\/build\/client\/storybook($|\/)/,
+    "/build/client",
+  ];
+
+  if (isStagingOrPreviewEnvironment) {
+    app.use(staticFileServer);
+  } else {
+    app.use(mountPathWithoutStorybook, staticFileServer);
+  }
```
frederike-ramin, line 64: @chohner pretty sure this is causing the "no security.txt on prod" error? We use a different mount path for production, so quite sure this is the difference?

frederike-ramin (approved):
Thank you for handling this, it makes a lot of sense to me, I just have a few comments, most not even really related to this PR, I'm sorry, handle as you wish^^

app/expressApp.ts
Spencer6497: furthermore, why is this a function and not defined inline?

app/services/session.server/redis.ts
Spencer6497, line 1: is there a reason we're using a global `let` instead of a discrete `RedisClient` class, with getters and setters?

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+  `rediss://default:${process.env.REDIS_PASSWORD?.trim()}@${process.env.REDIS_ENDPOINT ?? "localhost:6380"}`;
+
+export const expressApp = (
+  build: ServerBuild,
+  viteDevServer: ViteDevServer,
+) => {
+  const isStagingOrPreviewEnvironment =
+    process.env.ENVIRONMENT !== "production";
+
+  const redisClient = new Redis(redisUrl(), {
```
aaschlote, line 25: can we move this to a specific function and file, like, `createRedisClient`? You also can move the function `redisUrl`? You can add an unit test for that to see if the configs to create the redis are expected

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+import { getPosthogClient } from "./services/analytics/posthogClient.server";
+
+const redisUrl = () =>
+  `rediss://default:${process.env.REDIS_PASSWORD?.trim()}@${process.env.REDIS_ENDPOINT ?? "localhost:6380"}`;
+
+export const expressApp = (
+  build: ServerBuild,
+  viteDevServer: ViteDevServer,
+) => {
+  const isStagingOrPreviewEnvironment =
```
aaschlote, line 22: can be moved outside of the function scope right?

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  // handle SSR requests
+  const remixHandler = createRequestHandler({ build });
+
+  app.all("*", remixHandler);
+
+  return {
+    app,
+    cleanup: async () => {
+      const shutdownTimoutMs = 1000;
```
aaschlote, line 95: typo?  `shutdownTimeoutMs`

app/expressApp.ts
```diff
@@ -0,0 +1,111 @@
...
+    "/build/client",
+  ];
+
+  if (isStagingOrPreviewEnvironment) {
+    app.use(staticFileServer);
+  } else {
+    app.use(mountPathWithoutStorybook, staticFileServer);
+  }
+
+  // For the rate limiting to work, we have to set how many load balancers aka proxy hubs
```
aaschlote, line 66: this line until the line 86 sounds a scope for set the rate limit for the pdf pages, can we move to a function and keep `expressApp` with less code and easy to read?

server.js
```diff
@@ -1,89 +1,30 @@
-import { createRequestHandler } from "@remix-run/express";
```
aaschlote, line 1: I raised this question already in Slack, but asking again, why the `server.js` is JS and not Typescript?

aaschlote (approved):
LGTM.

Changes sounds good to me, although we can decoupling a few parts of the code (mentioned in few comments) so we can easily add some unit tests.

app/expressApp.ts
aaschlote: you can pass to this function the app as parameter.

package.json
```diff
@@ -103,7 +103,9 @@
     "@storybook/react-vite": "^8.4.2",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.0.1",
+    "@types/compression": "^1.7.5",
     "@types/eslint__js": "^8.42.3",
+    "@types/express": "^5.0.0",
```
pgurusinga, line 108: should this be the same version as express `^4.21.1`? or does it not matter?

server.js
```diff
@@ -1,89 +1,30 @@
...
+  console.log("Gracefully shuting down...");
+  server.close(async () => {
+    await cleanup();
+    console.log("Exiting process, goodbye");
+    process.exit(0);
+  });
+}
+
+process.on("SIGTERM", cleanupAndShutdown);
+process.on("SIGINT", cleanupAndShutdown);
```
pgurusinga, line 30: Now that we have graceful shutdowns, what’s the main goal? Are we trying to monitor it or just make sure our app closes down the right way?

aaschlote: From @ekl176 in Slack:

``
we can properly close all opened connections when gracefully shutting down the server. We have the suspicion that too many redis connections that might be kept open during the rolling deployment lead to issues.
Using the same redis connection for rate limiting and userdata enables us to safely check the connection in our healthcheck, preventing a pod to be deemed ready even when the redis connection isn't established yet.
``

app/services/session.server/redis.ts
chohner: There we still check that getRedisStatus() returns "ready"

app/services/redis/__test__/redisClient.test.ts
```diff
@@ -0,0 +1,29 @@
+import { quitRedis } from "../redisClient";
+
+describe("quitRedis", () => {
+  it("should call .quit() and resolve", async () => {
+    const redisMock = {
+      quit: vi.fn(() => Promise.resolve("OK" as const)),
```
aaschlote, line 6:
```suggestion
      quit: vi.fn().mockResolvedValue("OK"),
```

app/expressApp.ts
chohner: So: `build/client/assets` is where vite is building its js&css output. "everything else" therefore are all static files that are not build artifacts (fonts, images, favicons etc)

PR #1422, opened 2024-11-13
chore(enable async/Promise eslint rules)

Spencer6497:
This PR enables a number of sonarjs/typescript-eslint rules that concern async/await's and Promises.

### DIFF
--- a/app/components/inputs/autoSuggestInput/useDataListOptions.tsx
+++ b/app/components/inputs/autoSuggestInput/useDataListOptions.tsx
@@ -21,6 +21,7 @@ const useDataListOptions = (dataListType: DataListType) => {
       }
     };
 
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
     fetchData();
   }, [dataListType]);
 
--- a/app/entry.server.tsx
+++ b/app/entry.server.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/prefer-promise-reject-errors */
 import { PassThrough } from "stream";
 import type {
   ActionFunctionArgs,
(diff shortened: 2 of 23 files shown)


tests/e2e/flowPages/beratungshilfe/formular/finanzielleAngabenWohnung.ts
```diff
@@ -14,7 +14,7 @@ export async function startFinanzielleAngabenWohnung(
     () => beratungshilfeFormular.fillInput("apartmentCostOwnShare", "40"),
   ];
 
-  for await (const callback of stepCallbacks) {
+  for (const callback of await Promise.all(stepCallbacks)) {
```
chohner, line 17: this looks weird, shouldn't this just be `await Promise.all(stepCallbacks)`?

app/components/inputs/autoSuggestInput/useDataListOptions.tsx
```diff
@@ -21,6 +21,7 @@ const useDataListOptions = (dataListType: DataListType) => {
       }
     };
 
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
chohner, line 24: is this one left floating on purpose?

app/entry.server.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/prefer-promise-reject-errors */
```
chohner, line 1: is this a follow up? or simply out of scope?

app/components/inputs/autoSuggestInput/useDataListOptions.tsx
Spencer6497: yeah; this is the official recommendation of how to handle asynchronous function calls within a `useEffect`. we could put a `void` before it but I think that's a bit more unclear tbh

app/entry.server.tsx
Spencer6497: there are two spots in this file where this applies but where we're intentionally ignoring it; in another PR i'll put the ignores next to the relevant code for clarity

tests/e2e/flowPages/beratungshilfe/formular/finanzielleAngabenWohnung.ts
Spencer6497: i'm not sure what you mean? we still have to iterate through the promises

PR #1433, opened 2024-11-15
ZOV: add non-editable part of overview page

pgurusinga:


app/domains/fluggastrechte/formular/__test__/getAnnullierungInfoRendering.test.ts
```diff
@@ -0,0 +1,62 @@
...
+{{/hasBetween7And13DaysAnkuendigung}}
+{{^hasBetween7And13DaysAnkuendigung}}
+Start: {{#hasErsatzflugStartenEinStunde}}>1h earlier{{/hasErsatzflugStartenEinStunde}}{{^hasErsatzflugStartenEinStunde}}<1h earlier{{/hasErsatzflugStartenEinStunde}}
+Arrival: {{#hasErsatzflugLandenZweiStunden}}>2h later{{/hasErsatzflugLandenZweiStunden}}{{^hasErsatzflugLandenZweiStunden}}<2h later{{/hasErsatzflugLandenZweiStunden}}
+{{/hasBetween7And13DaysAnkuendigung}}
+`;
+
+const normalizeWhitespace = (str: string) => str.trim().replace(/\s+/g, " ");
+
+describe("getAnnullierungInfo Render", () => {
```
aaschlote, line 16: would you prefer to move this test to the file `app/util/__test__/fillTemplate.test.ts`?

aaschlote (approved):
besides my comment, LGTM!

image

app/domains/fluggastrechte/formular/__test__/getAnnullierungInfoRendering.test.ts
pgurusinga: I prefer to keep the implementation close, as I would miss writing the test if it is too distant, like in `fillTemplate.test.ts.`, would it be fine for you?

aaschlote: the main problem here is the `fillTemplate` does in the end what you are testing, given a content and object, replace the content by the object values.

Here you don't call anything from the formular stringReplacement, only the `mustache` library. If you delete any function stringReplacement this test still will be green.

For me, the tests should be:

- `app/util/__test__/fillTemplate.test.ts` -> Test given a content and object, replace given a content and object, replace the content by the object values ✅
- `app/domains/fluggastrechte/formular/__test__/stringReplacements.test.ts` -> Test given the formular context, return an object with the expected value ✅

I understand you created this tests to check if the library should replace the values for the business of the Fluggastrechte context, but thinking about the logic as our code has been organised in this repository, this tests doesn't bring any value at all.

pgurusinga: seems fillTemplate is already doing its job, I will delete this test

PR #1436, opened 2024-11-18
refactor(stepId): leading slash in app

chohner:
The CMS references pages with a leading slash. This PR aligns the app to do the same.

Changes:
- flowIds.ts: adjust `flowIdFromPathname` and `parsePathname` to leave leading slash
- stepIdConverter: remove leading slash in `stepIdToPath` and include it in `stateValueToStepIds`
- cms/index & pruner & buildFlowcontroller: dont pre-pend `/`
- result.server: fetches meta of flowId for the breadcrumb, instead of the result page
- progress: keys are paths, not ids. code got a bit ugly :/
- updates all the tests 

Things to check:
- laststep: prepend flow id?
- ~revert changes in progress?~

Todo:
- rename menu translations in cms


rbrtrfl (approved):
Just left two small comments which are more questions of understanding from me.
What's the process of rolling this change out?

app/routes/shared/result.server.ts
```diff
@@ -34,10 +35,8 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
     return redirect(flowController.getInitial());
 
   const [resultPageContent, parentMeta, defaultStrings] = await Promise.all([
-    fetchFlowPage("result-pages", flowId, stepId.replace("ergebnis/", "")),
-    fetchMeta({
-      filterValue: pathname.substring(0, pathname.lastIndexOf("/")),
-    }),
+    fetchFlowPage("result-pages", flowId, cmsStepId),
+    fetchMeta({ filterValue: flowId }),
```
frederike-ramin, line 39: To be sure: this is just a cleanup, right? The old code shouldn't have been broken with the original change in this PR?

app/services/flow/server/lastStep.ts
```diff
@@ -24,7 +24,7 @@ export async function loader({ request }: LoaderFunctionArgs) {
   const lastStep = await lastStepFromRequest(cookieHeader);
   const destination =
     lastStep && flowId in lastStep
-      ? lastStep[flowId]
+      ? flowId + lastStep[flowId]
```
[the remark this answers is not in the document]
frederike-ramin: Or in other words - how did this ever work before?

app/services/flow/server/progress.ts
```diff
@@ -2,6 +2,10 @@ import { createMachine } from "xstate";
 import { flows } from "~/domains/flows.server";
 import { type FlowStateMachine } from "./buildFlowController";
 
+function xstateIdToStepId(xstateId: string, machineId: string) {
+  return xstateId.replace(machineId, "").replaceAll(".", "/");
+}
```
frederike-ramin, line 7: This seems like the opposite to stepIdToPath in /app/services/flow/stepIdConverter.ts - should we also move it there?

app/services/flow/server/progress.ts
```diff
@@ -40,7 +44,19 @@ export function progressLookupForMachine(machine: FlowStateMachine) {
         }
       });
   }
-  return { progressLookup, total: Math.max(...Object.values(progressLookup)) };
+
+  // Convert all keys from xstateId (machineId.state.substate) to stepId
```
frederike-ramin, line 48:
```suggestion
  // Convert all keys from xstateId (machineId.state.substate) to stepId (machineId/state/substate)
```

app/services/flow/server/progress.ts
```diff
@@ -40,7 +44,19 @@ export function progressLookupForMachine(machine: FlowStateMachine) {
...
+    Object.entries(progressLookup).map(([key, val]) => [
+      xstateIdToStepId(key, machine.id),
+      val,
+    ]),
+  );
+
+  return {
+    progressLookup: progressLookupStepIds,
+    total: Math.max(...Object.values(progressLookup)),
+  };
```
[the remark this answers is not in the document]
frederike-ramin: I believe this comment pretty much says it all:

// Convert all keys from xstateId (machineId.state.substate) to stepId

We're handling with xstate IDs (machineId.state.substate) here because we need them for the lookup algorithm above. However, everywhere alse in the app that is not as close to xstate as this file/function here, we need the stepID in the format machineId/state/substate so we need to convert them. I agree, the code is not super.readable - and Chris agrees too, but the comment sums it up pretty nicely?

app/services/flow/stepIdConverter.ts
```diff
@@ -1,25 +1,29 @@
...
 // stateValue: xstate type StateValue = string | StateValueMap, ie: {persoenliche-daten: anzahl}
 // - non-nested states can be simple string (ie "start")
 
 export const stepIdToPath = (stepId: string) =>
-  stepId.replace(/\//g, ".").replace("ergebnis.", "ergebnis/").split(".");
+  stepId
+    .replaceAll("/", ".")
+    .replace("ergebnis.", "ergebnis/")
+    .split(".")
+    .slice(1);
```
frederike-ramin, line 14: What's this last slice for? I don't think I get it.

frederike-ramin (approved):
I like this change. Makes you wonder how we ever lived without it. :D

app/services/flow/server/lastStep.ts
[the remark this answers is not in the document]
pgurusinga: image

How can a machine like @rbrtrfl's comment? :D

pgurusinga (approved):

app/routes/shared/result.server.ts
chohner: yes, I'll pull these changes out of this branch

app/services/flow/stepIdConverter.ts
chohner: The leading slash is converted to an empty string at the beginning of the array. maybe slicing the first char of the stepId is more readable

PR #1441, opened 2024-11-18
Move to pdfkit

rbrtrfl:
- rename `/sections` folder with `/pdfForm` in `/domains/beratungshilfe/services/pdf`
- move related files into `/domains/prozesskostenhilfe/services/pdf/pdfForm`
- move `index.ts` from `/services/pdf/attachment to /domains/shared/pdf/attachment`
- move `/domains/shared/pdf` to `/domains/shared/services/pdf`
- move `/app/services/pdf/util` to `/app/domains/prozesskostenhilfe/services/pdf` along with corresponding tests
- move `/app/domains/share/services/pdf/attachment` to `/app/services/pdf/attachment`
- move `/app/domains/share/services/pdf/pdfStyles.ts` to `/app/services/pdf/`
- remove circular dependencies


app/domains/prozesskostenhilfe/services/pdf/__test__/util.test.ts
```diff
@@ -45,4 +48,26 @@ describe("utility functions", () => {
       ).toBe("1.120,00");
     });
   });
+
+  describe("removeDecimalsFromCurrencyString", () => {
```
aaschlote, line 52: this test is in the right folder?

app/domains/prozesskostenhilfe/services/pdf/util.ts
```diff
@@ -24,3 +26,10 @@ export const getTotalMonthlyFinancialEntries = (
       minimumFractionDigits: 2,
       maximumFractionDigits: 2,
     });
+
+export function removeDecimalsFromCurrencyString(
```
aaschlote, line 30: maybe it should be in the `app/util`, right?

app/domains/prozesskostenhilfe/services/pdf/__test__/util.test.ts
rbrtrfl: At the moment: Yes.

app/domains/prozesskostenhilfe/services/pdf/util.ts
rbrtrfl: The function is only used in the scope of filling pdf forms. So I would say it's correctly located where it is right now.
wdyt?

aaschlote: it's only used in the scope of filling pdf forms, but the function itself doesn't have anything related the pdf in the source code, or am I missing something?

PR #1446, opened 2024-11-19
feat(add integration tests against CMS)

Spencer6497:
This PR uses vitest to test the app's context against the existing data in live, deployed Strapi.

It also introduces [vitest Workspaces](https://vitest.dev/guide/workspace.html) to separate our unit and integration tests, and tweaks the CI script to run integration tests immediately following content file fetch.

- **add verify cms content test**
- **enable testing of strapi for xstate entries**
- **add custom jest matcher, fix recursive logic, structure tests more clearly**
- **add test for field names in xstate**
- **remove unused xstate pages from fluggastrechte and geldeinklagen flows, tweak error/warning messages**
- **add recursive zod key function, test strapi form fields against it**
- **move integration test to appropriate folder and add to vitest setup**
- **enable vitest workspaces and tweak scripts/CI pipeline**


tests/integration/util.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @example zodKeys(
+ *            z.object({
+ *              a: z.string(),
+ *              b: z.number(),
+ *              c: z.object({ d: z.string() }),
+ *              e: z.array({ f: z.string() })
+ *          })) => ['a', 'b', 'c.d', 'e#f']
+ */
+export function zodKeys<T extends z.ZodTypeAny>(schema: T): string[] {
+  // make sure schema is not null or undefined
```
chohner, line 50: I think most if not all of these comments could be removed for a general comment that describe the need for the big if tree (something like `Zod schemas have different ways of encoding their keys`)

tests/integration/util.ts
```diff
@@ -0,0 +1,115 @@
...
+  const allPossibleStates: string[] = [];
+
+  function _getAllPossibleStates(
+    states: Flow["config"]["states"],
+    path: string,
+  ) {
+    for (const state in states) {
+      if (ignoreList.includes(state)) continue;
+      if (!states[state].initial) {
+        allPossibleStates.push(`${path}/${state.replace(/^ergebnis\//, "")}`);
```
chohner, line 95: `${path}/${state.replace(/^ergebnis\//, "")}` could be moved outside of the if

tests/integration/util.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  }
+
+  _getAllPossibleStates(flow.config.states, "");
+
+  return allPossibleStates;
+}
+
+export function invertStrapiFormFields(formFields: FormFieldsMap) {
```
chohner, line 111: this is technically a general function, right?

tests/integration/verifyCmsContent.test.ts
```diff
@@ -0,0 +1,157 @@
...
+} from "tests/integration/util";
+import { getContext } from "~/domains/contexts";
+import type { FlowId } from "~/domains/flowIds";
+import { flowIds } from "~/domains/flowIds";
+import { flows } from "~/domains/flows.server";
+import { fetchAllFormFields } from "~/services/cms/fetchAllFormFields";
+import { getStrapiEntry } from "~/services/cms/getStrapiEntry";
+import type { StrapiSchemas } from "~/services/cms/schemas";
+
+const allStrapiData: AllStrapiData = {} as AllStrapiData;
```
chohner, line 17: the `as AllStrapiData` shouldnt be necessary

tests/integration/verifyCmsContent.test.ts
```diff
@@ -0,0 +1,157 @@
...
+          ],
+        }),
+        fetchAllFormFields(flowId),
+      ]);
+
+    allStrapiData[flowId] = {
+      "vorab-check-pages":
+        vorabCheckPages[0] === null
+          ? []
+          : (vorabCheckPages as StrapiSchemas["vorab-check-pages"]),
```
chohner, line 76: i'm surprised the type has be be enforced here, shouldn't `getStrapiEntry` return the correct type?

app/domains/beratungshilfe/formular/finanzielleAngaben/__test__/doneFunctionsEigentum.test.ts
```diff
@@ -538,7 +538,6 @@ describe("kraftfahrzeugeDone", () => {
               kilometerstand: 20000,
               anschaffungsjahr: "1990",
               baujahr: "1990",
-              bemerkung: "bemerkung",
```
chohner, line 541: the actual changes could probably be extraced and already merged, right?

chohner (approved):
lgtm, few questions

tests/integration/util.ts
Spencer6497: yeah. was torn about where to locate it, but decided on `tests/integration` specifically as we're only using it _currently_ in integration tests. would be happy to relocate it to somewhere like `fetchAllFormFields.ts`

tests/integration/verifyCmsContent.test.ts
Spencer6497: since we're defining `AllStrapiData` as `Record<FlowId, StrapiPages>`, and `FlowId` is a string literal type, ts complains about the missing keys

tests/integration/verifyCmsContent.test.ts
Spencer6497: `getStrapiEntry` returns `StrapiSchemas[T] | [null]`, and the `[null]` is what's causing issues (even with a proper null check)

.github/workflows/ci-pipeline.yml
```diff
@@ -39,8 +39,19 @@ jobs:
...
+  run-integration-tests:
     needs: [get-content-file]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/cached-checkout-install
+      - uses: actions/download-artifact@v4
+        with:
+          name: content-file
+      - run: npm run test:integration
```
pgurusinga, line 51: can this test run as a matrix in `npm-checks.yml`?

tests/integration/verifyCmsContent.test.ts
```diff
@@ -0,0 +1,157 @@
...
+});
+
+describe.each(flowIds)("Availability of %s content", (flowId: FlowId) => {
+  describe("pages", () => {
+    test(`all states that are referenced in the ${flowId} xState config are available in Strapi`, () => {
+      const pages = compileAllStrapiPages(flowId, allStrapiData);
+
+      const flow = flows[flowId];
+      const allPossibleStates = getAllPossibleStates(flow);
+      allPossibleStates.forEach((stateName) => {
```
pgurusinga, line 97: not a blocker, fyi, i saw in a benchmark that `for...of` or classic for loops faster than `forEach` https://muka-gergely.medium.com/how-i-optimized-javascript-loops-in-my-app-261e03a6be89

.github/workflows/ci-pipeline.yml
Spencer6497: since it relies on the `get-content-file` step and its output, it has to be separate unfortunately

pgurusinga: makes sense to use `content.json` to avoid network requests to Strapi staging

tests/integration/verifyCmsContent.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      };
+    }
+  },
+});
+
+beforeAll(async () => {
+  for (const flowId of flowIds) {
+    const [vorabCheckPages, resultPages, formFlowPages, formFields] =
+      await Promise.all([
+        getStrapiEntry({
```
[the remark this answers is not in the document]
Spencer6497: yeah, that's correct. i'd be happy to change the test to call `getStrapiEntryFromFile` instead for consistency, though I kind of like the idea of leaving it up to the developer which API to hit (local or deployed)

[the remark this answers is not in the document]
pgurusinga: Is there a chance that the test results don't match between the file and Strapi? (_content is not uptodate_)
For example, my integration test works on my local setup but fails during the build. If you think this won't happen, I'm okay with keeping it as is.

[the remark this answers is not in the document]
Spencer6497: mmmm actually, yeah, good point: if you're running `dumpCmsToFile` to generate your `content.json` with a local `STRAPI_API`, the tests could fail. the way I see it, to prevent this, we could either 
* trust the devs to either have an updated, accurate `content.json`
* write some kind of wrapper around `getStrapiEntry` in the test itself to only use the deployed API when running locally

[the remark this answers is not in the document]
pgurusinga: I have some doubts about the first option 😄 I rarely or never run `dumpCmsToFile`
Can we set the env var just for this test suite (overriding it)? That way, we won’t need to add a wrapper.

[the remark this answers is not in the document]
pgurusinga: after discussion in Slack; maybe we are overthinking a problem that doesn’t exist yet and let's see what would happen ✅

pgurusinga (approved):

tests/integration/verifyCmsContent.test.ts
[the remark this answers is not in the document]
Spencer6497: ^ after PMing about the above, decided against this for now, as it's somewhat of an edge case. seeing as the integration tests don't run automatically locally, and a failure in CI (with passing local test(s)) will lead to further investigation, I think this is fine as-is

PR #1453, opened 2024-11-20
Remove unneeded dependencies

aaschlote:
### Changes
Remove dependencies react-dropzone-esm and @types/sanitize-html

### DIFF
--- a/data/opensource-licenses.json
+++ b/data/opensource-licenses.json
@@ -25,6 +25,15 @@
     "licenseFile": "a2j-rechtsantragstelle/node_modules/@digitalservicebund/icons/README.md",
     "direct": true
   },
+  "@pdf-lib/fontkit": {
+    "licenses": "MIT",
+    "repository": "https://github.com/Hopding/fontkit",
+    "publisher": "Andrew Dillon",
+    "email": "andrew.dillon.j@gmail.com",
+    "path": "a2j-rechtsantragstelle/node_modules/@pdf-lib/fontkit",
+    "licenseFile": "a2j-rechtsantragstelle/node_modules/@pdf-lib/fontkit/README.md",
+    "direct": true
+  },
   "@remix-run/express": {
     "licenses": "MIT",
     "repository": "https://github.com/remix-run/remix",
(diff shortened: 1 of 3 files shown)


data/opensource-licenses.json
```diff
@@ -316,28 +333,6 @@
     "licenseFile": "a2j-rechtsantragstelle/node_modules/remix-validated-form/README.md",
     "direct": true
   },
-  "sanitize-html": {
```
pgurusinga, line 319: do we even use this? :O

data/opensource-licenses.json
```diff
@@ -175,15 +201,6 @@
     "licenseFile": "a2j-rechtsantragstelle/node_modules/isbot/LICENSE",
     "direct": true
   },
-  "license-checker-rseidelsohn": {
```
pgurusinga, line 178: we are using this, but why deleted?

aaschlote: this library is used to generated this file, so it's not needed at all here. It was removed from this file now, because in this [commit](https://github.com/digitalservicebund/a2j-rechtsantragstelle/commit/b7e280abd564e771e375533267e2891b8c35460f) was moved from the scope `optionalDependencies` to `devDependencies` and this file wasn't updated at all since this change

data/opensource-licenses.json
aaschlote: Nope, it was removed from this [commit](https://github.com/digitalservicebund/a2j-rechtsantragstelle/commit/a8940a082118bbbaa46a18ce4e23c8591846286a#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)

pgurusinga (approved):

data/opensource-licenses.json
Spencer6497: does that mean we can remove the [@types package](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/2a604a6c92c1a32d0c0f113bfc98061cfba40410/package.json#L116) too?

PR #1454, opened 2024-11-20
fix(repair broken cached-checkout-install step)

Spencer6497:
This PR fixes a bug where `npm ci` was being called every time `cached-checkout-install` was being called, regardless if `node_modules` was found in the cache or not. This should reduce our pipeline times by a considerable amount :)


.github/actions/cached-checkout-install/action.yml
```diff
@@ -14,16 +14,14 @@ runs:
       env:
         cache-name: cache-node-modules
       with:
-        path: ~/.npm
+        path: ./node_modules
```
pgurusinga, line 17: fyi https://github.com/actions/cache/blob/main/examples.md#node---npm

> Note It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci

Spencer6497: saw that; two thoughts:
1. to avoid this breaking across Node versions, we could hash the node version and add it to the key
2. npm ci only runs on a cache miss, so node_modules won't exist in that case

pgurusinga: Thank you for the explanation. I checked the build time for all previous pull requests and didn't see any significant improvement. Am I missing something?

Spencer6497: there's a _slight_ speedup for pull requests, but PRs only trigger the first 3 steps of the CI flow; every other step also uses `cached-checkout-install`, so full deployments should be faster by a few minutes, maybe 3 or so

pgurusinga (approved):

aaschlote (approved):
LGTM

PR #1459, opened 2024-11-21
feat: override to patched version cross-spawn

pgurusinga:
image


chohner (approved):

package.json
```diff
@@ -149,7 +149,8 @@
   "overrides": {
     "@remix-run/server-runtime": {
       "cookie": "^0.7.2"
-    }
+    },
+    "cross-spawn": "7.0.6"
```
aaschlote, line 153: it seems the problem comes only from the `@remix-run/dev`, can you try only to overwrite in this dependency?

pgurusinga: it's coming from multiple places, not only remix dev
image


image

PR #1467, opened 2024-11-22
feat: update vulnerable libs in docker

pgurusinga:
# Before
Screenshot 2024-11-22 at 17 42 40

# After
Screenshot 2024-11-22 at 17 36 06

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ARG APP_IMAGE=app
 FROM node:20-alpine AS app-base
 WORKDIR /a2j
 
-# Update vulnerable libraries in the base stage
+# TODO: Remove this Linux update and the force update for cross-spawn once the issue is fixed
 RUN apk update && \
   apk add --no-cache libcrypto3=3.3.2-r1 libssl3=3.3.2-r1
 COPY package.json package-lock.json ./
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,8 +4,17 @@ ARG APP_IMAGE=app
...
+# Update vulnerable libraries in the base stage
+RUN apk update && \
+  apk add --no-cache libcrypto3=3.3.2-r1 libssl3=3.3.2-r1
 COPY package.json package-lock.json ./
-RUN npm ci --omit=dev --omit=optional
+RUN npm ci --omit=dev --omit=optional && \
+  # Update npm and fix cross-spawn vulnerability
+  npm uninstall -g cross-spawn && \
+  npm cache clean --force && \
+  find /usr/local/lib/node_modules -name "cross-spawn" -type d -exec rm -rf {} + && \
```
pgurusinga, line 16: I need to do this step because I tried removing this line and the one below it on my local setup. But when I ran Docker, the cross-spawn issue was still there.

image

Dockerfile
```diff
@@ -4,8 +4,17 @@ ARG APP_IMAGE=app
 
 FROM node:20-alpine AS app-base
 WORKDIR /a2j
+
+# Update vulnerable libraries in the base stage
+RUN apk update && \
```
aaschlote, line 9: this might be a temporally solution, pretty sure they will fix it at some point. Could we add a TODO comment or something to remember it in the future?

aaschlote (approved):
LGTM!

chohner:
Nice work! Regarding the `cross-spawn` dependency - you probably need to override every single occurance in the `package.json` for it to work. I think thats preferable over hacking the dockerfile?

PR #1474, opened 2024-11-25
chore(Dockerfile): explicit npm upgrade & docs

chohner:
split dependency upgrade into separate layers + comments


Dockerfile
```diff
@@ -25,13 +25,15 @@ FROM ${CONTENT_IMAGE} AS contentStageForCopy
...
 
-# TODO: Remove this Linux update and the force update for cross-spawn once the issue is fixed
-RUN apk update && \
-  apk add --no-cache dumb-init libcrypto3=3.3.2-r1 libssl3=3.3.2-r1 && \
-  npm uninstall -g cross-spawn && \
-  npm cache clean --force && \
-  find /usr/local/lib/node_modules -name "cross-spawn" -type d -exec rm -rf {} + && \
-  npm install -g cross-spawn@7.0.6 --force
+# Fixes https://scout.docker.com/vulnerabilities/id/CVE-2024-9143?s=alpine
+# Remove once node:20-alpine containes updated openssl
```
aaschlote, line 29: can we keep the `# TODO`? I believe at some point they are going to fix it, so we don't need to do this workaround

pgurusinga (approved):
image
besides the one-liner sonarlint thing

Dockerfile
chohner: The comment mentions to remove the manual upgrade once the vulnerability is fixed. Or do you want the explicit `TODO:`?

aaschlote: I’m concerned we might overlook this comment in the future and forget to remove these lines. I recommend adding a `TODO` keyword so that our linter reminds us to revisit this periodically. This way, we can check if the node-alpine image has been updated with the latest OpenSSL version. Although the update of the node-alpine Docker image is beyond our control, the `TODO` will help ensure we don't miss it when it’s eventually fixed.

But I'm fine not having it.

aaschlote (approved):
LGTM

PR #1484, opened 2024-11-27
feat(Utilize translations inside of template + PDFs)

Spencer6497:
This PR injects flow-specific translations to be used inside CMS templates, as well as in our pdf filler functions.

It will also add a PDF attachment to the Prozesskostenhilfe formular, if a user has required Belege.


Spencer6497:
> Just one question out of curiosity, are we planning to have multiple languages in our pdfs? I thought all the PDF would be in german. Asking it, because it has a future plan to make the website in multiple languages.

No the PDF should always be in German to my knowledge. `Translations` here is a bit of a misnomer; in this instance these are just strings that we're using in several places across the app that need to be accessible to both the CMS and PDF

Spencer6497:
> What about to create a new collection type related the PDF and we don't enable the localization option? So it's not related the translation and it might not have a problem in case we enable multiple languages in the future. image

actually during our daily today we decided to split the PDF string injection + the app translation injection into two separate PRs :)

Spencer6497:
> Hm then we would have to define them in two places, which is what I wanted to avoid.... but that begs a good question: how do we use the same strings, some being translated (in the flow) and some not? (in the pdf?) or do we actually want to translate these pdf strings too, since they're a part of the Anhang and more instructive to the user than the recipient? I'll ask design

OK so design indicated that support for multiple languages is sufficiently far in the future as to not be a current concern at all. They did indicate, however, that instead of keeping Belege translations inside of `/prozesskostenhilfe/formular`, we may want to keep them inside of a translation scope such as `pkhBelege`. This would require the modification of how we're injecting translations into the flow/pdf and might be a bit annoying but we'll see

aaschlote:
> > Hm then we would have to define them in two places, which is what I wanted to avoid.... but that begs a good question: how do we use the same strings, some being translated (in the flow) and some not? (in the pdf?) or do we actually want to translate these pdf strings too, since they're a part of the Anhang and more instructive to the user than the recipient? I'll ask design
> 
> OK so design indicated that support for multiple languages is sufficiently far in the future as to not be a current concern at all. They did indicate, however, that instead of keeping Belege translations inside of `/prozesskostenhilfe/formular`, we may want to keep them inside of a translation scope such as `pkhBelege`. This would require the modification of how we're injecting translations into the flow/pdf and might be a bit annoying but we'll see

Ok, maybe we can keep like this, but in case you need these string in other parts of the pdf I would recommend to move in another strapi collection, because in theory these strings are not translations, but PDF Texts and it shouldn't be translated at all.

Spencer6497:
Currently blocked; waiting for design/content to add belege strings in the translations

PR #1488, opened 2024-11-27
refactor: get airport names in the server

aaschlote:
### Changes
 
Get the airport names in the server side and avoid to load the data.json in the client

Before
image

After
image


judithmh:
LGTM :)

judithmh (approved):

chohner:
hard to review the code but cool 👍🏼

chohner:
just not familiar with the airport logic, just wanted to say thanks for taking care of this :)

Another thought: I hope once we move markdown rendering into the back-end we could also populate this data there (instead of fetching inside the component) but its a cool solution for now!

aaschlote:
> just not familiar with the airport logic, just wanted to say thanks for taking care of this :)
> 
> Another thought: I hope once we move markdown rendering into the back-end we could also populate this data there (instead of fetching inside the component) but its a cool solution for now!

Ideally for the Overview component, we will move the logic to get airports names to stringReplacement using the translation, similar as I did for the Migration component. So this solution is only temporary until we move the Overview component in a better way (not hard code)

app/routes/step-over-airports.ts
```diff
@@ -0,0 +1,15 @@
+import { type LoaderFunctionArgs } from "@remix-run/node";
+import { getAirportNameByIataCode } from "~/domains/fluggastrechte/services/airports/getAirportNameByIataCode";
+
+export function loader({ request }: LoaderFunctionArgs) {
+  const { searchParams } = new URL(request.url);
+  const ersterZwischenstopp = searchParams.get("ersterZwischenstopp") ?? "";
+  const zweiterZwischenstopp = searchParams.get("zweiterZwischenstopp") ?? "";
+  const dritterZwischenstopp = searchParams.get("dritterZwischenstopp") ?? "";
```
pgurusinga, line 8: @aaschlote we need to safely parse these params as we did in the other endpoint (data list options) but I guess we can use Zod safe parse right?

app/domains/fluggastrechte/components/useStepOverAirports.tsx
```diff
@@ -0,0 +1,53 @@
...
+    ersterZwischenstoppName: "",
+    zweiterZwischenstoppName: "",
+    dritterZwischenstoppName: "",
+  });
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const response = await fetch(
+          `${STEP_OVER_AIRPORTS_PATH}?ersterZwischenstopp=${ersterZwischenstopp}&zweiterZwischenstopp=${zweiterZwischenstopp}&dritterZwischenstopp=${dritterZwischenstopp}`,
```
pgurusinga, line 29: @aaschlote we should utilise URL utility to append query params, if user has 1 stop then the query would be
`?ersterZwischenstopp="BER"&zweiterZwischenstopp=""&dritterZwischenstopp=""` and in the step-over-airports we default to empty string again

`const ersterZwischenstopp = searchParams.get("ersterZwischenstopp") ?? "";`

pgurusinga:
```ts
try {
  const params = new URLSearchParams(
    [
      ["ersterZwischenstopp", ersterZwischenstopp],
      ["zweiterZwischenstopp", zweiterZwischenstopp],
      ["dritterZwischenstopp", dritterZwischenstopp],
    ].filter(([_, value]) => value)
  );

  const url = `${STEP_OVER_AIRPORTS_PATH}?${params.toString()}`;

  const response = await fetch(URL);
...
```

probably something like this...

PR #1489, opened 2024-11-27
Remove restore-keys

aaschlote:
### Changes
Remove the `restore-keys` option, so we only get the cache from the key. I have noticed sometimes we are getting a node-module cache in dependent bot PRs, even we have a new dependency update.

image


Spencer6497 (approved):
nice catch 🎣

PR #1493, opened 2024-11-28
Refactoring formular server

aaschlote:
### Changes

- Decoupling formular.server in two pieces
  - buildFormularServerTranslations: build formular server translations


app/services/flow/formular/getFormularServerAdditionalData.ts
```diff
@@ -0,0 +1,39 @@
+import type { Context } from "~/domains/contexts";
+import type { FlowId } from "~/domains/flowIds";
+import type { Flow } from "~/domains/flows.server";
+import type { ArrayConfig } from "~/services/array";
+import { getArraySummaryData } from "~/services/array/getArraySummaryData";
+import type { CookieHeader } from "~/services/session.server";
+import { getMigrationData } from "~/services/session.server/crossFlowMigration";
+
+export const getFormularServerAdditionalData = async (
```
chohner, line 9: could we keep this as separate functions? they take different configuration and return different data, which to me sounds like its two different functions

chohner (commented):
good initiative to clean up formular.server!
But I have the feeling that instead of reducing the complexity we are just hiding it inside functions now 😅

app/services/flow/formular/getFormularServerAdditionalData.ts
aaschlote: If I do it, it would be the same as before, because it has two functions that it retrieves some particula data. I created this single function to merge the both actions and remove 11 lines from the `formular.server.ts`

aaschlote:
> good initiative to clean up formular.server! But I have the feeling that instead of reducing the complexity we are just hiding it inside functions now 😅

I noticed that some parts are similar to what we have in vorabcheck.server and result.server, while others differ. Refactoring this thoroughly would take some time. For this PR, my focus is simply to move certain parts into separate functions. However, we can revisit and improve it further in the future.

app/services/flow/formular/getFormularServerAdditionalData.ts
Spencer6497: I would honestly also prefer they be separate - even at the cost of extra LOC in this instance

aaschlote: well, it's two agains one, so I'll remove it and later we can think how to make it better 😃

PR #1500, opened 2024-11-28
chore(add Android Browser e2e configuration)

Spencer6497:
This PR adds an additional browser type, Android Mobile, to be tested concurrently during our e2e test suites, as most of our sites users are using this browser and we're not testing mobile viewports at all.

Yet to be seen how much slower this makes the e2e tests in CI, hence the PR


chohner (approved):
nice! I verified the setting locally. I think we can also reduce the timeouts (both per-test and per-expect) but thats for a separate PR.

rbrtrfl (approved):

aaschlote:
@Spencer6497 as we talked last Friday, can you check if it's possible to remove a few tests that it's not necessary to run in a second device? For example CSRF, A11y checks and etc.

PR #1508, opened 2024-12-02
test(flows): add coverage

chohner:
- adds coverage for step Ids visited during flow tests
- sets upper bound for untested steps


app/domains/__test__/flows.test.ts
```diff
@@ -105,21 +126,54 @@ describe("flow tests", () => {
...
+    const totalMissingStepCount = missingStepsEntries.reduce(
+      (total, [_, missingSteps]) => total + missingSteps.length,
+      0,
+    );
+
+    console.warn(
+      `Total of ${totalMissingStepCount} untested stepIds: `,
+      Object.fromEntries(missingStepsEntries),
+    );
+    expect(totalMissingStepCount).toBeLessThanOrEqual(126);
```
rbrtrfl, line 177: So this number is our current lucky number of untested steps? Do we want to remove this assertion after having filled the gaps or how do we plan to use it?

Suggestion: I think it would be useful to get the missing steps printed out within the failing assertion logs, not as a `console.warn` (under the assumption that we're moving towards zero missing steps).

app/domains/__test__/flows.test.ts
```diff
@@ -105,21 +126,54 @@ describe("flow tests", () => {
...
+
+          expect(visitedSteps).toEqual(expectedSteps);
+        });
       });
-    });
+    },
+  );
+
+  test("all steps are visited", () => {
+    const missingStepsEntries = Object.entries(allVisitedSteps)
```
rbrtrfl, line 158: I find this variable name misleading, because it basically holds all visited steps, not only the missing steps, or?

chohner: no, its only missing steps (note the `.map()` call)

app/domains/__test__/flows.test.ts
chohner: Right, this blocks anyone from adding new untested flow steps. Once we actively put effort into reducing this number we can keep decreasing it

app/domains/__test__/flows.test.ts
```diff
@@ -69,7 +87,10 @@ function getEnabledSteps({
  * - system under test should be in a certain state (step)
  */
 
-describe("flow tests", () => {
+describe.sequential("state machine form flows", () => {
```
Spencer6497, line 90: what's the purpose behind the `.sequential`/`.concurrent` calls?

Spencer6497 (approved):
excellent! This'll help a lot 👍

aaschlote (changes requested):
I appreciate the idea, but I’m not entirely on board with how this check has been implemented.

In my opinion, having a unit test fail because a certain number of tests are missing doesn’t make much sense. Why? Tests should be an integral part of the development process, and their absence shouldn't act as a blocker. If tests are missing, the question to ask is: why weren’t they created in the first place?

I believe this functionality would be more effective as a script that runs periodically, analyzing how many steps are missing tests. This way, we can address gaps over time without hindering the development process.

Moreover, I would strongly challenge any developer who creates new features or refactors existing ones without adding tests. Failing to write or include necessary tests reflects a significant issue in the quality of the product. Ensuring that every change is accompanied by appropriate tests is a fundamental responsibility, and it’s crucial for maintaining high standards in our work.

aaschlote (approved):
As we discussed, let's keep like this for now and iterate later  😃

PR #1517, opened 2024-12-03
chore(Kopfzeile): use kern-ux/native

chohner:
- use [native kern kopfzeile](https://kern-ux-plain-kern-ux-783aa9be89f1fecf3bdc3d2f5004d885b31ec77d9.usercontent.opencode.de/?path=/docs/komponenten-kopfzeile--docs)
- WARNING: imports KERN styling, which affects the layout of the whole site!


Spencer6497:
using the [react component](https://www.kern-ux.de/design-system/kopfzeile/) seems to work alright

frederike-ramin:
We tried ways to implement the original KERN component. However, the method `register` is causing hydration errors and does not agree with remix. As the deadline is very close, we decided to instead fix the issues with the previous re-implementation and get back to the KERN people about that issue to be able to integrate the correct componnnt later on.

PR #1518, opened 2024-12-03
feat(pdf-service): append hinweisblatt to pkh pdf

rbrtrfl:
https://app.asana.com/0/1203259475859667/1208875879671542


app/domains/prozesskostenhilfe/services/pdf/__test__/prozesskostenhilfePdfFromUserdata.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    vi.mocked(appendPagesToPdf).mockResolvedValue(mockUint8Array);
+    vi.mocked(loadHinweisblatt).mockResolvedValue(mockPdfDocument);
+
+    await prozesskostenhilfePdfFromUserdata(userData);
+
+    expect(pdfFillReducer).toHaveBeenCalled();
+    expect(fillPdf).toHaveBeenCalled();
+    expect(addMetadataToPdf).toHaveBeenCalled();
+    expect(appendPagesToPdf).toHaveBeenCalled();
+    expect(loadHinweisblatt).toHaveBeenCalled();
```
chohner, line 62: Not a blocker but personally I prefer to test the observable outcome, rather than the implementation details. But that would mean inspecting the pdf, right?

app/domains/prozesskostenhilfe/services/pdf/loadHinweisblatt.ts
```diff
@@ -0,0 +1,25 @@
...
+  yPositionsDruckvermerk: number[],
+  xPositionsDruckvermerk: number,
+) {
+  const filepath = path.resolve(
+    path.join(
+      process.cwd(),
+      dataDirectory,
+      "prozesskostenhilfe",
+      "Formular_Hinweisblatt_Prozess_Verfahrenskostenhilfe.pdf",
+    ),
```
ekl176, line 17: @rbrtrfl  can we add a test to fail in case that file gets renamed or moved?

app/domains/prozesskostenhilfe/services/pdf/__test__/prozesskostenhilfePdfFromUserdata.test.ts
rbrtrfl: For the majority of tests I would agree that observing the outcome is the way to go. But this file/function to me is more of / merely of a container of a series subtasks. For each subtask (e.g. adding the metadata) a test that expects a particular outcome is very useful. And in combination, the 'container' solely testing the implementation details should be enough IMO.

Nevertheless, I think this is an interesting topic to be discusses in our dev exchange?

PR #1519, opened 2024-12-03
feat(pdf-service): print name in signature form field of pkh pdf...

rbrtrfl:
...when `versandArt` is `digital`

Note: The signature area is not a form field. Therefore the name needs to be printed into it manually.

https://app.asana.com/0/1203259475859667/1208875879671528


Spencer6497 (approved):
Excellent work 🚀 send it

PR #1530, opened 2024-12-06
fix(pdf-service): styling fixes after move to pdfkit

rbrtrfl:
https://app.asana.com/0/1203259475859667/1208782862325370


app/domains/beratungshilfe/services/pdf/checklist/__test__/createChecklistSteps.test.ts
```diff
@@ -7,72 +7,64 @@ import { abgabeContext } from "~/domains/shared/formular/abgabe/context";
...
-    );
+    expect(mockPDFDocument.text).toHaveBeenCalledWith("4. Antrag abgeben");
     expect(mockPDFDocument.list).toHaveBeenCalledWith(
       [
         "Unterlagen zu Ihrem rechtlichen Problem",
         "Kopie Ihres aktuellen Mietvertrags",
       ],
+      expect.any(Number),
+      expect.any(Number),
       expect.any(Object),
```
Spencer6497, line 36: I'm a bit confused on what this code is actually asserting? Why do we need these here?

Spencer6497 (approved):
LGTM! Minor comment for clarification but otherwise great :)

app/domains/beratungshilfe/services/pdf/checklist/__test__/createChecklistSteps.test.ts
rbrtrfl: The pdfs for `abgabeArt` `ausdrucken` and `online` are not the same. They have slightly different content. The test is asserting that some of the relevant content is actually fed into the methods composing the pdf. The two lines you've highlighted are arguments of the `list` method (x and y position). But only the first argument is relevant, hence the arbitrary `expect.any(Number)`.

PR #1533, opened 2024-12-06
chore(update Freibetraege for 2024)

Spencer6497:
This PR simplifies and updates the Freibetrag for PKH/BerH by simplifying the interfaces, converting magic strings into constants, and moving year-by-year Freibetraege logic to a map for easy updates.

- **simplify freibetrag interface(s)**
- **update freibetraege for 2024, use year map for ease of updating, replace magic strings and update tests**
- **update unit test to ensure fallback works, regardless of whatever the actual, current year is**


app/domains/beratungshilfe/vorabcheck/__test__/freibetrag.test.ts
```diff
@@ -1,157 +1,197 @@
 import {
-  freibetrag,
+  BASE_ALLOWANCE as baseAllowance,
```
rbrtrfl, line 2: nit: I must admit that this feels like an unnecessary dislocation to reassign the import variable and switching their name back and forth. Maybe use distinct variable names like
```
const BASE_ALLOWANCE_CENTS = BASE_ALLOWANCE * 100
const incomeAllowanceCents = incomeAllowance * 100
```

app/domains/beratungshilfe/vorabcheck/freibetrag.ts
```diff
@@ -1,6 +1,64 @@
...
+type Freibetraege = {
+  incomeAllowance: number;
+  partnerAllowance: number;
+  dependentAdultAllowance: number;
+  children15To18Allowance: number;
+  children7To14Allowance: number;
+  childrenBelow6Allowance: number;
+};
+
+export const BASE_ALLOWANCE = 572;
```
rbrtrfl, line 14: question: when do you use `SCREAMING_SNAKE_CASE` and when `camelCase` for the constants?

rbrtrfl (approved):
Very nice work on the functionality with the map! Just left two comments, but nothing blocking.

app/domains/beratungshilfe/vorabcheck/freibetrag.ts
```diff
@@ -1,6 +1,64 @@
...
+    partnerAllowance: 619,
+    dependentAdultAllowance: 496,
+    children15To18Allowance: 518,
+    children7To14Allowance: 429,
+    childrenBelow6Allowance: 393,
+  },
+};
+export const latestFreibetraegeYear = Number(
+  Object.keys(freibetraegePerYear).at(-1),
+);
```
chohner, line 36: relying on the object order seems a little brittle. maybe instead  `Math.max([...Object.keys(freibetraegePerYear)])`?

app/domains/beratungshilfe/vorabcheck/freibetrag.ts
```diff
@@ -1,6 +1,64 @@
...
+
+export function getFreibetraege() {
+  const currentYear = today().getFullYear();
```
chohner, line 39: Making this a parameter would REALLY simplify the testing. I think its OK to leave this to the call-site

chohner (approved):
lgtm, to small remarks

app/domains/beratungshilfe/vorabcheck/freibetrag.ts
Spencer6497: `SCREAMING_SNAKE_CASE` should be used for top-level or global constants, and `camelCase` should be used for scoped constants (imo, happy to entertain other paradigms :))

PR #1534, opened 2024-12-06
feat(textarea): add maxlength via strapi defaulting to 5000

m0dh4x:
- schema validation with zod for BE and FE validation

----------
Co-authored-by: chris <christoph.hohnerlein@digitalservice.bund.de>


app/components/inputs/__test__/Textarea.test.tsx
```diff
@@ -27,15 +27,24 @@ afterEach(() => {
 describe("Textarea component", () => {
   it("renders without errors", () => {
     const componentName = "test-textarea";
+    const maxLength = 5000;
```
chohner, line 30: why does this need to be specified here? I think we should just check that:

1. a maxlength is set by default
2. passing a maxlength sets it

app/components/inputs/__test__/Textarea.test.tsx
```diff
@@ -154,4 +172,78 @@ describe("Textarea component", () => {
 
     expect(textarea.getAttribute("rows")).toEqual(TEXT_AREA_ROWS.toString());
   });
+
+  it("should enforce maxLength in textarea", () => {
```
chohner, line 176: Imo we need to test the platform (aka checking that the `<textarea>` element is actually working like the HTML spec)

chohner: (also, this test is already covered by the following one)

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -76,6 +76,7 @@ export const FeedbackFormBox = ({
           placeholder={feedbackTranslations["placeholder-feedback"]}
           role="status"
           innerRef={textAreaReference}
+          maxLength={TEXTAREA_CHAR_LIMIT}
```
chohner, line 79: isn't this the default?

m0dh4x: removed

app/components/inputs/__test__/Textarea.test.tsx
m0dh4x: yes, that makes sense. sry is a remained over from the previous implementation. will remove it

app/components/userFeedback/FeedbackFormBox.tsx
aaschlote, line 79: if the default value for this props is `TEXTAREA_CHAR_LIMIT`, why are we setting it here again?

app/components/inputs/__test__/Textarea.test.tsx
```diff
@@ -154,4 +167,40 @@ describe("Textarea component", () => {
 
     expect(textarea.getAttribute("rows")).toEqual(TEXT_AREA_ROWS.toString());
   });
+
+  it("should enforce maxCharacterLength limit", () => {
+    const maxLength = 10;
+    getInputProps.mockImplementationOnce(() => ({
+      id: "componentName",
+      maxLength: maxLength,
+      onChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => {
```
aaschlote, line 176: not sure if this test is acceptable, because here you forcing the event onChange removing the full text from the size of the maxLength value. Normally the HTML of the browser would do it. As this library can't reproduce this behaviour, we shouldn't test it.

app/components/userFeedback/FeedbackFormBox.tsx
m0dh4x: I had already changed it, but after undoing it, it was there again

PR #1535, opened 2024-12-09
fix(pdf-service): fix bug with German umlauts not rendered correctly in pdf

rbrtrfl:
I could only reproduce this bug with the arc browser. The root cause were two titles in `bargeld` and `wertsachen` where the 'ü' of `Eigentümer:in` was written as a NFD (normalized form canonical decomposition) which means the 'u' was typed first, followed by the combining diacritical marks. `pdfkit` does not seem to properly support encoding these characters in few cases (arc browser).
Proposal: Added a linter rule that detects NFDs.

https://app.asana.com/0/1203259475859667/1208834671333919


chohner (approved):

aaschlote (approved):
LGTM!

Out of curiosity, what are arc browsers? Any example?

rbrtrfl:
> LGTM!
> 
> Out of curiosity, what are arc browsers? Any example?

There is the one arc browser. Never heard of it before this bug: https://arc.net/

PR #1540, opened 2024-12-09
chore(lefthook): move lint to pre-push

chohner:
Linting now takes quite a bit of time due to requiring a typescript build

suggestion:
 - move linting hook from pre-commit to pre-push
 - run in parallel to unit tests

reason: we probably all have linting set up in our IDEs, the hook should just be a failsafe


aaschlote (approved):
LGTM

rbrtrfl (approved):

PR #1549, opened 2024-12-11
add aria-required attribute to input

judithmh:
aria-required attribute set to false if no inputRequired error in Strapi for input field set otherwise set to true


chohner (approved):
lgtm

PR #1552, opened 2024-12-13
Feat/validate 3 hours delay and arrivalDate can't be before departureDate

m0dh4x:
This is the PR for new validations in ZOV like `actualArrivalTime` should be at least 3hours later than `departureTime` or `actualArrivalDate` cant be before `departureDate`.


m0dh4x:
closed - using a more generalistic approach from @aaschlote

PR #1554, opened 2024-12-16
refactor(pdf-service): remove font size changes in pdf form

rbrtrfl:
- remove the font size changes and checked for implications on all cases
- change default font size for form fields to 10
- move s.A. string to const (not exactly this scope, but the pr seems small enough)


app/services/pdf/pdf.server.ts
```diff
@@ -14,6 +15,5 @@ export function changeStringField(stringField: StringField, form: PDFForm) {
   // This is because it impacts letter spacing e.g. in the fourth column of "G_ausgaben".
```
chohner, line 15: this comment could probably also be improved?

app/services/pdf/pdf.server.ts
```diff
@@ -14,6 +15,5 @@ export function changeStringField(stringField: StringField, form: PDFForm) {
...
   formField.setMaxLength();
   formField.setText(stringField.value);
-  const fontSize = getFontSizeFieldValue(stringField.name);
-  formField.setFontSize(fontSize);
+  formField.setFontSize(FONT_SIZE_DEFAULT);
```
chohner, line 18: do we even need to set a font size?

chohner (approved):
lgtm

app/services/pdf/pdf.server.ts
rbrtrfl: I'd say yes, because many form fields have a font size property `0` which means auto scale. That doesn't look good:
Screenshot 2024-12-16 at 10 19 53

PR #1556, opened 2024-12-16
feat(Add Tests for Failing Form Validation)

Spencer6497:
This PR aims to add a test suite against failing form validation, i.e. when invalid values are entered for form elements.

- **extract validated form element to its own component and replace instances thereof**
- **build out basic component test scaffolding**
- **build out test suite for Input Component with integer schema**


app/components/form/__test__/ValidatedFlowForm.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+          ]);
+        const nextButton = getByText("NEXT");
+        expect(nextButton).toBeInTheDocument();
+        expect(queryByText("error")).not.toBeInTheDocument();
+        fireEvent.blur(getByPlaceholderText("input"));
+        await expectErrorToExist(errorText);
+        fireEvent.click(nextButton);
+        await waitFor(() => {
+          expect(getByPlaceholderText("input")).toHaveFocus();
+        });
```
frederike-ramin, line 63: Should we check that the error exists after submitting, too?

app/components/form/__test__/ValidatedFlowForm.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+describe("ValidatedFlowForm", () => {
+  it("should render", () => {
+    fieldNameValidatorSpy.mockImplementationOnce(vi.fn());
+    const { getByText } = renderValidatedFlowForm([]);
+    expect(getByText("NEXT")).toBeInTheDocument();
+  });
+
+  describe("Input Component", () => {
+    describe("Intger Schema", () => {
```
frederike-ramin, line 37: Small typo here.
And: Is there a reason you test three failing cases? I would have expected one failing and one succeeding one per form component (input, checkbox, select etc). Just because we're not testing that our validations work but rather that *if* a validation fails, it is handled correctly by the form. Does that make sense?

app/components/form/__test__/ValidatedFlowForm.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+        ),
+      },
+    ],
+    { initialEntries: ["/"] },
+  );
+  return render(<RouterProvider router={router} />);
+}
+
+function buildStrapiInputComponent(
+  errorCodes: Array<z.infer<typeof StrapiFieldErrorSchema>>,
```
frederike-ramin, line 150: Wondering if we should move these into the `factories` folder? Because I imagine there will be more? esp. for the Strapi components we even already have a folder with similar stuff, I think: `/tests/factories/cmsModels`

frederike-ramin (commented):
Love the separation of the validated form component. With the tests, I think you're on the right path, just left a few comments. So appreciated that you made this a preliminary PR 🙏

PR #1558, opened 2024-12-18
PoC(XSD and XML in TS)

Spencer6497:
This PR is a minimal proof-of-concept for xJustiz communication in the app, enabling both
* Consuming XSDs and generating Typescript Type declarations
* Generation of valid XML from the aforementioned, strongly-typed javascript objects

**NOTES**
* npm run build:typesFromXSD generates the types. Unfortunately, it can only be run once (does not update the files if altered), and the generated output must be fixed (imports are incorrect)
* npm run build:xml runs a proof-of-concept script, `generateXMLFromObject.ts`, that uses the generated type to build a barebones object, build an XML from it, and print it out to file.
* the xcsd library we're using is massively out-of-date (5 years!), but I'm waiting to see if we can use [a newer fork](https://github.com/loanlink-nl/cxsd/issues/232). Unclear if this will fix the aforementioned issues.

- **utilize cxsd as proof-of-concept**
- **add proof-of-concept script for TS -> XML, re-generate types via cxsd**
- **tweak cxsd script and move output to sensible locations**
- **fix ts errors in generated files, add npm script for xml generation**


scripts/generateXMLFromObject.ts
```diff
@@ -0,0 +1,75 @@
...
+      return Object.fromEntries(
+        Object.entries(obj).map(([key, value]) => [
+          `tns:${key}`,
+          prependKeys(value),
+        ]),
+      );
+    }
+    return obj;
+  }
+  fs.writeFile(
```
khallad2, line 63: you can leverage `fs.promises.writeFile` for a cleaner asynchronous flow.

```
try {
    await fs.writeFile(
      "data/xml/generated.xml",
      builder.build(prependKeys(tsObject))
    );
  } catch (error) {
    console.error(error);
    throw error;
  }
```

scripts/generateXMLFromObject.ts
```diff
@@ -0,0 +1,75 @@
...
+    nachrichtenkopf: {
+      xjustizVersion: "",
+      eigeneNachrichtenID: "",
+      erstellungszeitpunkt: today(),
+    },
+  };
+  const builder = new XMLBuilder({
+    format: true,
+  });
+  function prependKeys(obj: any): any {
```
khallad2, line 50: why `any` ?
You might consider using a generic type or at least `unknown` to enforce some type safety.

scripts/generateXMLFromObject.ts
```diff
@@ -0,0 +1,75 @@
...
+        Object.entries(obj).map(([key, value]) => [
+          `tns:${key}`,
+          prependKeys(value),
+        ]),
+      );
+    }
+    return obj;
+  }
+  fs.writeFile(
+    "data/xml/generated.xml",
```
khallad2, line 64: Maybe consider making the output file path configurable rather than hardcoding.

Spencer6497:
Closing for now until we decide to pursue again :)

PR #1562, opened 2024-12-23
Validate multiple fields

aaschlote:
### Changes

- Move validatorForFieldNames to his own file and reuse in ValidateFlowForm.tsx and validateFormData.server.ts
- Create file `multipleFieldsFlowValidation` to setup multiple fields validation
- Create functions `validateDepartureAfterArrival`, `validateSameFlightPage`, `validateAnotherFlightPage` and `validateReplacementConnectionPage` to validate multiple fields values on the page
- Refactor `buildStepValidator` to add multiple fields validation in case exist on the flow and step id


chohner (approved):
Nice! You effectively built step validation (as opposed to our previous field validation), cool!

Two tiny comments, no blockers:
- Could we separate flowId & stepId (instead of passing the combined `pathname`)? I think it would be architecturally cleaner
- maybe `multi` instead of `multiple`?

aaschlote:
> Nice! You effectively built step validation (as opposed to our previous field validation), cool!
> 
> Two tiny comments, no blockers:
> 
> * Could we separate flowId & stepId (instead of passing the combined `pathname`)? I think it would be architecturally cleaner
> * maybe `multi` instead of `multiple`?

Hey @chohner  sure,

@m0dh4x what do you think about this renaming, is it better for you too?

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
```diff
@@ -0,0 +1,334 @@
...
+        direktAbflugsZeit: "14:00",
+        ersatzFlugAnkunftsDatum: "02.01.2024",
+        ersatzFlugAnkunftsZeit: "19:01",
+        bereich: "verspaetet",
+      });
+
+      expect(result.success).toBe(true);
+    });
+
+    it("should return success true given a departure date before three hours after the arrival and bereich annullierung", () => {
```
judithmh, line 213: "Für Fluggäste besteht in der Regel kein Anspruch auf Ausgleichszahlung, wenn deren Ersatzverbindung weniger als 1 Stunde früher gestartet und weniger als 2 Stunden später angekommen ist."
I think we have to still implement this rule (can also be done in another ticket but the test should then be removed or commented out)

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
```diff
@@ -0,0 +1,334 @@
...
+        bereich: "verspaetet",
+      });
+
+      expect(result.success).toBe(false);
+    });
+
+    it("should return success false given a departure date after the arrival", () => {
+      const result = validatorReplacementConnectionPage.safeParse({
+        direktAbflugsDatum: "02.01.2024",
+        direktAbflugsZeit: "14:00",
```
judithmh, line 277: maybe we could change this to "11:00" to make sure it really test for the date

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
```diff
@@ -0,0 +1,334 @@
...
+        andereErsatzverbindungAnkunftsDatum: "02.01.2024",
+        andereErsatzverbindungAnkunftsZeit: "19:01",
+        bereich: "verspaetet",
+      });
+
+      expect(result.success).toBe(true);
+    });
+
+    it("should return success true given a departure date before three hours after the arrival and bereich annullierung", () => {
```
judithmh, line 310: same as above:
"Für Fluggäste besteht in der Regel kein Anspruch auf Ausgleichszahlung, wenn deren Ersatzverbindung weniger als 1 Stunde früher gestartet und weniger als 2 Stunden später angekommen ist."
I think we have to still implement this rule (can also be done in another ticket but the test should then be removed or commented out)

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
```diff
@@ -0,0 +1,334 @@
...
+        direktAnkunftsZeit: "15:00",
+      });
+
+      expect(result.success).toBe(false);
+    });
+
+    it("should return success true given a departure date before the arrival", () => {
+      const result = validator.safeParse({
+        direktAbflugsDatum: "01.01.2024",
+        direktAbflugsZeit: "14:00",
```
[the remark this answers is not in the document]
aaschlote: the next test does it, you can check `should return success true given a departure time before the arrival`

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
```diff
@@ -0,0 +1,334 @@
...
+
+      expect(result.success).toBe(false);
+    });
+
+    it("should return success false given a departure date after the arrival", () => {
+      const result = validatorSameFlightPage.safeParse({
+        direktAbflugsDatum: "02.01.2024",
+        direktAbflugsZeit: "14:00",
+        tatsaechlicherAnkunftsDatum: "01.01.2024",
+        tatsaechlicherAnkunftsZeit: "15:00",
```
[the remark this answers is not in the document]
aaschlote: same as above, the next test check the time

m0dh4x:
@aaschlote multi sounds good for me :)

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
aaschlote: yeap, we can do it, but in the [ticket](https://app.asana.com/0/1207022303033556/1208686662646812) AC is only for Verspätung claims. Would you mind to check with UX and product about it?

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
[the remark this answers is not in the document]
judithmh: I saw it, but I am referring to that here we cannot be sure that the success comes from the date and not the time if the departure time is not later than the arrival time since we just test for success in general

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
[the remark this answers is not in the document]
judithmh: same as above: I meant to make sure that the successs comes from the departure date being before the arrival date and not the departure time before the arrival time

app/domains/fluggastrechte/formular/services/__test__/validation.test.ts
judithmh: same as in the other comments ;)

app/components/form/__test__/ValidatedFlowForm.test.tsx
```diff
@@ -13,10 +13,10 @@ import { getStrapiTextareaComponent } from "tests/factories/cmsModels/strapiText
 import { getStrapiTileGroupComponent } from "tests/factories/cmsModels/strapiTileGroupComponent";
 import ValidatedFlowForm from "~/components/form/ValidatedFlowForm";
 import type { StrapiFormComponent } from "~/services/cms/models/StrapiFormComponent";
-import * as buildStepValidator from "~/services/validation/buildStepValidator";
 import { checkedRequired } from "~/services/validation/checkedCheckbox";
 import { createDateSchema } from "~/services/validation/date";
 import { integerSchema } from "~/services/validation/integer";
+import * as validatorForFieldNames from "~/services/validation/stepValidator/validatorForFieldNames";
```
judithmh, line 19: Not super necessary since there is only one function in that file but maybe it could be changed to a direct import instead

app/services/validation/stepValidator/buildStepValidator.ts
```diff
@@ -0,0 +1,41 @@
...
+      fieldValidators[stepOrFieldName] = schemas[stepOrFieldName];
+    }
+  }
+
+  const baseSchema = z.object(fieldValidators);
+
+  const validationSchema = multiFieldsValidation
+    ? multiFieldsValidation(baseSchema)
+    : baseSchema;
+
```
pgurusinga, line 39: I was tracing the logic from bottom to top, and I got a bit confused at this point. It seems like `multiFieldsValidation` is optional, and I assumed that `buildStepValidator` can be called both with and without it, or am I wrong?

Could we also consider renaming baseSchema to singleFieldSchema? I think it might make the purpose clearer if that fits the context. What do you think?

app/services/validation/stepValidator/validatorForFieldNames.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function validatorForFieldNames(fieldNames: string[], pathname: string) {
+  const { flowId, stepId } = parsePathname(pathname);
+  const context = getContext(flowId);
+  const multiFieldsValidation = getMultiFieldsByStepIdValidation(
+    flowId,
+    stepId,
+  );
+
+  return buildStepValidator(context, fieldNames, multiFieldsValidation);
```
pgurusinga, line 14: OK, now I get it, multiFieldsValidation can be `undefined` so it will return single field validation, the function API is confusing a bit, could we somehow simplify the approach? I don't have a solution right now though 🙈 but it's not a blocker

app/components/form/__test__/ValidatedFlowForm.test.tsx
aaschlote: you need it due the `vi.spyOn` on this test

judithmh: ahh I see - thanks for the explanation :)

app/services/validation/stepValidator/validatorForFieldNames.ts
aaschlote: The idea is that you can setup by page (or step id) multi fields functions. Most of the time you don't need it, only when you need to validate two fields on the formular. The variable `multiFieldsValidation` will return a function in case the page has some setup for it, otherwise it'll return `undefined`

app/services/validation/stepValidator/buildStepValidator.ts
aaschlote: I renamed the variable to `validationFieldsSchema`, it should be more clear now 😄

judithmh (approved):

app/domains/fluggastrechte/formular/services/validation.ts
```diff
@@ -0,0 +1,224 @@
...
+      return z.NEVER;
+    }
+
+    if (
+      isStartTimestampLessThanThreeHours(departureDateTime, arrivalDateTime)
+    ) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "arrivalThreeHoursLessThanDeparture",
+        path: ["tatsaechlicherAnkunftsDatum"],
```
pgurusinga, line 174: Since this is an array, is it possible to combine it with the AnkunftsDatum and AnkunftsZeit so we won't need to call `ctx.addIssue()` again, or it has to be separated?

```suggestion
        path: ["tatsaechlicherAnkunftsDatum", "tatsaechlicherAnkunftsZeit"],
```

pgurusinga (approved):
beside my comment

app/domains/fluggastrechte/formular/services/validation.ts
pgurusinga: if we can put it together, then we can reduce all `ctx.addIssue()` in this file

aaschlote: I tried, but it doesn't work. The remix hook `useField` don't recognise arrays on the fieldName parameter, only a normal string.

PR #1573, opened 2024-12-31
chore(Add more security e2e tests)

Spencer6497:
This PR adds two test suites for additional security testing, namely:
1. Testing the server response headers for correct values
2. Ensuring invalid HTTP operations (DELETE, PUT, PATCH, etc) are unable to be executed, and result in an error.


tests/e2e/common/security.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  Vary: "Accept-Encoding",
+};
+
+test.describe("Security Tests", () => {
+  test("The server should send a response including the correct response headers", async ({
+    page,
+  }) => {
+    const response = await page.request.get("/");
+    await expect(response).toBeOK();
+    const headers = Object.entries(response.headers());
```
chohner, line 20: is the conversion to entries necessary? Not sure what `.headers()` returns exactly but I hope there is a `.contains()`, `.has()` or `.get()` API? Then instead of the `headers.some()` call we could simplify to `headers.get(key) === value`...

tests/e2e/common/security.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    Object.entries(expectedHeaders)
+      .map(([key, val]) => [key.toLocaleLowerCase(), val])
+      .forEach(([key, expectedVal]) => {
+        const responseHasExpectedHeader = headers.some(
+          ([name, value]) => name === key && value === expectedVal,
+        );
+        if (!responseHasExpectedHeader) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `Header ${key} with value ${expectedVal} was not found in the response`,
```
chohner, line 30: We don't know whether the value isn't there or a different value. so maybe instead we could log `Header {key} was expected to be X but instead was Y`

tests/e2e/common/security.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+import { defaultHeaders } from "~/rootHeaders";
+
+const expectedHeaders = {
+  ...defaultHeaders,
+  "Cache-Control": "no-store",
+  Connection: "keep-alive",
+  "Content-Encoding": "gzip",
+  "Content-Type": "text/html; charset=utf-8",
+  "Transfer-Encoding": "chunked",
+  Vary: "Accept-Encoding",
```
chohner, line 11: I wonder if we actually care about all of these but guess it can't hurt for now. but we can probably reduce this in the future if we stumble across this

chohner (approved):
lgtm, some small comments

tests/e2e/common/security.spec.ts
Spencer6497: probably not, I just added them for completeness 😅

tests/e2e/common/security.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { test, expect } from "@playwright/test";
+import { defaultHeaders } from "~/rootHeaders";
+
+const expectedHeaders = {
```
aaschlote, line 4: I’m not sure if this test is necessary, as we already have a unit test in `app/__test__/rootHeaders.test.ts` that checks the same functionality. If an issue arises in the code, the unit test will catch it. The main difference here is that this test interacts with the application directly, making it more effective in detecting changes that might occur in the server-side library of the application.

tests/e2e/common/security.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+          // eslint-disable-next-line no-console
+          console.warn(
+            `Header ${key} was expected to be ${expectedVal} but instead was ${actualValue}`,
+          );
+        }
+        expect(responseHasExpectedHeader).toBe(true);
+      });
+  });
+
+  test("Invalid HTTP operations should yield an error", async ({ page }) => {
```
aaschlote, line 35: we have e2e tests running for Chrome and Android, do we need to run it twice?

Spencer6497: nope, i'll add it to the ignore list

tests/e2e/common/security.spec.ts
Spencer6497: yeah i noticed the redundancy, and like you said it's unlikely this will ever fail when the unit test passes, but I figured it's a sufficiently low overhead test so why not add another layer of certainty

chohner: I think its a good idea to test the existence of security-related headers from the outside of the app. out of the two, I would rather remove the remix header tests and keep these here

pgurusinga: I think unit tests just make sure we’re getting the right object with the correct key and value (**_correct me if I am wrong_**).
But when we test with a real browser, we can see how the CSP actually behaves. Even a small typo or missing comma could make the headers invalid in the browser, though they might still seem fine in a unit test

pgurusinga (approved):

tests/e2e/common/security.spec.ts
pgurusinga: ups, I meant Headers not CSP lol

aaschlote: While E2E tests provide a more comprehensive validation by ensuring the expected rules are functioning effectively, they should be used judiciously. Adding every test as an E2E test can slow down the deployment process and potentially cause issues in the pipeline, as we’ve experienced in the past (though this particular test runs quickly and without problems).

To conclude, this test may not be necessary as it introduces redundancy and could be handled more efficiently with a unit test. However, if you prefer to keep it, I’m okay with that, though I wouldn’t recommend it.

aaschlote (approved):

PR #1581, opened 2025-01-06
performance: bundle translation requests

Spencer6497:
This PR aims to further reduce app latency by making a number of "cleverer" calls to the server, namely in batching calls to similar endpoints.


app/services/cms/index.server.ts
```diff
@@ -64,6 +85,24 @@ export const fetchTranslations = async (
...
+      return [
+        scopedTranslations.scope,
+        Object.fromEntries(
+          scopedTranslations.field.map(({ name, value }) => [name, value]),
+        ),
+      ];
+    }),
+  );
+}
+
```
ekl176, line 105: Good Idea to fetch multiple scopes in a single call. Would be nice to have a test for the new function

app/services/cms/filters.ts
```diff
@@ -1,14 +1,21 @@
 import type { StrapiLocale } from "./models/StrapiLocale";
 import type { ApiId } from "./schemas";
 
+/**
+ * See https://docs.strapi.io/dev-docs/api/rest/filters-locale-publication#filtering
+ * for all available filter operations and add them here as necessary
+ */
+export type FilterOperation = "$eq" | "$in";
+
```
frederike-ramin, line 9: Love it!

app/services/cms/index.server.ts
```diff
@@ -99,22 +131,27 @@ export const strapiPageFromRequest = ({ request }: { request: Request }) =>
...
+        ],
+      },
+    ],
+  });
 
-  const errorPageEntries = (await Promise.allSettled(errorPagePromises))
-    .filter(
-      (promise): promise is PromiseFulfilledResult<StrapiPage> =>
-        promise.status === "fulfilled",
-    )
```
frederike-ramin, line 114: Thanks for cleaning this up!

frederike-ramin (approved):
Looks like a great addition to the code for me! I assume that you clicked through a few pages to make sure the new functions don't break the pages? (just pretty hard to anticipate what this change might cause and especially how well-tested that part of the code is 🙈

PR #1582, opened 2025-01-07
lodash: use specific functions import from submodules

m0dh4x:
refactoring for using specific imports of submodules instead of entire lodash lib every time


pgurusinga:
@m0dh4x Can we compare the bundle sizes before and after to understand the gain? Perhaps a screenshot of the results from executing `npm run build` would help.

m0dh4x:
> @m0dh4x Can we compare the bundle sizes before and after to understand the gain? Perhaps a screenshot of the results from executing `npm run build` would help.

@pgurusinga in our case it does not make a difference on the output. both 13mb. But from my perspective it is a best practice thing and more readable. what do you think?

chohner (approved):
lgtm :)

pgurusinga (approved):

PR #1583, opened 2025-01-07
fix(Image): add fallback for alt text for SVGs

aaschlote:


app/components/Image.tsx
```diff
@@ -15,10 +15,9 @@ function Image({ url, width, height, alternativeText, ...props }: ImageProps) {
   return isSvg ? (
     <SVG
       {...props}
-      id="svg-image"
       src={url}
       width={width}
-      title={alternativeText}
+      title={alternativeText ?? "image"} //force to have an alternative text and avoid a11y errors
```
pgurusinga, line 20: @aaschlote should this be in German instead?

aaschlote: it could be, but the alternative text should come from Strapi, here it's just a fallback when it's missing!

PR #1586, opened 2025-01-08
refactor(reduce e2e tests against deployed app)

Spencer6497:
This PR aims to save time on CI deployments by reducing the number of tests we run against the live, deployed app.

- **add new playwright project, staging, and adapt CI pipeline to run these tests after deployment**
- **tweak ci-pipeline.yml to run preview tests**


Spencer6497:
1m17s opposed to ~9.5min 🤯

.github/workflows/e2e-test.yml
```diff
@@ -58,7 +59,7 @@ jobs:
       - name: Run Playwright e2e tests
         env:
           GERICHTSFINDER_ENCRYPTION_KEY: "${{ secrets.GERICHTSFINDER_ENCRYPTION_KEY }}"
-          E2E_BASE_URL: "${{ inputs.e2e-target == 'preview' && 'https://a2j-test.dev.ds4g.net/' || 'http://localhost:3000/' }}"
```
chohner, line 61: to avoid the duplication, could we not use a similar inline statement to run `--project=staging` ?

Spencer6497: totally, and that was my first approach, before I realized that all the sharding/matrix strategy stuff didn't necessarily apply to these tests and added unnecessary overhead. i'm more or less indifferent though, so if it seems unnecessarily duplicated I can rework it

playwright.config.ts
```diff
@@ -55,11 +55,11 @@ export default defineConfig({
   /* Configure projects for major browsers */
   projects: [
     {
-      name: "chromium",
+      name: "local-chromium",
       use: { ...devices["Desktop Chrome"] },
     },
     {
-      name: "Mobile Android",
+      name: "local-mobile",
```
chohner, line 62: could we maybe call them something like `full`?

playwright.config.ts
```diff
@@ -68,6 +68,16 @@ export default defineConfig({
         "**/security.spec.ts",
       ],
     },
+    {
+      name: "staging",
```
chohner, line 72: again - maybe call them `reduced`? or `crucial`? 🤔

chohner: actually - maybe we could make this a [dependency](https://playwright.dev/docs/test-projects#dependencies) of the other two projects, which then always gets run. and we'd only have to specified the `critical` project when running against `preview`?

playwright.config.ts
pgurusinga: quick question: are we actually using Android here, or is this just a test with a mobile viewport? (I was on holiday and missed this info)

PR #1589, opened 2025-01-09
Fix & adapt linting rules

chohner:
This mainly enables two stylistic rules for consistency:

## "@typescript-eslint/consistent-type-definitions": ["warn", "type"]
Prefer `type` over `interface`, which we 95% already use

note: fixable by eslint so easy to switch to `interface` if we want!

## "@typescript-eslint/array-type": ["warn", { default: "array-simple" }]
prefer `[]` notation for simple types (like `string[]`), and `Array<>` for complex ones (like `Array<string | undefined>`

Note: fixable by eslint too


eslint.config.js
```diff
@@ -133,29 +129,20 @@ export default tseslint.config(
...
       // typescript-eslint
-      "@typescript-eslint/no-explicit-any": "warn",
-      "@typescript-eslint/consistent-type-imports": "warn",
-      "@typescript-eslint/only-throw-error": "off", // TODO: determine whether or not we should enable this
+      "@typescript-eslint/only-throw-error": "off", // disabled, as remix/react-router can throw redirects
       "@typescript-eslint/no-unused-vars": [
         "warn",
         { argsIgnorePattern: "^_", varsIgnorePattern: "^_" },
       ],
+      "@typescript-eslint/consistent-type-definitions": ["warn", "type"],
```
aaschlote, line 137: what's the big advantage to use type instead of interface besides the `readability in the project`?

app/domains/beratungshilfe/vorabcheck/isIncomeTooHigh.ts
```diff
@@ -10,7 +10,9 @@ export const isIncomeTooHigh: GenericGuard<BeratungshilfeVorabcheckContext> = ({
   const miete = moneyToCents(context.miete) ?? 0;
   const zahlungen = moneyToCents(context.weitereZahlungenSumme) ?? 0;
   const unterhalt =
-    (context.unterhalt == "yes" && moneyToCents(context.unterhaltSumme)) || 0;
+    context.unterhalt == "yes"
```
aaschlote, line 13: what this change is related the PR?

eslint.config.js
chohner: for now its just consistency. if we want to switch to interface (which seems to be slightly preferred) we can now easily do it..

app/domains/beratungshilfe/vorabcheck/isIncomeTooHigh.ts
chohner: yes, I believe `prefer-optional-chain`

eslint.config.js
m0dh4x: from my perspective `types` are more flexible (unions, primitives) and `types` are immutable (better readability and avoid mistakes)

PR #1590, opened 2025-01-09
Add pdf table info annuellierung

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1207867714962548


app/domains/fluggastrechte/services/pdf/sections/reason/factsOfCases/table/getConnectionDetails.ts
```diff
@@ -1,10 +1,106 @@
...
+  const replacementFlightDescription =
+    getReplacementFlightDescription(userData);
+
+  return {
+    info: `${announcementDescription}\n${replacementFlightDescription}`,
+    timeTable: ["--", "--", "--"], // to be added in following tickets
+  };
+};
+
+function getConnectionDetailsDelayOrNotMoved(
```
pgurusinga, line 101: what case is "not moved"?

pgurusinga: I think in this context the name should be `Denied Boarding` rather than `not moved`, what do you think? Or we could go with “Nicht-Befoerderung,” which might make it a bit clearer for us

aaschlote: I would like to use some english term, so `No boarding`, would be fine for you?

pgurusinga: yes, seems reasonable

pgurusinga (approved):

PR #1591, opened 2025-01-09
feat: same airports will show error(s)

pgurusinga:
## Problem
User can select same airports on 2 react-select component

## Solution
- Add validation and throw error if the select value are same


aaschlote (approved):
LGTM

PR #1592, opened 2025-01-10
ZOV feat: change anrede text field to a dropdown

m0dh4x:
we changed the current textfield for "Anrede" to a dropdown with the following options

- keine Angabe -> results in display empty string
- Herr
- Frau

### DIFF
--- a/app/domains/fluggastrechte/services/pdf/sections/__test__/getFullPlaintiffName.test.ts
+++ b/app/domains/fluggastrechte/services/pdf/sections/__test__/getFullPlaintiffName.test.ts
@@ -20,8 +20,14 @@ describe("getFullPlaintiffName", () => {
   });
 
   it("should return the full plaintiff name given the anrede, vorname and nachname", () => {
-    const actual = getFullPlaintiffName("Herr", undefined, "Test", "Test");
+    const actual = getFullPlaintiffName("herr", undefined, "Test", "Test");
 
     expect(actual).toEqual("Herr Test Test");
+  });
+
+  it("should return the full plaintiff name without anrede if anrede is none, vorname and nachname", () => {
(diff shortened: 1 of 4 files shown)


app/domains/geldEinklagen/formular/context.ts
```diff
@@ -21,7 +21,7 @@ import {
 } from "~/services/validation/YesNoAnswer";
 
 const persoenlicheDaten = {
-  ...omit(sharedPersoenlicheDaten, ["anrede"]),
+  ...omit(sharedPersoenlicheDaten),
```
chohner, line 24: you could just remove the `omit` call here

app/domains/beratungshilfe/formular/persoenlicheDaten/context.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from "~/domains/shared/formular/persoenlicheDaten/context";
 
 export const beratungshilfePersoenlicheDaten = {
-  ...omit(persoenlicheDaten, ["anrede", "title"]),
+  ...omit(persoenlicheDaten, ["title"]),
```
chohner, line 9: actually - I think we also don't use `title` @frederike-ramin ?

chohner (approved):
lgtm?

app/domains/geldEinklagen/formular/context.ts
m0dh4x: ✅

app/domains/beratungshilfe/formular/persoenlicheDaten/context.ts
m0dh4x: notice, if you dont use it than i will remove it as well :)

PR #1603, opened 2025-01-13
feat(docs): FGR - add ADR for temporary consent storage

m0dh4x:


doc/adr/0015-fgr-storing-consent-in-object-storage.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Relational Databases (RDB)
+- NoSQL Databases
+- Object Storage
+
+## Decision
+
+We will use an Object Storage solution to store the temporary consent data.
+
+### Technical Factors
```
chohner, line 28: Does this service need to allow for deletion of specific user data? was this considered?

aaschlote (approved):
LGTM

pgurusinga (approved):

doc/adr/0015-fgr-storing-consent-in-object-storage.md
```diff
@@ -0,0 +1,63 @@
...
+## Status
+
+Accepted
+
+## Context
+
+As part of our compliance with German legal and GDPR requirements ("Anscheinbeweis"), we need
+to temporarily store consent data until the "Erprobungsgesetz" is completed. The data that needs to be stored is minimal:
+
+- Random identifier
```
frederike-ramin, line 14: "random identifier"? I suppose it's rather the session identifier?

doc/adr/0015-fgr-storing-consent-in-object-storage.md
chohner: What if there is a deletion request? (sorry to be nitpicky here)

doc/adr/0015-fgr-storing-consent-in-object-storage.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Negative
+
+- Limited query capabilities if requirements change
+
+## Notes
+
+- Data format and structure specification not yet defined
+- Monitor usage patterns and costs
+- Document cleanup/decommissioning process when "Erprobungsgesetz" is completed
```
frederike-ramin, line 63: Nice

judithmh (approved):

doc/adr/0015-fgr-storing-consent-in-object-storage.md
```diff
@@ -0,0 +1,63 @@
...
+- Implementation effort is minimal compared to setting up and maintaining a database
+
+### Business Factors
+
+- Temporary solution until "Erprobungsgesetz" is completed, bucket can be deleted 365 days after the last saved consent at the latest
+- Cost-effective: approximately 1/3 of the cost compared to database solutions based on 1000 users per month
+- No long-term commitment required
+- Minimal operational overhead
+
+### Risks
```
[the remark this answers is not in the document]
frederike-ramin: - we need a mechanism to temporarily store the data for a day
- we need a cron job (or similar) to save the file once a day

doc/adr/0015-fgr-storing-consent-in-object-storage.md
m0dh4x: we are pretty sure, that there will be no deletion request

doc/adr/0015-fgr-storing-consent-in-object-storage.md
m0dh4x: solved and changed to uniqe identifier

doc/adr/0015-fgr-storing-consent-in-object-storage.md
```diff
@@ -0,0 +1,63 @@
+# 15. FGR - Storing consent in OTC object storage (OBS)
```
aaschlote, line 1: reminder, check the number of the ADR before to merge it!

doc/adr/0015-fgr-storing-consent-in-object-storage.md
```diff
@@ -0,0 +1,63 @@
...
+- Temporary solution until "Erprobungsgesetz" is completed, bucket can be deleted 365 days after the last saved consent at the latest
+- Cost-effective: approximately 1/3 of the cost compared to database solutions based on 1000 users per month
+- No long-term commitment required
+- Minimal operational overhead
+
+### Risks
+
+- Limited query capabilities (not relevant for our use case)
+
+## Consequences
```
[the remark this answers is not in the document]
m0dh4x: Hello, we can implement a integration test using mocked S3 client.

doc/adr/0015-fgr-storing-consent-in-object-storage.md
m0dh4x: hmm we have actually two "14". Think its a good idea to change it and follow the natural order @aaschlote

doc/adr/0015-fgr-storing-consent-in-object-storage.md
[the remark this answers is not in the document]
frederike-ramin: The PR is already merged, so no reason for me to write this, but I'd still like to write down my counterarguments:
- This is explicitly a temporary solution, not a long-term system
  - The creating of the data might be temporary, the actual storing is not temporary though (I wouldn't consider 1 year to be temporary), so the potential migration concerns are still valid and belong as arguments in this ADR, I think. Not saying that it should impact you decision to a point that you should flip it to database (not my place) but I still think they are valid arguments.
- The implementation (the simplest form is 1 file per user and is a simple putObject) and maintenance effort is minimal compared to maintain a database (we actually don't use)
  - Thank you for changing the "once per day" mechanism in the description. That indeed gets rid of the second half of my concerns.
- The cost is approximately 1/3 compared to database solutions
  - completely fine, but no argument against including the above proposed arguments in the ADR.
- When we implement the final consent management system later, we can easily decommission this temporary solution
  - True. But either we have two places where consent data is stored then or would have to migrate the "old" solution into the new one. thus, the first two of my arguments.

Just to make sure you understand me correctly: I understand why and that you chose S3 buckets. I don't have to agree with it but I understand it. I just want this ADR to reflect all considerations, not only the ones supporting the chosen solution. This is what ADRs are for - that we can go "back in time" and see why we decided something some time ago and if we just didn't consider another solution or if we were aware of the drawbacks.

doc/adr/0015-fgr-storing-consent-in-object-storage.md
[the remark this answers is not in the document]
frederike-ramin: Will you include that in the Consequences then?

PR #1611, opened 2025-01-15
adr(Data Rendering Pipeline)

Spencer6497:
This PR aims to outline our data rendering pipeline broadly, addressing inefficiencies and next steps for solving them.

- **copy ADR from old branch**
- **adr: Data Rendering Pipeline**


doc/adr/0015-data-rendering-pipeline.md
```diff
@@ -0,0 +1,55 @@
...
+
+  ```
+  <Heading {...getHeadingProps(strapiContent)} />
+  ```
+
+## Consequences
+
+- Server-side html rendering and sanitization will both improve performance and reduce the security concerns of handling these operations client-side.
+
+  - This also means that we'll have to use a lot of `dangerouslySetInnerHTML`
```
pgurusinga, line 46: Isn't the name `dangerouslySetInnerHTML` kind of dangerous [(link)](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html)? :D I mean, why did you write `reduce the security concerns of handling these operations client-side` in the previous line?

doc/adr/0015-data-rendering-pipeline.md
```diff
@@ -0,0 +1,55 @@
...
+      1. Translating the Strapi content into component props takes place in the compatibility function, `getXXXProps()`, in this case, `getHeaderProps()`
+      2. The actual Strapi content emitted is markdown, which we then convert to html and sanitize inside of `<RichText />`
+   2. `formElements` follows a similar pathway to the steps outlined above, but it gets rendered by `<ValidatedFlowForm />`, which builds and handles form field validations
+      1. `<StrapiFormComponents />` is akin to `<PageContent />` (see above), in that it takes a strapi component, like `form-elements.time-input` and maps it to `<StrapiTimeInput />`
+
+Our current data rendering pipeline from Strapi to the Application has some inefficiencies:
+
+- As explained in (2)(i)(b) above, we're both rendering the html and sanitizing it to remove potentially malicious content on the _frontend_. This is completely unnecessary, bad for performance and potentially bad for security. We should be rendering and sanitizing the markdown on the server-side.
```
pgurusinga, line 23: How did you come to the conclusion that this is done on the front end? I’m curious to know! I thought Remix is primarily server-side rendering and only adds client code for interactivity.

aaschlote: I believe what @Spencer6497 means it's after the hydration, the sanitizing is called again on the client due some change on the parent component, making the component re-rendering again. So the sanitizing is called twice, either in the backend and frontend.

My questions are: 

1. How bad is the perfomance today? 
2. How can we measure before and after the changes? 

Because the sanitizing is not a bad performance consumer process.

Geek note: In the React 19 we will have a feature called [React compiler](https://react.dev/learn/react-compiler), so basically it can avoid the component re-rendering on the client and not calling the sanitizing again. But until we have react 19 and enable this feature it might take a time ☹️

doc/adr/0015-data-rendering-pipeline.md
Spencer6497: it's only dangerous if you're using untrusted or unsanitized data, which we would be using neither (coming directly from our CMS, and pre-sanitized on the server-side, so `dangerouslySetInnerHTML` only receives explicitly safe content)

doc/adr/0015-data-rendering-pipeline.md
Spencer6497: @pgurusinga Sorry, I think there's a misunderstanding; by render I mean "parse markdown into html", handled by [this line](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/00b64aa27756d9521215119a25e878f987eb67e2/app/components/RichText.tsx#L44). We're also [sanitizing](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/00b64aa27756d9521215119a25e878f987eb67e2/app/components/RichText.tsx#L52) shortly after parsing, in the same `<RichText>` component.

Spencer6497: @aaschlote so as a little proof-of-concept, I moved parsing + sanitizing in _just_ the index page from the frontend to the backend (`loader`). 
* Before: ~6.487 ms (all sections of the page being parsed + sanitized in the frontend, separately)
* After: ~0.40ms
roughly 16x faster. as for measuring the changes before and after, maybe we could use the DevTools performance toolbar somehow? honestly not sure

chohner: not only faster in runtime but also smaller bundle size, as we dont need to ship the sanatizing and markdown library to the client

PR #1620, opened 2025-01-16
test(pkh): add missing flow tests

chohner:


tests/integration/util.ts
```diff
@@ -13,12 +13,7 @@ export type AllStrapiData = Record<FlowId, StrapiPages>;
 /**
  * These pages do not exist in strapi and instead exist in our own codebase
  */
-export const ignoreList = [
-  "redirect-vorabcheck-ergebnis",
-  "partner-start",
-  "abschluss",
-  "unterhalt", // antragstellende person unterhalt
-];
```
Spencer6497, line 21: i think we still need to ignore these pages? do the integration tests all pass?

app/domains/prozesskostenhilfe/formular/finanzielleAngaben/einkuenfte/xStateConfig.ts
```diff
@@ -633,6 +591,32 @@ export const getProzesskostenhilfeEinkuenfteSubflow = (
           },
         },
       },
+      ...(isPartnerFlow && {
+        [stepIds.unterhaltFrage]: {
```
Spencer6497, line 595: i know some places in the codebase we're consuming the flow structure in a static way (like `flows[flowId].config`); do these conditional branches have any affect?

Spencer6497 (approved):
great work cleaning this up 👍

tests/integration/util.ts
chohner: Yes they pass. I could remove `partner-start` and `unterhalt` because they are now being carved out of the xstate config, rather than just avoided with a guard

app/domains/prozesskostenhilfe/formular/finanzielleAngaben/einkuenfte/xStateConfig.ts
chohner: `getProzesskostenhilfeEinkuenfteSubflow()` produces two static configs: one with and one without prefix. Their structure differs slightly but I would say they are actually reduced in complexity because the step order is not decided at statemachine runtime (the partner-specific states simply don't exist in the other)

PR #1621, opened 2025-01-17
refactor(homepage): get logo by div parent parent

aaschlote:
### Problem
The current test to check if the logo of BMJ was available on the homepage was not working with the command `.getByRole("img")` anymore.

### Solution
Get the logo by the parent parent using the selector Id `#BMJ-Logo-Banner`

### DIFF
--- a/tests/e2e/critical/homepage.spec.ts
+++ b/tests/e2e/critical/homepage.spec.ts
@@ -11,7 +11,7 @@ test.beforeEach(async ({ page }) => {
 test.describe("homepage", () => {
   testPageToBeAccessible();
 
-  test("BMF logo is displayed", async ({ page }) => {
+  test("BMJ logo is displayed", async ({ page }) => {
     const logoImage = page.locator(LOGO_DIV_ID).locator("div").locator("img");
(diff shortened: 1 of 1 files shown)


tests/e2e/critical/homepage.spec.ts
```diff
@@ -2,17 +2,18 @@ import { test, expect } from "@playwright/test";
...
 });
 
 test.describe("homepage", () => {
   testPageToBeAccessible();
 
-  test.skip("BMF logo is displayed", async ({ page }) => {
-    const allImages = await page.getByRole("img").all();
-    expect(allImages.length).toBeGreaterThan(0);
-    await expect(allImages[0]).toBeVisible();
+  test("BMF logo is displayed", async ({ page }) => {
```
Spencer6497, line 14: not sure why it was spelled `BMF` previously but we can probably change to `BMJ` :)

Spencer6497 (approved):
🏆

tests/e2e/critical/homepage.spec.ts
aaschlote: nice catch, fixed!

PR #1627, opened 2025-01-17
ZOV: Add airlines address in overview page

pgurusinga:


app/domains/fluggastrechte/components/useAirlineDetails.tsx
```diff
@@ -0,0 +1,44 @@
...
+        const response = await fetch(
+          `${AIRLINE_DETAILS_PATH}${airlineByIataCode}`,
+        );
+
+        if (response.ok) {
+          const data = await response.json();
+          const { name } = AirlineDetailsSchema.parse(data);
+          setAirlineDetails({ name });
+        } else {
+          setAirlineDetails({ name: "" });
```
aaschlote, line 31: I believe this else condition is not needed, the state is already initialised as empty, right?

app/routes/__test__/api.airline.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    };
+
+    vi.mocked(getAirlineNameByIataCode).mockReturnValue("Lufthansa");
+
+    const response = loader(mockArgs);
+
+    expect(response).toBeInstanceOf(Response);
+    expect(response.status).toBe(200);
+
+    return response.text().then((text) => {
```
aaschlote, line 28: what about?

```
const data = await response.json();
expect(data).toEqual({ name: "Lufthansa" });
```

Remember to make the test async

app/routes/__test__/api.airline.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { loader } from "~/routes/api.airline.$iataCode";
+
+vi.mock(
+  "~/domains/fluggastrechte/services/airlines/getAirlineNameByIataCode",
+  () => ({
+    getAirlineNameByIataCode: vi.fn(),
+  }),
+);
+
+describe("iataCode loader", () => {
```
aaschlote, line 13:
```suggestion
describe("api.airline loader", () => {
```

app/routes/api.airline.$iataCode.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getAirlineNameByIataCode } from "~/domains/fluggastrechte/services/airlines/getAirlineNameByIataCode";
+import { defaultHeaders } from "~/rootHeaders";
+
+export function loader({ params }: LoaderFunctionArgs) {
+  const { iataCode } = params;
+  const name = getAirlineNameByIataCode(iataCode);
+
+  const responseData = { name };
+
+  return new Response(JSON.stringify(responseData), {
```
aaschlote, line 11: If you do like this, it might work as well

```
return { name };
```

You would need to change your unit test for that.

aaschlote (approved):
LGTM besides my comments

app/domains/fluggastrechte/components/AirlineDetailsOverviewCard.tsx
```diff
@@ -0,0 +1,48 @@
+import EditButton from "@digitalservicebund/icons/CreateOutlined";
+import Button from "~/components/Button";
+import Heading from "~/components/Heading";
+
+type AirlineAddress = {
+  name?: string;
+  adresse?: string;
```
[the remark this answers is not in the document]
aaschlote: never mind, it's in german, I forgot it!

app/domains/fluggastrechte/components/useAirlineDetails.tsx
pgurusinga: I think so too, can remove this redundant part

app/routes/api.airline.$iataCode.ts
pgurusinga: how would I add security headers in that case?

aaschlote: it should be there already, due the `root.tsx`?

pgurusinga: This is what I get when I return directly as object.
image

and this is with new Response() 
image

reference: https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html#security-headers

PR #1634, opened 2025-01-20
refactor(parse and sanitize markdown on server)

Spencer6497:
This PR shifts the responsibility of parsing and sanitizing markdown from the `<RichText/>` component to the zod schema definitions via `.transform()`. To do this, I've added a new `buildRichTextValidation` zod schema type, and replaced all instances of `zod.string()` with it, where applicable. Further cleanup PR to come.
- **add new parseAndSanitize method, move sanitize to same file**
- Create new buildRichTextValidation() zod schema method

---
  - https://app.asana.com/0/0/1208848279655976


Spencer6497:
Tests are failing because at certain points in the CMS, we've nested list items (asterisks in markdown evaluate to `<li>` tags) inside of Mustache conditional statements. Since the markdown is now getting parsed + sanitized _way_ before stringReplacements are being evaluated, we run into an issue where syntactically important DOM elements (like opening `<ul>` tags) get trapped inside `{{ #conditional }}` logic and get deleted when those conditions evaluate to false

chohner:
> Tests are failing because at certain points in the CMS, we've nested list items (asterisks in markdown evaluate to `<li>` tags) inside of Mustache conditional statements. Since the markdown is now getting parsed + sanitized _way_ before stringReplacements are being evaluated, we run into an issue where syntactically important DOM elements (like opening `<ul>` tags) get trapped inside `{{ #conditional }}` logic and get deleted when those conditions evaluate to false

Is that a sign of malformed content? I would've hoped that markdown conversion & template replacement to be commutative...

Spencer6497:
> > Tests are failing because at certain points in the CMS, we've nested list items (asterisks in markdown evaluate to `<li>` tags) inside of Mustache conditional statements. Since the markdown is now getting parsed + sanitized _way_ before stringReplacements are being evaluated, we run into an issue where syntactically important DOM elements (like opening `<ul>` tags) get trapped inside `{{ #conditional }}` logic and get deleted when those conditions evaluate to false
> 
> Nice find. Is that a sign of malformed content? I would've hoped that markdown conversion & template replacement to be commutative...

Nope, unfortunately the content is correct :/ It would be commutative if we were _only_ doing simply text replacements, but since we're executing IF-ELSE logic with conditionals, important HTML gets omitted after it's been parsed/sanitized from markdown

app/util/__test__/strings.test.ts
```diff
@@ -118,3 +119,25 @@ describe("stripTrailingSlashFromURL", () => {
     });
   });
 });
+
+describe("removeMarkupTags", () => {
+  it("should remove markup tags", () => {
+    expect(removeMarkupTags("<p>test</p>")).toBe("test");
```
pgurusinga, line 125: I'm wondering if I understand correctly. I ran this test locally: `expect(removeMarkupTags("<br>test")).toBe("test");`, and it passed. Does this mean we remove HTML-style self-closing tags like `<br>`?

app/components/inputs/tile/TileRadio.tsx
```diff
@@ -1,15 +1,8 @@
-import type { Renderer } from "marked";
 import { useField } from "remix-validated-form";
 import TileTag, { type TileDescriptionProps } from "./TileTag";
 import Image, { type ImageProps } from "../../Image";
 import RichText from "../../RichText";
 
-const paragraphRenderer: Partial<Renderer> = {
-  paragraph({ text }) {
-    return `<p class="ds-subhead">${text}</p>`;
```
aaschlote, line 9: can you check if this component renders the `RichText` with the paragraph with this className?

app/util/__test__/strings.test.ts
Spencer6497: yes. this helper function should remove _all_ markup tags, for the purpose of checking whether or not a string is _truly_ empty. with the way we're first parsing then _later_ injecting template strings/conditionals, it can happen that we end up with a string like `<p></p>\n`, and this function removes all of that to see if it's empty

app/components/inputs/tile/TileRadio.tsx
Spencer6497: I added a new test suite, `richtest.test.ts`, as the rendering is occurring inside of `buildRichTextValidation`

app/util/__test__/strings.test.ts
pgurusinga: To recap, if we use a single line break tag in the rich text, will it be removed? Would that cause any content issues?

Spencer6497: no; this function isn't used to actually parse/modify the markdown or html, it's only used to test whether or not an already-parsed html string is empty or not, after removing tags. Furthermore it's currently only used in the List component and in the InlineNotice, two components in which we're conditionally rendering based on stringReplacement conditions 😵‍💫 confusing!

pgurusinga: makes sense. thanks for the explanation!

PR #1638, opened 2025-01-21
ZOV: add new consent page

pgurusinga:


app/domains/fluggastrechte/formular/context.ts
```diff
@@ -9,6 +10,7 @@ import { fluggastrechtProzessfuehrungDaten } from "./prozessfuehrung/context";
 import { fluggastrechtStreitKostenDaten } from "./streitwertKosten/context";
 
 export const fluggastrechtContext = {
+  datenverarbeitungZustimmung: checkedRequired,
```
aaschlote, line 13:
```suggestion
  datenVerarbeitungsZustimmung: checkedRequired,
```

aaschlote (approved):
Can you please add the new context to the doneFunction as well?

LGTM!

app/domains/fluggastrechte/formular/context.ts
pgurusinga: btw Datenverarbeitung is a word, I don't think we could camel-case that

aaschlote: ahh ok, because you created a page called `daten-verarbeitung`, with dash, I thought it was two words

pgurusinga: makes sense, should we change the URL to `datenverarbeitung` ? because it was the requirement in the ticket to have dash in between

tests/e2e/domains/fluggastrechte/formular/startFluggastrechteFormular.ts
```diff
@@ -16,6 +16,9 @@ export async function startFluggastrechteFormular(
   await expectPageToBeAccessible({ page });
   await formular.clickNext();
 
+  // /fluggastrechte/formular/grundvoraussetzungen/daten-verarbeitung
+  await formular.fillCheckboxes("datenverarbeitungZustimmung");
```
pgurusinga, line 20: @aaschlote This test is supposed to fail, but it shows a green result instead. This happens because this line doesn’t click "next." are we skipping this flow test?

aaschlote: when did you create the page? We had some content deployments today

aaschlote: Fixing in this [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1640)

PR #1642, opened 2025-01-22
feat(Feedback component): Encourage more actionable positive feedback

khallad2:
- Conditional follow-up question: Based on the helpful rating, the follow-up question should be adapted. The design is also slightly adapted
- after positive rating: "Was fanden Sie besonders hilfreich?" (new)
- after negative rating: "Was können wir verbessern?" (as is)
- Add "(optional)" at the end of the follow-up question
- Add success message above follow-up question: Display success message with checkmark above the follow-up question (see design). This way users should not feel obliged to input sth here.
- For consistency, we then also need to adjust the success message on the final screen (see design)


app/components/userFeedback/index.tsx
```diff
@@ -22,6 +22,9 @@ export const USER_FEEDBACK_ID = "user-feedback-banner";
 export default function UserFeedback(props: Readonly<UserFeedbackProps>) {
   const { pathname } = useLocation();
   const [shouldFocus, setShouldFocus] = useState(false);
+  const [positiveFeedback, setPositiveFeedback] = useState<boolean | null>(
+    null,
+  );
```
Spencer6497, line 27: instead of tracking a boolean, `positiveFeedback`, I'd rather track an enum, say, `UserFeedback.Positive` or `UserFeedback.Negative`

app/components/userFeedback/index.tsx
```diff
@@ -50,14 +53,18 @@ export default function UserFeedback(props: Readonly<UserFeedbackProps>) {
                 <RatingBox
                   url={pathname}
                   heading={props.rating.heading}
-                  onSubmit={applyFocus}
+                  onSubmit={(feedback) => {
+                    setPositiveFeedback(feedback);
+                    applyFocus();
+                  }}
```
Spencer6497, line 59: i think instead of defining this in the DOM here, I'd move it into a function above like `onRatingSubmitted()`

app/components/userFeedback/RatingBox.tsx
```diff
@@ -12,42 +12,68 @@ export const userRatingFieldname = "wasHelpful";
...
-  readonly onSubmit: () => void;
+  readonly onSubmit: (positiveFeedback: boolean) => void;
 };
 
 export const RatingBox = ({ heading, url, onSubmit }: RatingBoxProps) => {
   const ratingFetcher = useFetcher();
   const [jsAvailable, setJsAvailable] = useState(false);
+  const [positiveFeedback, setPositiveFeedback] = useState<boolean | null>(
+    null,
+  );
```
Spencer6497, line 23: i don't think we actually need to be tracking this state here, as it's being handled in the parent component (`UserFeedback`)

app/components/userFeedback/RatingBox.tsx
```diff
@@ -12,42 +12,68 @@ export const userRatingFieldname = "wasHelpful";
...
   useEffect(() => setJsAvailable(true), []);
+  useEffect(() => {
+    if (positiveFeedback !== null) {
+      onSubmit(positiveFeedback);
+    }
+  }, [positiveFeedback, onSubmit]);
```
Spencer6497, line 29: same here, I don't think we need to be watching for changes to this and triggering another submission

app/components/userFeedback/RatingBox.tsx
```diff
@@ -12,42 +12,68 @@ export const userRatingFieldname = "wasHelpful";
...
         method="post"
         action={`/action/send-rating?url=${url}&js=${String(jsAvailable)}`}
         preventScrollReset={true}
-        onSubmit={onSubmit}
       >
         <ButtonContainer>
           <Button
             iconLeft={<ThumbUpIcon />}
-            look="tertiary"
+            look={positiveFeedback === true ? "primary" : "tertiary"}
```
Spencer6497, line 57: why are we changing the look of the button? plus, doesn't this component disappear when one of these are clicked?

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -66,11 +68,23 @@ export const FeedbackFormBox = ({
       preventScrollReset={true}
       onSubmit={onSubmit}
     >
+      <div className="flex items-center text-base font-bold mb-[1em]">
+        <CheckCircleIcon className="w-[2em] h-[1.5em] mr-[0.25em] text-green-600 " />
+        <p>{feedbackTranslations["success-message"]}</p>
```
Spencer6497, line 73: in the designs, it looks like "Vielen Dank" is bold, but "Ihre Antwort wurde übermittelt" is not; we should probably split these into two translation strings and style them differently

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -66,11 +68,23 @@ export const FeedbackFormBox = ({
...
+              {feedbackTranslations["positive-feedback-question"]}
+            </label>
+          )}
+          {!positiveFeedback && (
+            <label htmlFor={FEEDBACK_FIELD_NAME} className="ds-label-01-bold">
+              {feedbackTranslations["negative-feedback-question"]}
+            </label>
+          )}
           <p className="ds-text-02-reg text-gray-800">
             {feedbackTranslations["heading-personal-data-feedback"]}
```
Spencer6497, line 89: i believe this should actually be `feedbackTranslations['placeholder-feedback']`, reading "Bitte tragen Sie keine persönliche Daten ein...." and the new placeholder for the Textarea should read "Ihre Antwort" (defined in a new translation string probably)

app/components/userFeedback/FeedbackFormBox.tsx
Spencer6497, line 1: couldn't leave a comment inline but I think we're getting rid of the Abschicken icon? at least from the designs it looks that way, but I would clarify

app/components/userFeedback/PostSubmissionBox.tsx
```diff
@@ -19,14 +20,17 @@ export const PostSubmissionBox = ({ shouldFocus }: Props) => {
...
-        role="status"
-        tabIndex={-1}
-        innerRef={headingReference}
-      />
+      <div className="flex items-center text-base font-bold mb-[1em]">
+        <CheckCircleIcon className="w-[2em] h-[1.5em] mr-[0.25em] text-green-600 " />
+        <Heading
+          look="ds-label-01-bold"
+          tagName="h2"
+          text={feedbackTranslations["heading-post-submission"]}
```
Spencer6497, line 28: same thing mentioned before but I think the bolded "Vielen Dank" needs to be separate from "Ihr Feedback hilft uns...."

Spencer6497 (changes requested):
Great work on this! 🚀  Left a few comments that I think could simplify our implementation :)

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -111,3 +109,23 @@ export const FeedbackFormBox = ({
     </ValidatedForm>
   );
 };
+
+export const FeedbackTitle = ({
```
aaschlote, line 113: Can we move this component to its own file? From what I’ve observed, this component is also used in other components and already has its own test file.

app/components/userFeedback/feedbackTranslations.ts
```diff
@@ -3,6 +3,9 @@ import { useTranslations } from "~/services/translations/translationsContext";
 
 const feedbackTranslationsKeys = [
   "heading-feedback",
+  "success-message",
+  "positive-feedback-question",
+  "negative-feedback-question",
```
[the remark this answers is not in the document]
khallad2: of course, i would add them before deployment, when the PR is approved.
thanks for adding them.

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -37,15 +38,24 @@ export type FeedbackBoxProps = {
   readonly destination: string;
   readonly shouldFocus: boolean;
   readonly onSubmit: () => void;
+  readonly feedback?: FeedbackType;
+};
+
+const getWasHelpfulFromUrl = (url: string) => {
+  const params = new URLSearchParams(url);
+  const value = params.get("wasHelpful");
+  return value === "yes" || value === "no" ? value : null;
```
Spencer6497, line 47: if we're just testing for _any_ non-falsy value, could we do something like 
```suggestion
  return value ?? null;
```

app/routes/__test__/action.send-rating.test.ts
```diff
@@ -31,14 +31,18 @@ describe("/action/send-rating route", () => {
...
       `http://localhost:3000/action/send-rating?url=${ratingPath}&js=false`,
       options,
     );
     const response = await action({ request, params: {}, context: {} });
     expect(response.status).toEqual(302);
-    expect(response.headers.get("location")).toEqual(ratingPath);
+    const location =
+      new URL(response.headers.get("location") ?? "").pathname +
+      new URL(response.headers.get("location") ?? "").search +
+      new URL(response.headers.get("location") ?? "").hash;
```
Spencer6497, line 44: instead of building 3 URL objects, could we destructure one, like so?:
```suggestion
    const {pathname, search, hash} = new URL(response.headers.get("location") ?? "")
```

app/routes/action.send-rating.ts
```diff
@@ -54,9 +54,13 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
+
   const clientJavaScriptAvailable = searchParams.get("js") === "true";
+  if (clientJavaScriptAvailable) {
+    return json({ success: true }, { headers });
+  }
 
-  return clientJavaScriptAvailable
-    ? json({ success: true }, { headers })
-    : redirect(`${url}#${USER_FEEDBACK_ID}`, { headers });
+  return redirect(`${updatedUrl.toString()}#${USER_FEEDBACK_ID}`, { headers });
```
Spencer6497, line 65: why did we split up the ternary?

Spencer6497 (approved):
excellent work! so glad you were able to get it working. left a few non-blocking comments :)

app/routes/action.send-rating.ts
khallad2: i thought it just might be more readable. should i revert it ?

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -37,15 +38,24 @@ export type FeedbackBoxProps = {
   readonly destination: string;
   readonly shouldFocus: boolean;
   readonly onSubmit: () => void;
+  readonly feedback?: FeedbackType;
+};
+
+const getWasHelpfulFromUrl = (url: string) => {
+  const params = new URLSearchParams(url);
+  const value = params.get("wasHelpful");
+  return value ?? null;
```
chohner, line 47: I think this could just be `(new URLSearchParams(url)).get("wasHelpful")`? I don't really see the reason to return `null` here

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -57,6 +67,18 @@ export const FeedbackFormBox = ({
     }
   }, [shouldFocus]);
 
+  if (!feedback) {
+    const wasHelpful = getWasHelpfulFromUrl(location.search);
+    if (wasHelpful === "yes") feedback = FeedbackType.Positive;
+    else if (wasHelpful === "no") feedback = FeedbackType.Negative;
```
chohner, line 73: since we already have a function that looks at the URL, couldn't we do the parsing there? I feel something like this might be more clear

```typescript
const feedbackValue = feedback ?? getFeedbackFromUrl(location.search)

if (!feedbackValue) return null;
```

app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -57,6 +67,18 @@ export const FeedbackFormBox = ({
...
+    else return null;
+  }
+
+  const feedbackText = {
+    [FeedbackType.Positive]: feedbackTranslations["positive-feedback-question"],
+    [FeedbackType.Negative]: feedbackTranslations["negative-feedback-question"],
+  }[feedback];
```
chohner, line 80: you could define the FeedbackType -> lookup string map statically:

```typescript
const feedbackTypeMap = {
FeedbackType.Positive: "positive-feedback-question"
FeedbackType.Negative: "negative-feedback-question"
}
....
feedbackTranslations[feedbackTypeMap[feedbackValue]]
```


or since its binary:
```typescript
const feedbackTextKey = feedbackValue === FeedbackType.Positive ? "positive-feedback-question" : "negative-feedback-question"
feedbackTranslations[feedbackTextKey]
```

chohner (commented):
just some nitpicks

PR #1644, opened 2025-01-22
Implement user storage consent

aaschlote:


app/services/flow/server/flowTransitionValidation.ts
```diff
@@ -20,6 +20,24 @@ export function getFlowTransitionConfig(currentFlow: Flow) {
     : undefined;
 }
 
+function getAsyncFlowActions(currentFlow: Flow) {
```
pgurusinga, line 23: Hey, I just noticed that this function is for actions specific to a step, not the whole flow. What do you think about renaming `getAsyncFlowActions` to `getAsyncStepActions`? It feels a bit odd to have a step-specific function mixed in with the flow stuff.
if you agree then we can have a `stepTransitionValidation` file?

app/services/s3/__test__/storeConsentFgrToS3Bucket.test.ts
```diff
@@ -0,0 +1,96 @@
...
+});
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  vi.useFakeTimers();
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+});
```
pgurusinga, line 45: should this be scoped inside first `describe()` within the module? would this affect other tests since it's scoped in globally?

app/services/s3/storeConsentFgrToS3Bucket.ts
```diff
@@ -0,0 +1,51 @@
+import { PutObjectCommand } from "@aws-sdk/client-s3";
```
pgurusinga, line 1: How about renaming the `s3` directory to `externalDataStorage`? I think it sounds more generic and clear. We can keep the function name with `s3` since it still refers to storing data in that service. What do you think?

app/services/s3/storeConsentFgrToS3Bucket.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const getFolderDate = () => {
+  return today()
+    .toLocaleDateString("de-DE", {
+      day: "2-digit",
+      month: "2-digit",
+      year: "numeric",
+    })
+    .replaceAll(".", "-");
+};
```
pgurusinga, line 21: can we use existing Date util?

```ts
const getFolderDate = () => {
  return toGermanDateFormat(today()).replaceAll(".", "-");
};
```

app/services/s3/storeConsentFgrToS3Bucket.ts
```diff
@@ -0,0 +1,51 @@
+import { PutObjectCommand } from "@aws-sdk/client-s3";
+import { config } from "~/services/env/env.server";
+import { today } from "~/util/date";
+import { createClientDataConsentFgr } from "./createClientDataConsentFgr";
+import { sendSentryMessage } from "../logging";
+import { getSessionIdByFlowId } from "../session.server";
+
+const createConsentDataBuffer = (sessionId: string, request: Request) => {
```
pgurusinga, line 8: this is just my thought; sometimes, we pass the whole request object, which can be unnecessary. but i am fine as is

```suggestion
const createConsentDataBuffer = (sessionId: string, { headers }: { headers: Headers }) => {
```

pgurusinga (commented):
I am unsure about the async action naming. what do you think?

app/services/env/env.server.ts
```diff
@@ -15,6 +15,11 @@ type Config = {
   SAML_SP_METADATA_PATH: string;
   SAML_SP_SECRET_KEY_PATH: string;
   SAML_IDP_CERT?: string;
+  AWS_S3_REGION: string;
+  AWS_S3_ENDPOINT: string;
+  AWS_S3_DATA_CONSENT_FGR_ACCESS_KEY: string;
+  AWS_S3_DATA_CONSENT_FGR_SECRET_KEY: string;
+  AWS_S3_DATA_CONSENT_FGR_BUCKET_NAME: string;
```
chohner, line 22: maybe we should drop the `AWS_` prefix here (and everywhere throughout)? 😅

app/services/s3/createClientDataConsentFgr.ts
```diff
@@ -0,0 +1,21 @@
+import { S3Client } from "@aws-sdk/client-s3";
+import { config } from "~/services/env/env.server";
+
+let instance: S3Client | undefined = undefined;
+
+export const createClientDataConsentFgr = () => {
```
chohner, line 6: I would (strongly) vote for having a shared storage service, that we can then scope by app service (consent, file-upload, ...).
So this would be a generic s3 instance

app/services/s3/storeConsentFgrToS3Bucket.ts
```diff
@@ -0,0 +1,51 @@
...
+  return today()
+    .toLocaleDateString("de-DE", {
+      day: "2-digit",
+      month: "2-digit",
+      year: "numeric",
+    })
+    .replaceAll(".", "-");
+};
+
+export const storeConsentFgrToS3Bucket = async (request: Request) => {
```
chohner, line 23: if we use a shared storage service, this could be moved into its own `consent` service and act as the main interface to the application

app/services/s3/storeConsentFgrToS3Bucket.ts
chohner: same idea, but why not simply `sessionId: string , headers: Headers`?

app/services/s3/__test__/storeConsentFgrToS3Bucket.test.ts
aaschlote: I prefer to have the before and after each scopes outside of the `describe` methods. But having them inside or outside it doesn't cause any problem globally, because every test file runs in a different thread.

app/services/s3/storeConsentFgrToS3Bucket.ts
aaschlote: sure

app/services/s3/storeConsentFgrToS3Bucket.ts
aaschlote: yeap, much better!

app/services/s3/createClientDataConsentFgr.ts
aaschlote: to be discussed in the Dev Exchange next week!

init-s3-bucket.py
```diff
@@ -0,0 +1,12 @@
...
+
+s3_client = boto3.client(
+    "s3",
+    endpoint_url=f"http://localhost:4566",
+    aws_access_key_id="test",
+    aws_secret_access_key="test",
+    region_name="eu-central-1"
+)
+
+s3_client.create_bucket(Bucket="a2j-rechtsantragstelle-data-consent-fgr", CreateBucketConfiguration={
```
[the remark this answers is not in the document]
aaschlote: yeap, if I change the name, I need to change the name for the local development .env as well

app/services/env/env.server.ts
pgurusinga: I think I understand why Alex did this, based on my experience with the `aws` cli. This is how AWS typically shows examples for environment variables. The CLI tools also accept environment variables in this format.

reference: https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html

app/services/s3/storeConsentFgrToS3Bucket.ts
aaschlote: for the first comment, this is file is related store something in the S3 bucket. As we already discussed, we renamed the folder to be `externalDataStorage`.

For the second comment, I can change, but in case a new flow action is added and needs another data from the request (body, formData, referrer or etc), all the functions must be changed again.

app/domains/fluggastrechte/formular/index.ts
```diff
@@ -46,6 +47,10 @@ const flowTransitionConfig: FlowTransitionConfig = {
   eligibleSourcePages: ["/ergebnis/erfolg"],
 };
 
+const asyncFlowActions = {
+  "/grundvoraussetzungen/datenverarbeitung": storeConsentFgrToS3Bucket,
+};
```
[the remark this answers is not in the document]
aaschlote: I would prefer to have here for now, but we can move to the `steps` config from this [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1339) once it's implemented 😃

app/services/flow/server/flowTransitionValidation.ts
aaschlote: actually I added in the wrong file, it should have a specific file for these functions. I'll fix it.

About the naming, it's not only for the step, because before I get the step I need to check if the flow has a configuration of actions setup. I can rename the config to `asyncFlowStepActions`. What do you think?

init-s3-bucket.py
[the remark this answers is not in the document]
aaschlote: and btw, this file is executed on the docker container and not in your local development or app

app/services/flow/server/flowTransitionValidation.ts
pgurusinga: ah ok now I got it, then I am fine keeping it and if this gets confusing, we can improve later

PR #1648, opened 2025-01-23
test(pkh flow): add forward-only flow tests

chohner:
We have a few flow steps (mostly warning or array pages) that are not reachable using the `BACK` event. therefore, they can't really be tested inside a flow-test. I propose a forward-only test to at least cover the existing transition


Spencer6497 (approved):
great idea! 🚀

app/domains/__test__/flows.test.ts
```diff
@@ -158,6 +161,31 @@ describe.sequential("state machine form flows", () => {
     },
   );
 
+  // Array pages cannot be tested above since they aren't reachable using the `BACK` transition
+  // However, we can still verify that their `SUBMIT` transition is correct
+  const forwardOnlyTests = {
```
aaschlote, line 166: should be `submitOnlyTests`?

aaschlote (approved):
LGTM

PR #1650, opened 2025-01-23
feat(file upload): Basic File Upload Infrastructure

Spencer6497:
This PR adds the skeleton for file upload handling, including exposing a "test" page accessible only in non-prod environments.

- **add basic plumbing for file uploads**
- **add file upload component**
- **add multipart form data and enable data saving with RVF**
- **temporarily adjust number of tests**


app/util/file/fileSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+export const fileSchema = (
+  typeof window === "undefined" ? z.any() : z.instanceof(File)
+)
+  .refine(
+    (file?: File) => !!file && file.name !== "" && file.size > 0,
+    "File required",
+  )
+  .refine(
+    (file?: File) =>
+      file?.type === "application/pdf" && file?.name.endsWith(".pdf"),
```
pgurusinga, line 14: why do we need file postfix`.pdf` ? I tested it on my machine by changing `.pdf` to uppercase `.PDF`, and it's still valid pdf file

app/routes/shared/formular.server.ts
```diff
@@ -230,29 +245,72 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
+      return Promise.resolve(
+        JSON.stringify({
+          etag: s3UploadResult?.ETag,
+          createdOn: new Date(),
+          filename,
+          sizeKb: file.size / 1024,
+        }),
+      );
+    },
+  );
```
pgurusinga, line 281: this means if we don't scope this to stepId, All pages will call this function right?
you could use the API `asyncFlowActions` in the flow e.g. from ZOV
```ts
const asyncFlowActions = {
  "/grundvoraussetzungen/datenverarbeitung": storeMainPersonConsent,
  "/persoenliche-daten/weitere-personen/person/daten":
    storeMultiplePersonsConsent,
};
```
to scope down the function call

pgurusinga: If you do the above, I think you won't need to make any code changes in this `formular.server.ts`.

app/routes/shared/formular.server.ts
```diff
@@ -208,7 +214,16 @@ export const loader = async ({
       navItems,
       postFormContent: cmsContent.postFormContent,
       preHeading: cmsContent.preHeading,
-      stepData,
+      // REMOVE  ME
```
pgurusinga, line 217: remove ?

app/routes/shared/formular.server.ts
pgurusinga: check it out here: https://github.com/digitalservicebund/a2j-rechtsantragstelle/commit/69058bb4bcf7cc7bd280b4938f8c281494530ad9

Spencer6497: Correct, all pages call this function, which imo is the correct behavior now that we've also accordingly changed the Form's encoding type to be [multipart](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1650/files#diff-b27d29866e111234ace6aea4fe49472f15801294ce131766bbeaf2ce1035272eR33). theoretically we _could_ move this to an `asyncFlowAction`, but then we'd have to call said flow actions way earlier in the action, and also tweak the return type/error/validation handling

app/routes/shared/formular.server.ts
```diff
@@ -230,29 +245,72 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
-      const headers = { "Set-Cookie": await commitSession(flowSession) };
-      return new Response("success", { status: 200, headers });
-    } catch (err) {
-      return new Response((err as Error).message, { status: 422 });
-    }
-  }
+  const requestCopy = request.clone();
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let validationResult: ValidationResult<any> | undefined;
+  let formData = await unstable_parseMultipartFormData(
```
aaschlote, line 251: Can we move all this logic to a separate file? I'd prefer to keep formular.server.ts from getting too long.

app/routes/shared/formular.server.ts
```diff
@@ -230,29 +245,72 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
+        return;
+      }
+      // Must convert to File
+      const dataArr = [];
+      for await (const chunk of data) {
+        dataArr.push(chunk);
+      }
+      const file = new File(dataArr, filename, { type: contentType });
+      const submittedData = { [name]: file };
+      validationResult = await validatorForFieldNames(
```
aaschlote, line 264: are we going to do the validation of the file size and type only in the front-end or can we do also in the back end?

app/routes/shared/formular.server.ts
aaschlote: and also for every page that it has the `FileUpload` component we would need to add it, I guess the `asyncFlowAction` won't be helpful here

app/routes/shared/formular.server.ts
```diff
@@ -230,29 +245,72 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
-  if (validationResult.error)
+    validationResult = await validateFormData(pathname, relevantFormData);
+  } else if (!validationResult.error) {
+    // file upload, need to de-serialize values
+    for (const [key, val] of formData.entries()) {
+      const parsedValue = JSON.parse(val as string);
+      validationResult.data[key] = parsedValue;
+    }
+  } else {
     return validationError(
```
aaschlote, line 309: when this `validationError` is going to be called if the page doesn't have a fileUpload?

app/routes/shared/formular.server.ts
```diff
@@ -230,29 +245,72 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
+          filename,
+          sizeKb: file.size / 1024,
+        }),
+      );
+    },
+  );
 
-  const validationResult = await validateFormData(pathname, relevantFormData);
+  // non-file upload case, parse formData normally
+  if (!validationResult) {
```
aaschlote, line 284: what does happen in a page it has the FileUpload and normal Input? Because it sounds the code I can have only on the page either `FileUpload` or the other components, I can't have both

app/components/inputs/FileInput.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <div>
+      <label htmlFor={name}>{label}</label>
+      <input
+        {...getInputProps({ id: name })}
+        type="file"
+        accept=".pdf"
+        aria-errormessage={error && errorId}
+        aria-invalid={error !== undefined}
+      ></input>
+      {error && <InputError id={errorId}>{error}</InputError>}
```
aaschlote, line 22: can we add a test to `ValidatedFlowForm` or are we going to replace this component later?

app/routes/shared/formular.server.ts
Spencer6497: that's what we're doing in this line; the `fileSchema` has `refine` statements that validate it

app/util/file/fileSchema.ts
Spencer6497: i dunno, i figured it wouldn't hurt; is it even possible to have type == 'application/pdf' and a different file extension?

app/components/inputs/FileInput.tsx
Spencer6497: will be replaced by [this](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1758)

app/routes/shared/formular.server.ts
Spencer6497: ah you're right, it needs to be fully outside the conditional and test against `validationResult.error`

app/routes/shared/formular.server.ts
Spencer6497: ah good point.... I think i'll move the non-file upload case to inside of the FileUploadHandler (as we're again parsing _all_ forms/fields with this function) and pull it all into its own file

app/routes/shared/formular.server.ts
pgurusinga: ah got it, because of the error validation, also can we extract this out to separate func? formular file is getting too big

app/util/file/fileSchema.ts
pgurusinga: it's up to you, it's a non-blocker, you never know what users do with their files 😆

app/routes/shared/formular.server.ts
pgurusinga: ah @aaschlote  already mentioned this :D

app/services/flow/server/parseMultipartFormData.ts
```diff
@@ -0,0 +1,110 @@
...
+    updateSession(flowSession, { [arrayName]: arrayToMutate });
+    const headers = { "Set-Cookie": await commitSession(flowSession) };
+    return new Response("success", { status: 200, headers });
+  } catch (err) {
+    return new Response((err as Error).message, { status: 422 });
+  }
+}
+
+async function dataToFile(
+  data: AsyncIterable<Uint8Array<ArrayBufferLike>>,
```
chohner, line 101: I would prefer to be more descriptive here, maybe `collectAsyncBufferLike`? Furthermore, using `File` we should mark this file as server-only (`parseMultipartFormData.server.ts`)

app/services/flow/server/parseMultipartFormData.ts
```diff
@@ -0,0 +1,110 @@
...
+  arrayIndexFromFormData,
+  arrayFromSession,
+  deleteFromArrayInplace,
+} from "~/services/session.server/arrayDeletion";
+import { validateFormData } from "~/services/validation/validateFormData.server";
+import { filterFormData } from "~/util/filterFormData";
+
+const bytesInKilobyte = 1024;
+
+export async function parseMultipartFormData(
```
chohner, line 18: I feel like there is too much happening here. Maybe splitting the parsing and validation?

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
```diff
@@ -0,0 +1,43 @@
...
+    const flowId = flowIdFromPathname(new URL(request.url).pathname);
+    if (!flowId)
+      throw new Error(`Attempted to upload user file outside of known flow`);
+    const sessionId = await getSessionIdByFlowId(flowId, cookieHeader);
+
+    return await s3Client.send(
+      new PutObjectCommand({
+        Bucket: config().S3_DATA_STORAGE_BUCKET_NAME,
+        Body: new Uint8Array(await file.arrayBuffer()),
+        Key: createFolderKey(sessionId, flowId, crypto.randomUUID()),
```
chohner, line 31: `fileHash` to me sounds like this should contain the actual hash of the file, not a `randomUUID`. If this indeed shouldn't calculate the file hash I would suggest to move the UUID generation into `createFolderKey()`

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getSessionIdByFlowId } from "~/services/session.server";
+
+const USER_FILES_FOLDER = "user-files";
+
+const createFolderKey = (
+  sessionId: string,
+  flowId: string,
+  fileHash: string,
+) => {
+  return `${USER_FILES_FOLDER}${flowId}/${sessionId}/${fileHash}`;
```
chohner, line 15: no slash between `USER_FILES_FOLDER` and `flowId`?

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export async function uploadUserFileToS3(request: Request, file: File) {
```
chohner, line 18: could we avoid passing the whole `Request` object into this function? it makes it harder to test and more intransparent what information we actually need (flow id and session id).

app/routes/shared/formular.server.ts
```diff
@@ -230,32 +236,23 @@ export const action = async ({ request }: ActionFunctionArgs) => {
...
-      const arrayToMutate = arrayFromSession(arrayName, flowSession);
-      deleteFromArrayInplace(arrayToMutate, index);
-      updateSession(flowSession, { [arrayName]: arrayToMutate });
-      const headers = { "Set-Cookie": await commitSession(flowSession) };
-      return new Response("success", { status: 200, headers });
-    } catch (err) {
-      return new Response((err as Error).message, { status: 422 });
-    }
+  const clonedFormData = await request.clone().formData();
+  if (clonedFormData.get("_action") === "delete") {
```
chohner, line 240: I wonder if we could get away without cloning here..

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
Spencer6497: `flowId` actually begins with a slash

app/routes/shared/formular.server.ts
Spencer6497: I think we actually have to, in order to access the `_action` field, as it's nested in `formData`... and as soon as we call `formData()`, the request is transformed and it can't be called again

app/util/file/fileSchema.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from "zod";
+
+const maxFileSize = 10 * 1024 * 1024; // 10MB
```
chohner, line 3: I guess this might become a parameter in the future but for now - maybe `10MB_IN_BYTES` for self-descriptive variable?

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
```diff
@@ -0,0 +1,68 @@
...
+export async function uploadUserFileToS3(
+  cookieHeader: string | null,
+  url: string,
+  file: File,
+) {
+  try {
+    const s3Client = createClientS3DataStorage();
+    const flowId = flowIdFromPathname(new URL(url).pathname);
+    if (!flowId)
+      throw new Error(`Attempted to upload user file outside of known flow`);
```
chohner, line 49: why not simply pass the flowId?

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
```diff
@@ -0,0 +1,68 @@
...
+import { flowIdFromPathname } from "~/domains/flowIds";
+import { config } from "~/services/env/env.server";
+import { createClientS3DataStorage } from "~/services/externalDataStorage/createClientS3DataStorage";
+import { sendSentryMessage } from "~/services/logging";
+import { getSessionIdByFlowId } from "~/services/session.server";
+
+const USER_FILES_FOLDER = "user-files";
+const bytesInKilobyte = 1024;
+
+const createFolderKey = (sessionId: string, flowId: string) => {
```
chohner, line 11: `FlowId` type?

app/services/flow/server/parseMultipartFormData.server.ts
```diff
@@ -0,0 +1,98 @@
...
+  const relevantFormData = filterFormData(formData);
+  try {
+    const { arrayName, index } = arrayIndexFromFormData(relevantFormData);
+    const arrayToMutate = arrayFromSession(arrayName, flowSession);
+    deleteFromArrayInplace(arrayToMutate, index);
+    updateSession(flowSession, { [arrayName]: arrayToMutate });
+    const headers = { "Set-Cookie": await commitSession(flowSession) };
+    return new Response("success", { status: 200, headers });
+  } catch (err) {
+    return new Response((err as Error).message, { status: 422 });
```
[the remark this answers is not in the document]
Spencer6497: Since it's a self-contained helper function, I just moved it to `arrayDeletion.ts`

app/util/file/fileSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const fileSchema = (
```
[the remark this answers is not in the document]
Spencer6497: very true :)

app/util/file/fileSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+  typeof window === "undefined" ? z.any() : z.instanceof(File)
```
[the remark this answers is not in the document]
Spencer6497: it'll still validate normally, we just lose the added layer of `z.instanceOf` safety

app/services/externalDataStorage/storeUserFileToS3Bucket.ts
Spencer6497: mainly to avoid passing it through 3 layers of functions :/

PR #1651, opened 2025-01-27
ZOV: each addition person needs consent

pgurusinga:


app/domains/fluggastrechte/formular/persoenlicheDaten/context.ts
```diff
@@ -29,6 +30,7 @@ export const fluggastrechtePersoenlichDaten = {
       .object({
         ...persoenlicheDatenSchema,
         buchungsnummer: optionalOrSchema(bookingNumberFlightSchema),
+        datenverarbeitungZustimmung: checkedRequired,
```
aaschlote, line 33: I haven't seen in the AC of the ticket, but the user should be always given the consent, right?

PR #1656, opened 2025-01-27
chore: allow config to configure array index

pgurusinga:
## ⚠️ Before Merging ⚠️ 
Please deploy content, as we need to update the string interpolation
TODO:
- [ ] Replace `{{ indexPerson }}` with `{{ indexArray }}`


app/components/arraySummary/ArraySummaryDataItems.tsx
```diff
@@ -41,8 +41,7 @@ const ArraySummaryDataItems = ({
   const heading = interpolateSerializableObject(
     translations[`${category}.label.heading`] ?? "",
     {
-      indexPerson: (itemIndex + WEITERE_PERSONEN_START_INDEX).toString(),
-      indexArray: (itemIndex + 1).toString(),
+      indexArray: ((customStartDisplayIndex ?? 1) + itemIndex).toString(),
```
aaschlote, line 44: maybe we can keep the logic as before:

```suggestion
      indexArray: (itemIndex + (customStartDisplayIndex ?? 1)).toString(),
```

aaschlote (approved):
LGTM!

app/services/array/index.ts
```diff
@@ -11,6 +11,7 @@ export type ArrayConfigServer = {
   url: string;
   initialInputUrl: string;
   statementKey: AllContextKeys;
+  customStartDisplayIndex?: number;
```
chohner, line 14: Could we name this `indexOffset` or similar?

pgurusinga: The name `indexOffset` might be misleading as it suggests adjusting the actual index, but it only affects how the index is shown in the UI. To clarify its purpose, I suggest `displayIndexOffset`. What do you think? Also, @aaschlote, since you suggested the change earlier.

app/components/arraySummary/ArraySummaryDataItems.tsx
pgurusinga: since we are writing the code from left to right and `customStartDisplayIndex` is optional, my brain can think better if the optional variable lives on the left side :D but if you like I can revert

app/services/array/index.ts
aaschlote: sounds good for me, maybe a small comment on the side of this config attribute does it might help others 😃

pgurusinga:
PKH
Image Description
Image Description

PR #1666, opened 2025-01-29
Fix(Feedback component): get feedback state based on cookies

khallad2:
- check the cookies for the feedback result from previous rating 
    - if wasHelpful already has (true or false) that means the user already gave rating before then we render the message 
    - if the cookies doesn't have wasHelpful then we show the rating box to give rating


app/components/userFeedback/index.tsx
```diff
@@ -27,11 +27,13 @@ export enum FeedbackType {
 export default function UserFeedback(props: Readonly<UserFeedbackProps>) {
   const { pathname } = useLocation();
   const [shouldFocus, setShouldFocus] = useState(false);
-  const [feedback, setFeedback] = useState<FeedbackType>();
-
   const rootLoaderData = useRouteLoaderData<RootLoader>("root");
   const bannerState = rootLoaderData?.bannerState ?? BannerState.ShowRating;
```
aaschlote, line 31: can't we not create an object where is stored in the root from the server to the client the `bannerState` and the `feedbackResult`? It would be something:

```
feedbackState: {
result: "something",
banner: "anotherThing"
}
```

app/root.tsx
```diff
@@ -81,11 +85,21 @@ export const meta: MetaFunction<RootLoader> = () => {
   ];
 };
 
+const getFeedbackResult = async (
+  request: Request,
+): Promise<boolean | undefined> => {
+  const cookieHeader = request.headers.get("Cookie");
+  const { getSession } = getSessionManager("main");
+  const session = await getSession(cookieHeader);
+  return session.get("wasHelpful")?.["/hilfe"] ?? undefined;
```
aaschlote, line 94: please remember the feedback component is not only on the `hilfe` page, but anywhere in the website. Check the function`getFeedbackBannerState`

app/components/userFeedback/index.tsx
khallad2: i think this is similar to what happens now in the root.tsx -> loader function. 
the loader function gets the stored feedbackResult from the cookies and returns json(rootLoaderData) that contains the feedbackResult along with the all other data to be passed to userFeedback/index.tsx then showFeedback or showRating based on the feedbackResult that comes from root.tsx.

so simply everything we need in rootLoaderData object.
my approach was trying to avoid nested objects like
rootLoaderData {
      feedbackState: {
            result: "something",
            banner: "anotherThing"
       }
}

but if you think it's important to be nested object please let me know .

aaschlote: Exactly, we have right now in the root a variable for the bannerState (created by me btw) and now you need a new one for the result of the feedback. I'm aware you need the values in the root.tsx, no doubts about it.

What I want to avoid it's to have more and more variables in the root, so you can wrap up in the root.tsx as an object called `feedback`. And any case you need a third or fourth variable for the feedback, you don't need to create more variables, just add to the object 😃

Spencer6497 (approved):
looks great! 🚀

PR #1671, opened 2025-01-30
refactor(AutoSuggestionInput): add more a11y messages

aaschlote:


app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
```diff
@@ -0,0 +1,223 @@
...
+              translations,
+            );
+          }
+
+          message += getTranslationByKey(
+            GUIDANCE_INPUT_OPEN_MENU,
+            translations,
+          );
+
+          if (isMulti) {
```
aaschlote, line 80: maybe we can remove this scope of `isMulti`, we don't use this feature. What do u think?

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
```diff
@@ -0,0 +1,223 @@
...
+  ) => {
+    const { action, label = "", labels, isDisabled } = props;
+    switch (action) {
+      case "deselect-option":
+      case "pop-value":
+      case "remove-value": {
+        const message = getTranslationByKey(ON_CHANGE_DESELECTED, translations);
+        return message.replace("{{label}}", label);
+      }
+      case "clear":
```
aaschlote, line 112: this options might not get reach, because to clear the value we use the button on the left corner. Maybe we can remove it as well.

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
```diff
@@ -0,0 +1,223 @@
...
+        return "";
+    }
+  },
+
+  onChange: <Option, IsMulti extends boolean>(
+    props: AriaOnChangeProps<Option, IsMulti>,
+  ) => {
+    const { action, label = "", labels, isDisabled } = props;
+    switch (action) {
+      case "deselect-option":
```
aaschlote, line 106: do we have this option in your component? Not sure if we have it.

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
pgurusinga: yeah I think we could remove it.

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
pgurusinga: actually this action is called when I clear the value :(
image
image
image

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
pgurusinga: in our current implementation not, but this is API from the `react-select`, this comes from the doc, it's related to multiple selection, I guess we can remove
> (Multiple) Deselecting an option from the list

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
aaschlote: Ok, I couldn't simulate it, but if you can we can keep it

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
aaschlote: yeap, better remove, to keep the code simple :)

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
```diff
@@ -0,0 +1,213 @@
...
+            );
+          }
+
+          message += getTranslationByKey(
+            GUIDANCE_INPUT_OPEN_MENU,
+            translations,
+          );
+
+          return message;
+        } else {
```
aaschlote, line 79: the `else` condition is needed, just return empty after the scope above

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
```diff
@@ -0,0 +1,213 @@
...
+        const labelsJoined = labels.join(", ");
+        const key =
+          labels.length > 1
+            ? ON_CHANGE_INITIAL_INPUT_FOCUS_PLURAL
+            : ON_CHANGE_INITIAL_INPUT_FOCUS_SINGULAR;
+        const message = getTranslationByKey(key, translations);
+        return message.replace("{{labels}}", labelsJoined);
+      }
+      case "select-option": {
+        if (isDisabled) {
```
aaschlote, line 114: can you check if this option is reachable (isDisabled)? We disabled the component, but the user can't touch it!

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
```diff
@@ -0,0 +1,213 @@
...
+      case "remove-value": {
+        const message = getTranslationByKey(ON_CHANGE_DESELECTED, translations);
+        return message.replace("{{label}}", label);
+      }
+      case "clear":
+        return getTranslationByKey(ON_CHANGE_CLEAR, translations);
+      case "initial-input-focus": {
+        const labelsJoined = labels.join(", ");
+        const key =
+          labels.length > 1
```
aaschlote, line 107: the `labels` is only for multiFields, I guess we can remove it as well

aaschlote (commented):
Besides my comments, LGTM!

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
pgurusinga: 🤔 you mean not needed?

aaschlote: ops, yeah, not needed

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
pgurusinga: I cannot reach it either, and I'm unsure if that's correct behavior. I asked Marion and will wait for the result.

aaschlote: I believe it's.

judithmh (approved):

app/components/inputs/autoSuggestInput/accessibilityConfig/ariaLiveMessages.ts
pgurusinga: alright, I will remove this.

m0dh4x (approved):
LGTM

PR #1673, opened 2025-01-30
update(Feedback component)

khallad2:


app/services/feedback/getFeedbackBannerState.ts
```diff
@@ -1,14 +1,20 @@
 import { type Session } from "@remix-run/node";
-import type { BannerState } from "~/components/userFeedback";
+import { BannerState } from "~/components/userFeedback";
 
 export const bannerStateName = "bannerState";
+export const userRatingFieldName = "wasHelpful";
 
 export const getFeedbackBannerState = (session: Session, url: string) => {
```
aaschlote, line 7: Can rename to `getFeedbackData`? Before it was getting only the info about the Banner state of the feedback, now it's getting data of the feedback, what's included the `bannerState` and `resultOfFeedback`

aaschlote, line 7: Can we add unit test for that?I guess it was missing before as well.

app/root.tsx
```diff
@@ -160,12 +146,7 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
       ),
       videoTranslations: translations.video,
       accessibilityTranslations: translations.accessibility,
-      bannerState: {
-        feedbackResult: await getFeedbackResult(request, cookieHeader),
-        state:
-          getFeedbackBannerState(mainSession, pathname) ??
-          BannerState.ShowRating,
-      },
+      bannerState: getFeedbackBannerState(mainSession, pathname),
```
aaschlote, line 149: either `feedback` or `feedbackData`. This value does contain more than the bannerState

app/services/feedback/getFeedbackBannerState.ts
```diff
@@ -1,14 +1,20 @@
...
-  const getSessionForFeedbackState = session.get(bannerStateName) as Record<
-    string,
-    BannerState
-  >;
-  if (getSessionForFeedbackState) {
-    return getSessionForFeedbackState[url];
-  }
+  const getSessionForFeedbackState =
```
aaschlote, line 8: It's variable, so better call:

```suggestion
  const feedbackState =
```

or 

```suggestion
  const state =
```

app/services/feedback/getFeedbackBannerState.ts
```diff
@@ -1,14 +1,20 @@
...
+    (session.get(bannerStateName) as Record<string, BannerState>) ??
+    BannerState.ShowRating;
+
+  const getSessionForFeedbackResult =
```
aaschlote, line 12: Same here:

```suggestion
  const feedbackResult =
```
or
```suggestion
  const result =
```

app/components/userFeedback/index.tsx
```diff
@@ -29,9 +29,8 @@ export default function UserFeedback(props: Readonly<UserFeedbackProps>) {
   const [shouldFocus, setShouldFocus] = useState(false);
   const rootLoaderData = useRouteLoaderData<RootLoader>("root");
   const bannerState =
-    rootLoaderData?.bannerState.state ?? BannerState.ShowRating;
-
-  const feedbackResult = rootLoaderData?.bannerState.feedbackResult
+    rootLoaderData?.bannerState.state[pathname] ?? BannerState.ShowRating;
```
aaschlote, line 32: do you need to get the state by the `pathname?`

aaschlote (changes requested):
Thanks for the changes, it sounds much better now. But we can do more improvements 😃

app/components/userFeedback/index.tsx
khallad2: yes it's mandatory because the record you were retrieving already has the key as the pathname.

app/root.tsx
khallad2: i don't get what you mean ?

app/services/feedback/getFeedbackBannerState.ts
khallad2: i was following the same naming conventions you have used not to look odd, but yeah you are right the naming was not the best, will change it.

app/components/userFeedback/index.tsx
aaschlote: Right now we are getting the bannerState direct from the url in the server, check [here](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/app/services/feedback/getFeedbackBannerState.ts#L12)

You changed in the function to return all the banner states from all the pages in the server (`getFeedbackData`), but you can retrieve the banner state specific of the page as you are doing for the `result`

app/services/feedback/getFeedbackBannerState.ts
aaschlote: That naming convention was wrong, my mistake 🥲

Thanks for fixing it!

app/components/userFeedback/index.tsx
khallad2: i have added it to the getFeedbackData

app/services/feedback/getFeedbackData.ts
```diff
@@ -6,11 +6,11 @@ export const userRatingFieldName = "wasHelpful";
 
 export const getFeedbackData = (session: Session, url: string) => {
   const state =
-    (session.get(bannerStateName) as Record<string, BannerState>) ??
+    (session.get(bannerStateName) as Record<string, BannerState>)?.[url] ??
     BannerState.ShowRating;
 
   const result =
-    (session.get(userRatingFieldName)?.[url] as Record<string, BannerState>) ??
+    (session.get(userRatingFieldName) as Record<string, BannerState>)?.[url] ??
```
aaschlote, line 13: shouldn't be a record of `FeedbackType`?

``(session.get(userRatingFieldName) as Record<string, FeedbackType>)?.[url] ??``

aaschlote (approved):
LGTM!

PR #1680, opened 2025-01-31
Enable integration tests

aaschlote:
### Background
We have an integration test to verify whether the pages and contexts in the on-flow configuration are available in Strapi. To obtain the latest Strapi content data, we retrieve the most recent `content.json` from production.

### Problem
When new fields are added to the context of existing pages, the integration test fails because these fields are not present in the latest `content.json` from production. Additionally, we cannot deploy the updated content because the system crashes if a field is not yet defined in the on-flow configuration context.

### Solution
Before running the integration tests, we first load the latest content data from Strapi and then execute the tests. This ensures that the tests run against the most up-to-date Strapi content, preventing failures due to missing fields.


tests/integration/runTests.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+# Get the latest content data from Strapi before run the tests
+npm run build:localContent
+CMS="FILE"
+# After get the latest content data from Strapi, run the tests
+npx vitest run --project=integration
```
chohner, line 6: I'm confused. I thought the idea is to run the test directly against strapi? why would we download content and force an offline test?

package.json
```diff
@@ -33,7 +33,7 @@
     "test:e2e": "npx playwright test",
     "test": "vitest run --project=!integration",
     "test:watch": "vitest --project=!integration",
-    "test:integration": "vitest run --project integration",
+    "test:integration": "./tests/integration/runTests.sh",
```
chohner, line 36: I would strongly prefer a solution without shell script, like `npm run build:localContent && CMS="FILE" vitest run --project integration`

tests/integration/runTests.sh
Spencer6497: this was also my understanding?

aaschlote: yeap, but I would need to make a big refactor in the tests and how Strapi gets slow after the update to the version 5 the tests might fail due the timeout (vitest problems). 

So getting the all current Strapi content and tests offline it'll be the same thing as testing directly against Strapi. And it runs fast as well, because I don't need to retrieve every single flow or content from strapi to test it.

package.json
aaschlote: Why would you strongly prefer a solution without shell script? I did a shell script to avoid a long command line. And in case we need to add more options in the command line, adding in the shell script is easier than adding here.

chohner: For me: Much easier to understand whats going on + less files in the repo (especially hard-to-debug executables like shell scripts)

aaschlote: For me, it's much easier to read multiple lines in a shell script than to read a long single-line command. 😅

tests/integration/runTests.sh
pgurusinga: fyi; I think this script might not throw any errors in the pipeline if something goes wrong unless you turn on `set -e`, which you don't need to do if it's in package.json

aaschlote: @pgurusinga I believe you commented in the wrong thread discussion 😄

pgurusinga: I think I am in the correct thread 😄, but shouldn't we add `set -e` here?
correct me if I am wrong, but If `npx vitest run --project=integration` fails (e.g., tests fail), the script still completes normally without stopping or indicating failure at the script level and npm won't know. 

When using the npm command directly, the `set -e` option is unnecessary, as npm stops executing the script if any command fails.
I think it's preferable to use the npm task directly rather than relying on the complexity I mentioned earlier, which involves missing `set -e`.

reference: https://draculaservers.com/tutorials/bash-set-e-explained/

aaschlote: @pgurusinga as we have seen, the pipeline will fail in case the integration test fails. See [here](https://github.com/digitalservicebund/a2j-rechtsantragstelle/actions/runs/13177221979/job/36779272218)

pgurusinga: Pre-downloading the content makes the test run faster. Should we merge this pull request if there is no further discussion?


image

pgurusinga (approved):

PR #1686, opened 2025-02-03
Ref: replace-axios-with-fetch, fix unit test and update package.json

khallad2:
We use axios for making http calls at three locations in the app
non-native package increases bundle size (slower performance), and increases chance of security issues


app/routes/health.ts
```diff
@@ -1,17 +1,33 @@
...
+      logError({ error: "Redis connection not ready" });
+      return new Response("ERROR: Redis connection not ready", { status: 503 });
+    }
+
     if (config().CMS === "STRAPI") {
-      await axios.get(`${config().STRAPI_HOST}/_health`, {
-        validateStatus: (status) => status < 300,
-        headers: { Authorization: "Bearer " + config().STRAPI_ACCESS_KEY },
-      });
+      const response: Response = await fetch(
```
chohner, line 13: I don't think `response` needs an explicit type?

app/routes/health.ts
```diff
@@ -1,17 +1,33 @@
...
+        `${config().STRAPI_HOST}/_health`,
+        {
+          headers: { Authorization: `Bearer + ${config().STRAPI_ACCESS_KEY}` },
+        },
+      );
+
+      if (!response.ok) {
```
chohner, line 20: maybe `fetch().then((response) => {})` is a more useful pattern here?

app/routes/__test__/health.test.ts
```diff
@@ -0,0 +1,124 @@
+import { describe, it, expect, beforeEach } from "vitest";
```
chohner, line 1: nice job on testing this 🙏🏼

app/routes/health.ts
```diff
@@ -1,17 +1,33 @@
...
 import { logError } from "~/services/logging";
 import { getRedisStatus } from "~/services/session.server/redis";
 
-export const loader = async () => {
+export const loader = async (): Promise<Response> => {
   try {
-    if (getRedisStatus() !== "ready") throw Error("Redis connection not ready");
+    if (getRedisStatus() !== "ready") {
+      logError({ error: "Redis connection not ready" });
+      return new Response("ERROR: Redis connection not ready", { status: 503 });
```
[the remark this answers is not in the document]
khallad2: i thought it's common error status 503 => service not available when Redis is down.
that's why i return a response with 503, do you think it's useless?

app/services/cms/__test__/getStrapiEntryFromApi.test.ts
```diff
@@ -22,25 +21,32 @@ describe("services/cms", () => {
...
-      expect(axiosGetSpy).toHaveBeenNthCalledWith(
+      expect(fetchSpy).toHaveBeenNthCalledWith(
         1,
         `${expectedStagingRequestUrl}&filters[slug][$eq]=foobar`,
-        expect.anything(),
+        expect.objectContaining({
+          headers: expect.objectContaining({
+            Authorization: expect.stringContaining("Bearer"),
+          }),
+        }),
```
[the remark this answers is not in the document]
khallad2: Okay, i thought validating headers in every test is a good approach.

app/routes/health.ts
khallad2: my thoughts of not using .then were:
- following current practice which already using async/await for most of asynchronous calls in the app.
- promise chaining can easily get hard to read/maintain in the future in case of adding more nested logic.

please let me know why you prefer chaining with .then() in this case?

app/services/cms/getStrapiEntryFromApi.ts
```diff
@@ -42,17 +41,25 @@ const buildUrl = <T extends ApiId>({
...
+const makeStrapiRequest = async <T extends ApiId>(
+  url: string,
+): Promise<{ data: StrapiSchemas[T] | null }> => {
+  const response = await fetch(url, {
     headers: {
       Authorization: "Bearer " + config().STRAPI_ACCESS_KEY,
     },
   });
 
+  if (response.status >= 500) {
```
pgurusinga, line 53: shouldn't we add 4XX error handling? e.g invalid or expired key or do you think this won't happen?
```ts

    if (response.status >= 400 && response.status < 500) {
      ...
    }
```

khallad2: Thanks, that's valid point, from my POV it can happen, i just replaced axios validation condition with this if condition. so i would do something like

```
 (!response.ok) {
   ...
 }
```

pgurusinga: let's do it similarly; we can relay what we receive.
example:
```ts
    if (!ok) {
      throw new Error(
        `Strapi request failed with error: ${response.status} - ${JSON.stringify(responseBody)}`
      );
    }
```

app/routes/__test__/health.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const response = await loader();
+    expect(response.status).toBe(200);
+    const text = await response.text();
+    expect(text).toBe("I'm fine, thanks for asking :)");
+  });
+
+  it("skips Strapi check if CMS is something else", async () => {
+    getRedisStatus.mockReturnValue("ready");
+    config.mockReturnValue({
```
aaschlote, line 112: maybe you can create a function for this mock, as you are calling more than one time and just modify the CMS value. It could be:

```
const mockConfig = (CMS: string) => {
config.mockReturnValue({
      CMS,
      STRAPI_HOST: "http://fake-strapi.example",
      STRAPI_ACCESS_KEY: "FAKE_KEY",
    });
}
```

app/services/cms/getStrapiEntryFromApi.ts
```diff
@@ -42,17 +41,27 @@ const buildUrl = <T extends ApiId>({
...
+): Promise<{ data: StrapiSchemas[T] | null }> => {
+  const response = await fetch(url, {
     headers: {
       Authorization: "Bearer " + config().STRAPI_ACCESS_KEY,
     },
   });
 
+  if (!response.ok) {
+    throw new Error(
+      `Strapi request failed with error: ${response.status} - ${JSON.stringify(response.body)}`,
```
aaschlote, line 55: maybe getting the `json` of the response is better?

```
const content = await response.json();
    throw new Error(
      `Strapi request failed with error: ${response.status} - ${content}`,
    );
```

app/services/cms/__test__/getStrapiEntryFromApi.test.ts
```diff
@@ -22,25 +21,33 @@ describe("services/cms", () => {
     };
     const expectedStagingRequestUrl = `${API_URL}pages?populate=*&pLevel&locale=sg`;
 
-    const axiosGetSpy = vi.spyOn(axios, "get");
+    const fetchSpy = vi.spyOn(global, "fetch");
```
aaschlote, line 24: I'm getting this in my VSCode when I open this file, can you please check if you are seeing it also?

image

app/services/flow/submission.server.ts
```diff
@@ -9,16 +8,25 @@ type SubmitRequest = {
...
   };
 
-  return (
-    await axios.post(
-      `${FIT_CONNECT_ADAPTER_ENDPOINT}/api/sender/submit`,
-      data,
-      { headers: { "Content-Type": "application/json" } },
-    )
-  ).data;
+  const headers = {
```
aaschlote, line 17: why this variable is called headers if contains body and method?

aaschlote (commented):
I wrote some comments

app/services/cms/__test__/getStrapiEntryFromApi.test.ts
khallad2: i have no errors also, `npm run lint` no errors. 
image

app/services/cms/getStrapiEntryFromApi.ts
khallad2: **_content will use Object's default stringification format ('[object Object]') when stringified_**

aaschlote: I didn't get your last comment

app/routes/__test__/health.test.ts
khallad2: better to be in beforeEach scope.

app/services/cms/getStrapiEntryFromApi.ts
khallad2: i meant we have to Stringfy the content /json anyway to display the error?

`const content = await response.json();
    throw new Error(
      `Strapi request failed with error: ${response.status} - ${content}`,
    );
// eslint-error: content will use Object's default stringification format ('[object Object]') when stringified
`
i have applied exception handling 
`try {
    const response = await fetch(url, {
      headers: {
        Authorization: `Bearer ${config().STRAPI_ACCESS_KEY}`,
      },
    });

    return response.json();
  } catch (error: unknown) {
    throw new Error(`Strapi request failed: ${(error as Error)?.message}`);
  }`

app/routes/health.ts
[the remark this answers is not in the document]
chohner: not at all, was more thinking out loud on the semantics of returning a `Response` vs throwing and `Error`

app/routes/__test__/health.test.ts
khallad2: here you can find the changes 
https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1686/commits/ba68c3fdc40872b09d407a911623c8fcc97382a5

app/routes/health.ts
chohner: no, totally up to you!

Another note: I generally prefer to reduce the numbers of `return` statements, so maybe something like this could simplify the control flow? (again up to you)

```typescript
const message = response.ok ? "..." : "..."
const status = response.ok ? 200 : 503
return new Response(message, { status })
```

app/services/cms/getStrapiEntryFromApi.ts
aaschlote: I don't see the eslint error, but now checking the property of the `body`, the body will return the value of request method and for `get` methods will be null. Check [here](https://developer.mozilla.org/en-US/docs/Web/API/Request/body).

app/routes/__test__/health.test.ts
```diff
@@ -0,0 +1,117 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import { loader } from "../health";
+
+vi.mock("~/services/session.server/redis", () => {
+  return {
+    getRedisStatus: vi.fn(),
+  };
+});
+
+vi.mock("~/services/env/env.server", () => {
```
aaschlote, line 10: maybe you can do only this

```
vi.mock("~/services/env/env.server", () => ({
  config: vi.fn(),
}));
```

app/services/cms/getStrapiEntryFromApi.ts
khallad2: yes that will be {}, so i surrounded it with try/catch to display bubbled error and throw it in an exception. 
here is the commit https://github.com/digitalservicebund/a2j-rechtsantragstelle/commit/ba68c3fdc40872b09d407a911623c8fcc97382a5

do you see issues with the new updates?

aaschlote: yeap

aaschlote (approved):
LGTM!

pgurusinga (approved):

PR #1697, opened 2025-02-06
fix: ignore end of line during parsing

pgurusinga:
## Problem
Users can click without choosing an airline, and validations do not fail. When they click forward, and then return to the airline page, it is set to Malta Air.

## Solution
- Malta is missing the IATA code
- Fix EoL error 

## Next Steps
We need to throw an error when the data being parsed is invalid.

### DIFF
--- a/data/airlines/data.json
+++ b/data/airlines/data.json
@@ -861,8 +861,8 @@
   },
   {
     "name": "Malta Air Limited",
-    "iata": "",
-    "isInEU": false,
+    "iata": "MW",
+    "isInEU": true,
     "arbitrationBoard": "RV"
   },
   {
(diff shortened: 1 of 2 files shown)


data/airlines/data.json
```diff
@@ -861,8 +861,8 @@
   },
   {
     "name": "Malta Air Limited",
-    "iata": "",
-    "isInEU": false,
+    "iata": "MW",
```
aaschlote, line 864: maybe it's better to use the iata code `MAY`. [Reference](https://en.m.wikipedia.org/wiki/Malta_Air)

aaschlote (approved):
can you add this condition to the `app/services/validation/airline.ts` schema?

``
.nonempty({
    message: "invalid_airline_code",
  })
``

So we force the schema to no empty.

PR #1700, opened 2025-02-06
refactor(annullierung): adapt flow when ersatzflug is no

aaschlote:
Ticket: https://app.asana.com/0/1207022303033556/1209116565459196

### Flow

GIVEN a user is filling out the Klage flow in case of Annullierung
WHEN reaching the section Flugdaten
THEN there is a new page /flugdaten/ersatzverbindung-daten and this page is visible
IF in Vorab-Check on page /vorabcheck/ersatzflug user selected [JA]
show the new page after page /flugdaten/geplanter-flug if 0 Zwischenstopps
show the new page after page e /flugdaten/verspaeteter-flug-1/2/3 if 0 remaining flight after seleceted flight
show the new page after page /flugdaten/anschluss-flug-verpasst
ELSE IF don't show this new page at all but directly go to /flugdaten/zusaetzliche-angaben after the described pages above
page after the new page should always be /flugdaten/zusaetzliche-angaben


app/domains/__test__/flows.test.ts
```diff
@@ -15,7 +15,7 @@ import { testCasesBeratungshilfeFormularGrundvoraussetzungen } from "~/domains/b
 import { testCasesBeratungshilfeRechtsproblem } from "~/domains/beratungshilfe/formular/rechtsproblem/__test__/testcases";
 import { testCasesBeratungshilfe } from "~/domains/beratungshilfe/vorabcheck/__test__/testcases";
 import { type Context } from "~/domains/contexts";
-import { testCasesFluggastrechteFormularFlugdatenAnnullierung } from "~/domains/fluggastrechte/formular/flugdaten/__test__/testcasesAnnullierung";
+import { testCasesFluggastrechteFormularFlugdatenAnnullierungWithErsatzflugYes } from "~/domains/fluggastrechte/formular/flugdaten/__test__/testcasesAnnullierungWithErsatzflugYes";
```
pgurusinga, line 18: this testcase naming is getting super long like a train 🚋 🚋 🚋 🚋 🚋 🚋 
we might need to find a better structure, but not here.

app/domains/fluggastrechte/components/ActualArrivalCards.tsx
```diff
@@ -26,14 +26,17 @@ function ActualArrivalCards({
...
+      return (
+        <SummaryDataOverviewCard
+          title="Angebotene Ersatzverbindung"
+          buttonUrl={`${FLOW_ID}/flugdaten/ersatzverbindung-daten`}
+          data={getAnnullierungData(userData)}
+          translations={translations}
+        />
+      );
+    }
+    return <></>;
```
pgurusinga, line 39:
```suggestion
    return null;
```
https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null

I assume if we return Fragment, it will render an empty div

pgurusinga (approved):
beside my comment

app/domains/__test__/flows.test.ts
aaschlote: yeap, I was thinking to move the unit tests of the tests cases to the domain level, so we can have less imports and small names

PR #1701, opened 2025-02-06
Remove strapi prop getter

khallad2:
**Refactor Replace getXXXProps with Zod .transform() for StrapiParagraph**

- Refactors StrapiParagraph to use Zod .transform() instead of the getXXXProps functions.
- Fixes related errors and updates affected unit tests, particularly around StrapiParagraph.
Streamlines the codebase by removing redundant “props” logic and centralizing validation in Zod.


aaschlote:
I know it's in draft, but you can check the possibility to use the [custom option](https://zod.dev/?id=custom-schemas). So we can reuse the props type of component to validate the data from strapi. I know it might not work in all components, but in a few components we can do it and we avoid to create two objects for the component, one for the schema and another for props of the component. Example:

```
export const StrapiVideoSchema2 = z
  .custom<VideoProps>()
  .and(z.object({ __component: z.literal("page.video") }))
  .and(HasOptionalStrapiIdSchema);
```

app/components/Footer.tsx
```diff
@@ -6,7 +6,7 @@ import { StandaloneLink } from "./StandaloneLink";
 
 type LinkProps = {
   url: string;
-  text?: string;
+  text?: string | null;
```
chohner, line 9: this isn't obvious but we have avoided `null` until now (Which makes the interface to the CMS a bit more annoying). any reason this would need to be here now?

app/components/BoxWithImage.tsx
```diff
@@ -38,7 +38,13 @@ const BoxWithImage = ({
...
         className={`lg:shrink-0 overflow-hidden ${hasTextContent ? variantWidths[variant] : "max-w-full"}`}
       >
-        <Image {...image} />
+        <Image
+          url={image.url}
+          className={image.className}
+          height={image.height}
+          width={image.width}
+          alternativeText={image.alternativeText}
+        />
```
chohner, line 47: since `image` is of type `ImageProps`, couldn't this stay the same?

app/root.tsx
```diff
@@ -132,7 +131,7 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
           !flowIdFromPathname(pathname)?.match(/formular|antrag/),
         showKopfzeile,
       },
-      footer: getFooterProps(strapiFooter),
+      footer: strapiFooter,
```
chohner, line 134: why is this removed?

app/root.tsx
```diff
@@ -229,7 +228,7 @@ function App() {
             )}
             target={skipToContentLinkTarget}
           />
-          <CookieBanner content={getCookieBannerProps(cookieBannerContent)} />
+          <CookieBanner content={cookieBannerContent} />
```
chohner, line 231: same here?

app/routes/$.tsx
```diff
@@ -17,5 +18,11 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
 };
 
 export default function Index() {
-  return <PageContent content={useLoaderData<typeof loader>().content} />;
+  return (
+    <PageContent
+      content={
+        useLoaderData<typeof loader>().content as StrapiContentComponent[]
```
chohner, line 24: This (and all the following) type casting means something is wrong with the types

app/services/cms/getStrapiEntryFromFile.ts
```diff
@@ -36,7 +36,6 @@ export const getStrapiEntryFromFile: GetStrapiEntry = async <T extends ApiId>(
             ? value.includes(relevantField as string)
             : relevantField === value;
         }
-
```
chohner, line 39: one more change we can remove from the PR :)

app/services/cms/models/StrapiHeader.ts
```diff
@@ -28,8 +28,7 @@ type StrapiHeaderComponent = z.infer<typeof StrapiHeaderComponentSchema>;
 export const getHeaderProps = ({
   content,
   heading,
-}: StrapiHeader): HeaderProps =>
-  omitNull({ heading, content: content && getRichTextProps(content) });
+}: StrapiHeader): HeaderProps => omitNull({ heading, content: content });
```
chohner, line 31: nitpick: `content` is enough (instead of `content: content`)

app/services/cms/models/StrapiImage.ts
```diff
@@ -33,16 +33,6 @@ export const StrapiImageSchema = z
...
-    if (!cmsData) return undefined;
-    const { url, width, height, alternativeText } = cmsData;
-    return {
-      url,
-      width,
-      height,
-      alternativeText,
-    };
-  });
+  .transform((cmsData) => cmsData ?? null);
```
chohner, line 36: let's not leak `null` into our front-end here

app/root.tsx
```diff
@@ -237,7 +236,10 @@ function App() {
           </TranslationContext.Provider>
           <footer>
             <Footer
-              {...footer}
+              {...(footer as unknown as Record<
+                string,
+                StrapiContentComponent[]
+              >)}
```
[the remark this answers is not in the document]
khallad2: one second, this is not the last changes i believe. i removed all (unknown) casts.    (outdated)

app/services/cms/models/StrapiBox.ts
```diff
@@ -31,6 +31,6 @@ export const StrapiBoxComponentSchema = StrapiBoxSchema.extend({
 export const getBoxProps = (cmsData: StrapiBox): BoxProps => {
   return omitNull({
     ...pick(cmsData, "label", "heading", "buttons", "identifier"),
-    content: cmsData.content && getRichTextProps(cmsData.content),
+    content: cmsData.content,
```
Spencer6497, line 34: can we collapse this into the `pick`?

app/services/cms/models/StrapiImage.ts
```diff
@@ -33,16 +33,6 @@ export const StrapiImageSchema = z
...
-  .transform((cmsData) => {
-    if (!cmsData) return undefined;
-    const { url, width, height, alternativeText } = cmsData;
-    return {
-      url,
-      width,
-      height,
-      alternativeText,
-    };
-  });
```
Spencer6497, line 45: i thought we were stripping off extraneous (CMS) properties? I don't think we need to edit `StrapiImage`

app/services/cms/__test__/index.test.ts
```diff
@@ -22,16 +22,18 @@ describe("services/cms", () => {
...
-      vi.mocked(getStrapiEntry).mockReturnValue(Promise.resolve([footerData]));
-      expect(await fetchSingleEntry("footer")).toEqual({
-        ...footerData,
-        image: pick(footerData.image, [
-          "url",
-          "alternativeText",
-          "width",
-          "height",
-        ]),
-      });
```
Spencer6497, line 34: I think we can leave the test as-is once we revert the change to `StrapiImage`

app/services/cms/models/StrapiParagraph.ts
```diff
@@ -1,18 +1,17 @@
...
-export const getRichTextProps = (cmsData: StrapiParagraph): RichTextProps => ({
-  html: cmsData.text,
-});
+  .merge(HasOptionalStrapiIdSchema)
+  .transform((cmsData) =>
+    omitNull({
+      __component: "basic.paragraph" as const,
+      html: cmsData.text,
+      id: cmsData.id,
+      text: cmsData.text,
```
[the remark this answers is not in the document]
khallad2: because in different files it's used as content?.text

for ex:
app/services/meta/metaFromMatches.ts:20:50 
app/services/cms/__test__/index.test.ts:36:35 

i can remove text and change it to always access html attribute since we assign html: cmsData.text anyway.

app/services/cms/models/StrapiImage.ts
khallad2: you are actually right, that was pulled from the first commit on the image component.
updating it.

khallad2:
closed to be moved to a new simpler approach.

PR #1702, opened 2025-02-06
refactor(Kopfzeile): Align Kopfzeile & page header with content container

Spencer6497:
This PR ensures consistent style across the Kopfzeile, page header, and content container components.

- **align Kopfzeile with main container on non-flow pages**
- **add Kopfzeile component test, adapt Storybook**
- **adapt page header and breadcrumbs also**


app/components/PageHeader.tsx
```diff
@@ -19,14 +21,23 @@ export default function Header({
...
+    {
+      "container lg:min-w-[59rem] lg:!mx-auto !py-20": alignToMainContainer,
+    },
+  );
   return (
     <header>
-      {showKopfzeile && <Kopfzeile />}
+      {showKopfzeile && (
+        <Kopfzeile alignToMainContainer={alignToMainContainer} />
+      )}
```
chohner, line 38: why do these classes get adjusted inside and outside the kopfzeile? Any chance we could just apply it to the outer `<header>` element?

app/components/Breadcrumbs.tsx
```diff
@@ -9,45 +10,49 @@ export type Breadcrumb = {
...
 export default function Breadcrumbs({
   breadcrumbs,
   linkLabel,
+  alignToMainContainer = true,
 }: BreadcrumbsProps) {
   const validBreadcrumbs = breadcrumbs?.filter(
     (breadcrumb) => breadcrumb.title !== undefined,
   );
+  const navClassNames = classNames("py-8 px-16 flex flex-wrap items-center", {
+    "container lg:min-w-[59rem] lg:!mx-auto !py-8": alignToMainContainer,
```
chohner, line 25: would it make sense to pull the `container lg:min-w-[59rem] lg:!mx-auto` out into a variable to use across all instances?

app/components/Breadcrumbs.tsx
```diff
@@ -9,45 +10,49 @@ export type Breadcrumb = {
...
-                text={breadcrumb.title ?? ""}
-                url={breadcrumb.url ?? ""}
-                className="increase-tap-area"
-              />
-            )}
-          </div>
-        ))}
-      </nav>
+      <div className="bg-blue-100">
+        <nav className={navClassNames} aria-label="Breadcrumbs">
```
chohner, line 31: Any chance we could avoid the extra nesting here? its probably because the padding is using marging, rather than padding?

chohner (approved):
lgtm, just two more thoughts

app/components/Breadcrumbs.tsx
Spencer6497: yeah totally! just need to find an appropriate place to put it (without creating a circular dependency 😅 )

PR #1706, opened 2025-02-07
feat: allow to pass alternative schema

pgurusinga:
For the FGR case, we need to set the value of the `annullierung` schema to an empty string, as it is optional.


aaschlote (approved):
LGTM!

PR #1720, opened 2025-02-11
doc(adr): Add non-javascript support adr

frederike-ramin:
In fact, this ADR only puts in writing what we are already doing for the most part. This is supposed to serve as a reference for future implementations.

If you think anything is not in line with how we're currently developing and/or with what you had in mind how we will be developing in the future, please leave a comment 🙏


doc/adr/0021-users-without-javascript.md
```diff
@@ -0,0 +1,48 @@
...
+  - Improved accessibility for users with disabilities, limited bandwidth, or JavaScript disabled/unavailable.
+  - Increased resilience to network issues and JavaScript errors.
+  - Simplified testing of core functionality.
+  - Alignment with best practices for web development.
+
+- **Negative:**
+
+  - Increased development effort for implementing features in both a non-JS and JS version (though this is mitigated by focusing on core functionality first).
+  - Potential limitations on the complexity of interactive features in non-JS mode.
+  - Requires a consistent and disciplined approach to development and testing to ensure adherence to the progressive enhancement strategy.
```
chohner, line 48: Somehow the pro/con framing at the end feels out of place, but I don't know exactly why. Maybe because these aren't concrete `consequences` but rather feel like part of the discussion / context.

Since the pros are already mentioned there - maybe move the negative points up as `known challenges` and just remove the whole section here?

chohner (approved):
nice, just some minor comments!

doc/adr/0021-users-without-javascript.md
```diff
@@ -0,0 +1,48 @@
...
+Our framework Remix is built with progressive enhancement in mind, providing a functional experience even during JavaScript loading or failures.
+
+## Decision
+
+We will adopt a **Progressive Enhancement** strategy for all features.
+This means:
+
+1.  **Core Functionality First:** All essential features must be fully functional and accessible without JavaScript. This ensures that every user, regardless of their JavaScript capabilities, can accomplish the primary tasks the application is designed for.
+
+2.  **Enhancement with JavaScript:** JavaScript will be used to _enhance_ the user experience where appropriate. This allows us to add interactive elements, richer interfaces, and improved performance for users with JavaScript enabled. Examples include the auto-suggest input, dynamic form validation, and advanced UI components.
```
Spencer6497, line 29: without JavaScript, what happens when the user enters invalid data? for example, if they enter an option not included in an auto-suggest? or enters an invalid number? is that data saved?

Spencer6497 (approved):
LGTM 🌟 one question out of curiosity!

doc/adr/0021-users-without-javascript.md
chohner: We already run the same (or stricter, see post code) validation on the backend before saving (otherwise this would also be a huge security issue).

This also helps with the no-JS case, as [rvf](https://www.remix-validated-form.io/supporting-no-js) consumes the schema violations returned by zod + previously sent data and rebuilds the form. It works so magically that we don't have to think about it..

So: Data is always submitted via POST request on hitting the `SUBMIT` button, the server does the schema validation and redirects to a another page. On invalid data without JS, the same page is returned, this time with errors attached to the offending form fields.

doc/adr/0021-users-without-javascript.md
frederike-ramin: I have to say, I like the conciseness of the pro and con arguments here. If someone agrees with you, I'm happy to do what you suggested though.

doc/adr/0021-users-without-javascript.md
```diff
@@ -0,0 +1,48 @@
...
+## Context
+
+Currently, our application is built with the principle that core functionality should be accessible and usable even without client-side JavaScript.
+This aligns with our goal of ensuring a resilient and broadly accessible experience.
+
+We recognize that "non-JS users" aren't just those with JavaScript disabled, but also include users with slow connections, JavaScript errors, or while the JavaScript is still downloading ([Everyone has JavaScript, right?](https://www.kryogenix.org/code/browser/everyonehasjs.html)).
+However, we are now encountering situations where specific features, like the upcoming upload component or the auto-suggest input, require JavaScript to work intuitively.
+
+Our approach needs to address the balance between rich interactive features (requiring JavaScript) and the core usability for all users.
+The GOV.UK Service Manual provides guides on how to best approach this in the governmental area (see [here](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) and [here](https://www.gov.uk/service-manual/technology/using-progressive-enhancement)).
```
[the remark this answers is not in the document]
frederike-ramin: Moved it do decision. But I'm just gonna leave it as resources you can read upon further if you'd like to.

ingriddorioschulze (approved):

PR #1723, opened 2025-02-11
refactor(getXXXProps): migrate StrapiBox  

khallad2:


app/components/Box.tsx
```diff
@@ -8,7 +8,7 @@ export type BoxProps = {
   identifier?: string;
   label?: HeadingProps;
   heading?: HeadingProps;
-  content?: RichTextProps;
+  content?: RichTextProps | null;
```
Spencer6497, line 11: i think we're trying to avoid this

app/services/cms/models/StrapiBox.ts
```diff
@@ -1,36 +1,25 @@
...
 import { StrapiHeadingSchema } from "./StrapiHeading";
-import { StrapiParagraphSchema, getRichTextProps } from "./StrapiParagraph";
+import { getRichTextProps, StrapiParagraphSchema } from "./StrapiParagraph";
 
-const StrapiBoxSchema = z
+export const StrapiBoxSchema = z
   .object({
-    label: StrapiHeadingSchema.nullable(),
-    heading: StrapiHeadingSchema.nullable(),
+    __component: z.literal("page.box"),
```
Spencer6497, line 12: i believe this has to be inside the transform, as it doesn't exist in Strapi and in theory would break schema validation

app/services/cms/models/StrapiBox.ts
```diff
@@ -1,36 +1,25 @@
...
+    label: StrapiHeadingSchema.optional(),
+    heading: StrapiHeadingSchema.optional(),
```
Spencer6497, line 14: i think these should be defined as 
```suggestion
    heading: StrapiHeadingSchema.nullable().transform(omitNull).optional(),
```

Spencer6497 (changes requested):
Looking good! a few notes :)

app/services/cms/models/StrapiBox.ts
khallad2: i think we should use optional instead of converting to nullable and then transforming the null to undefined.
optional returns value | undefined and it looks more cleaner, from my POV

app/services/cms/models/StrapiBox.ts
khallad2: i think it's the same as before, we were extending the type by defining the __component the same way, also here we don't have to cast to `as const`

Spencer6497: i think there may be a misunderstanding here: anything that comes _before_ the `transform` _must_ match what's returned by strapi, otherwise schema validation will fail. the `__component` field is a field that's known only to our frontend, and not strapi, so it has to be added on _after_ the validation, inside the transform

app/services/cms/models/StrapiBox.ts
Spencer6497: unfortunately we have to define them first with `nullable`, as that's what exists in Strapi; Strapi has no conception of `optional`; that gets added on after schema validation + transformation

app/components/Box.tsx
khallad2: i think this is the cleanest way i found to match the types, as content is nullable.

pgurusinga: Why don't we omit the `null` option in the schema and make it optional like other keys? That way, we won't have to make this change.

PR #1724, opened 2025-02-11
refactor(getXXXProps): migrate StrapiHeader

khallad2:


app/components/Header.tsx
```diff
@@ -3,7 +3,7 @@ import RichText, { type RichTextProps } from "./RichText";
 
 export type HeaderProps = Readonly<{
   heading: HeadingProps;
-  content?: RichTextProps;
+  content?: RichTextProps | null;
```
aaschlote, line 6: I believe we decided to not use `null`, but transform the `null` values as `undefined`

aaschlote: you can use this schema StrapiRichTextOptionalSchema

PR #1731, opened 2025-02-12
feat: use translation for breadcrumb aria label

pgurusinga:


app/root.tsx
```diff
@@ -234,6 +234,7 @@ function App() {
             breadcrumbs={breadcrumbs}
             alignToMainContainer={header.alignToMainContainer}
             linkLabel={header.linkLabel}
+            translations={{ ...accessibilityTranslations }}
```
aaschlote, line 237: These translations are available using the hook `useTranslations` you can get the values like this:

```
const { accessibility: translations } = useTranslations();
```

So pass the translation via props redundant.

app/components/Breadcrumbs.tsx
```diff
@@ -10,26 +10,29 @@ export type Breadcrumb = {
...
 export default function Breadcrumbs({
+  translations,
   breadcrumbs,
   linkLabel,
   alignToMainContainer = true,
 }: BreadcrumbsProps) {
   const validBreadcrumbs = breadcrumbs?.filter(
     (breadcrumb) => breadcrumb.title !== undefined,
   );
+  const ariaLabelTranslation = translations?.["header-breadcrumb"];
```
aaschlote, line 27: what about to use the function `getTranslationByKey` to retrieve the translation?

app/components/Footer.tsx
```diff
@@ -45,7 +48,10 @@ export default function Footer({
           </div>
         </div>
 
-        <nav className="flex flex-wrap gap-x-32 gap-y-8" aria-label="Footer">
+        <nav
```
aaschlote, line 51: can we add an unit test for this as well?

aaschlote (commented):
nice work, I wrote a few comments!

app/root.tsx
pgurusinga: FYI, To make this custom hook work:
1. Wrap `<Breadcrumbs>` and `<Footer>` with Translation Context since they are only one level down.
2. Add `useMemo` to use `useTranslations` for the error page.
3. De-structure `videoTranslations`, `feedbackTranslations`, and `accessibilityTranslations` in `useMemo` for the error page to prevent TypeScript errors.

app/components/Footer.tsx
pgurusinga: what `this` are you referring to? if it's `nav` element, it's already tested

image

app/root.tsx
pgurusinga: is it worth an effort for 1 level passing the props?

aaschlote: I haven't noticed it, maybe we can keep passing as props

app/components/Footer.tsx
aaschlote: this is the `Breadcrumbs.tst` test, I mean here in this comment the `Footer.tsx`

aaschlote (approved):
LGTM!

PR #1742, opened 2025-02-14
Kontopfändung vorabcheck flow

khallad2:


app/services/featureFlags/index.ts
```diff
@@ -7,7 +7,8 @@ export type FeatureFlag =
   | "showProzesskostenhilfeFlow"
   | "showERV"
   | "showKopfzeile"
-  | "showBundID";
+  | "showBundID"
+  | "showKontopfändungFlow";
```
Spencer6497, line 11: we should probably enforce the following consistency to umlauts in strings:
* user-facing strings should always have umlauts
* code-only variables, strings, etc should use their vowel counterparts

app/domains/schulden/vorabcheck/xstateConfig.ts
```diff
@@ -0,0 +1,10 @@
+import type { Config } from "~/services/flow/server/buildFlowController";
+import { schuldenZwangsvollstreckungContext } from "./context";
+
+export const zwangsVollStreckungVorabcheckXstateConfig = {
```
Spencer6497, line 4: if we're keeping the name, `zwangsvollstreckung` should all be lowercase, as it's one word

app/domains/schulden/vorabcheck/context.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from "zod";
+import { YesNoAnswer } from "~/services/validation/YesNoAnswer";
+export const context = {
+  zwangsvollstreckung: YesNoAnswer,
```
Spencer6497, line 4: is this just a demo/dummy piece of context?

Spencer6497 (commented):
Looks good so far! Open question regarding naming though: I see three distinct names in the code:
* kontopfändung
* zwangsvollstreckung
* schuldenZwangsvollstreckung
Which is the right name for this feature? we should probably be consistent in our naming, as I found it sort of hard to understand. I suggest just `kontopfaendung`

app/domains/schulden/vorabcheck/context.ts
khallad2: Yes it's the initial step for the flow with handling flowIds more than 2 levels, added description to the PR.

app/services/flow/server/__test__/flowTransitionValidation.test.ts
```diff
@@ -35,6 +35,7 @@ describe("flowTransitionValidation", () => {
     "/fluggastrechte/vorabcheck": createMockFlow("vorabCheck"),
     "/fluggastrechte/formular": createMockFlow("formFlow"),
     "/prozesskostenhilfe/formular": createMockFlow("formFlow"),
+    "/schulden/kontopfaendung/wegweiser": createMockFlow("formFlow"),
```
Spencer6497, line 38: i think this is technically a `vorabCheck`

Spencer6497 (commented):
Before approving this, I really want us to nail the naming here. @frederike-ramin i see we're referring to this primarily as `schuldenKontopfaendungWegweiser` in the code, and while that mirrors the url, I'm not sure its the most accurate and readable. This is technically a vorabcheck, no? why not just `kontopfaendungVorabcheck`? thoughts? feelings? am I insane?

app/services/flow/server/__test__/flowTransitionValidation.test.ts
frederike-ramin: Haha, "insane" - no. :D 
Technically it is a vorabcheck, so whenever the decision is "am I using formFlow or vorabCheck" (like in the example here) the answer would be "vorabCheck". However, the naming I like being "wegweiser" everywhere, so we all use the same wording(esp. together with the designers).

I summary, I think we should
- go for "vorabCheck" where there is a specific distinction between "formFlow" and "vorabcheck" and
- go for "wegweiser" everywhere else.

Either way, formFlow is not correct in this context anywhere^^

@khallad2 I believe it's probably not well explained what the difference between "formFlow" and "vorabcheck" are. If you have questions, don't hesitate to ask and I'd recommend the following [ADR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/doc/adr/0012-domain-oriented-architecture.md) for context.

khallad2:
Here are my thoughts, please correct me if i am wrong.

if there is a possibility to have something like `schulden/Kontopfaendung/something-else` in the future, that means  `schuldenKontopfaendungWegweiser` or `schuldenKontopfaendungWegweiserVorabcheck` make sense because it describes the relation to `KontopfaendungWegweiser`.

if this is the only `vorabcheck` we will have in `schulden`, like there is no possibility to have  `schulden/Kontopfaendung/something-else` then we can change it to `schuldenVorabcheck`

app/services/featureFlags/index.ts
```diff
@@ -7,7 +7,8 @@ export type FeatureFlag =
   | "showProzesskostenhilfeFlow"
   | "showERV"
   | "showKopfzeile"
-  | "showBundID";
+  | "showBundID"
+  | "showKontopfaendungWegweiserFlow";
```
Spencer6497, line 11: don't forget to update the feature flag in PostHog before you merge :)

Spencer6497 (approved):
excellent work!

PR #1749, opened 2025-02-17
feat: add country to FGR PDF

pgurusinga:
this includes the country of the airlines, the claiming person and other plaintiff


aaschlote (approved):
LGTM!

PR #1756, opened 2025-02-18
Build kontopfändung basisinfo flow

khallad2:
Create states for `Abfrage Basisinfos`


app/domains/schulden/vorabcheck/xstateConfig.ts
```diff
@@ -14,41 +14,85 @@ export const kontopfaendungWegweiserXstateConfig = {
...
-          {
-            target: "pKonto",
-            guard: ({ context }) => context.hasKontopfaendung === "maybe",
-          },
         ],
         BACK: "start",
       },
     },
-    Ergebnisseite: {
+    ergebnisseite: {
```
Spencer6497, line 28: I believe we're missing [this](https://www.figma.com/design/5HvORPPZQTQaJrXtYC3cgB/Zwangsvollstreckung?node-id=1008-28903&t=42LWF3m9RejajYCU-4) page

app/domains/schulden/vorabcheck/xstateConfig.ts
Spencer6497, line 1: so instead of having a structure like this:
```
{
    "start": {},
    "kontopfaendung": {},
    "ergebnisseite": {}
}
```

with all of the states as siblings, I would group all of them into a top level state, abfrageBasisinfos, like so:

```
{
    "abfrageBasisInfos": {
        "states": {
            "start": {},
            "kontopfaendung": {},
            "ergebnisseite": {}
        }
    }
}
```

that way, not only is the vorabcheck easier to read, but also, if/when we decide to add a sectioned progress bar later, it's much easier

app/domains/schulden/vorabcheck/context.ts
```diff
@@ -1,8 +1,20 @@
...
+    "privatpersonen",
+    "behörden",
+    "nicht-zurückzahlen",
+    "finanzamt",
+    "beitragsservice",
+    "unterhaltsschulden",
+    "strafe-nicht-zahlen",
+    "rechnungen-nicht-bezahlen",
+    "glaeubigerin-unbekannt",
+  ]),
```
Spencer6497, line 16: tricky one but here is somewhere I'd actually invoke an [english translation](https://digitalservicebund.atlassian.net/wiki/spaces/DIGITALSER/pages/1509425158/Language+in+Code+using+German+or+English+words#Our-Solution%3A-The-Hybrid-Approach) (not only to get rid of the umlauts in the enum, which I think may cause problems if we try to infer its type, but also for clarity). I think we should keep the proper, more-or-less untranslateable nouns (like Finanzamt, Beitragsservice), and translate the rest

Also, I think we're missing 3 options

Spencer6497 (changes requested):
looks good so far!

app/domains/schulden/vorabcheck/context.ts
khallad2: **English Translation:** 
- i think the english translation will be even more confusing here also i think it's a domain specific. 
- so it maps perfectly to the design and provide more consistency.
- so if there is no severe issue i would like to keep it in german
**Missing options:** 
the design is being updated and not finalized yet. (it will be a constant update until design team finishes)

app/domains/schulden/vorabcheck/xstateConfig.ts
khallad2: can you please clarify which page is missing **ergebnisseite** or the **PKonto** page you've mentioned in the link?

- ergebnisseite is a [dummy page exists here](https://a2j-rast-strapi.dev.ds4g.net/admin/content-manager/collection-types/api::vorab-check-page.vorab-check-page/fhjvaz1b4l6wowvtk82wmkse?plugins%5Bi18n%5D%5Blocale%5D=sg) for now and it's not included in the scope of this ticket, Agreed with Frede and Nicolai on the ticket [here](https://app.asana.com/0/0/1209421982081871/f)  

- In the link you've mentioned it refers to `pKonto` state and it's page [exists here](https://a2j-rast-strapi.dev.ds4g.net/admin/content-manager/collection-types/api::vorab-check-page.vorab-check-page/otzgut8kj34htema1nterj6n?plugins%5Bi18n%5D%5Blocale%5D=sg) .

app/domains/schulden/vorabcheck/context.ts
Spencer6497: i think there may be a misunderstanding regarding "domain specific"; my understanding was as follows:
* terms like "Kontopfändung" or "pKonto" are "domain-specific" in the sense that their only possible context is in the domain, and translating them would make no sense
* terms like "nicht-zurückzahlen" (which, after looking at the Figma, should probably be more explicit as "Kredit": _what exactly_ isn't zurück-zahl-able?) are both general and easily exchanged with their english equivalents -- in this case, 'credit'

though i could be misunderstanding what @frederike-ramin meant, and would be curious to hear her thoughts.

Also, after looking at the Figma, it seems as if the design itself (i.e. number of options) is finalized and "Ready for Dev", but the designers are still working on the copy, unless there's some context I'm missing?

Spencer6497: overall, i would suggest
```suggestion
  debtSource: z.enum([
    "private",
    "authorities",
    "credit",
    "finanzamt",
    "beitragsservice",
    "supportPayments",
    "fine",
    "unpaidBill",
    "unknown",
  ]),
```

app/domains/schulden/vorabcheck/xstateConfig.ts
Spencer6497: ah ok sorry I didn't see it in the xState. could we move the pKonto entries above ergebnisseite so this file reads like the logical flow?

also, I just noticed that hasPKonto should actually be an enum instead of a yesNoAnswer, as 4 different answers are possible.

app/domains/schulden/vorabcheck/context.ts
frederike-ramin: I'm almost regretting this language-in-code thing a bit. The main purpose was: "Don't overthink it."
I personally am leaning towards all-German words. But that's just personal preference. I think both options are fine. I would probably stick to "schulden" (as we also use it in the url) and the "kredit" distinction seems important in contrast to "nicht-zurückzahlen" semantically but otherwise, I wouldn't overthink it.

app/domains/schulden/vorabcheck/xstateConfig.ts
khallad2: but the 4 options actually referring to only 2 answers, if you follow the arrows in Figma you find the first two options are leading to the same result `yes` and target the [next page](https://www.figma.com/design/5HvORPPZQTQaJrXtYC3cgB/Zwangsvollstreckung?node-id=1008-29273&t=SAKjVUsm4I0XOetW-0)  , also the second two options are referring to also the same result `no` and target the [sofort info page](https://www.figma.com/design/5HvORPPZQTQaJrXtYC3cgB/Zwangsvollstreckung?node-id=1008-28904&t=SAKjVUsm4I0XOetW-0).

app/domains/schulden/vorabcheck/context.ts
khallad2: okay then we stick with German and schulden for now. 
 updating `nicht-zurückzahlen`.

app/domains/schulden/vorabcheck/context.ts
```diff
@@ -1,8 +1,26 @@
...
+  "gesetzlichenKrankenkasse",
+  "rechnungenNichtBezahlen",
+  "beitragsserviceGEZ",
+  "unterhaltsschulden",
+  "strafeNichtZahlen",
+  "finanzamt",
+  "hauptzollamt",
+  "nichtSagen",
+  "weissNicht",
+]);
```
Spencer6497, line 17: i would still shorten and further clarify like so:

```suggestion
export const schuldenType = z.enum([
  "privat",
  "behoerden",
  "kredit",
  "krankenkasse",
  "rechnung",
  "beitragsservice",
  "unterhalt",
  "strafe",
  "finanzamt",
  "hauptzollamt",
  "nichtSagen",
  "weissNicht",
]);
```

app/domains/schulden/vorabcheck/xstateConfig.ts
Spencer6497: the 4 options refer to 4 different _answers_, even if they only end up in 2 possible _outcomes_. The 4 different options are actually distinct from one another:

1. No pKonto
2. Has pKonto (active)
3. Inactive pKonto
4. Applied for pKonto

Oftentimes content designers will use these answers later to tailor content pages, or even alter flow logic. also we want our code to mirror the underlying business logic, meaning the answers (options) are separate from the outcomes (guards/result pages)

app/domains/schulden/vorabcheck/xstateConfig.ts
khallad2: nice note, thanks!

app/domains/schulden/vorabcheck/xstateConfig.ts
khallad2: That makes sense, Thanks for the explanation.
changed all to enum since they are more than 2 options.

PR #1761, opened 2025-02-20
 Suggested fix for sentry syntax error - No identifiers allowed d…

khallad2:
referring to this [issue on sentry](https://digitalservice.sentry.io/issues/6300607414/events/e7ae6a1ff7394055a2011151e2ed82a3/)

we have a syntax error SyntaxError in `routes/beratungshilfe._index`
Error `No identifiers allowed directly after numeric literal`

**Context:**
The error is caused by the file named `$.tsx` when referenced directly in the re-export line. 

**Fix**
By renaming it to something like `index-content.tsx` and updating the import, we avoid such minification issue.


Spencer6497:
I'm not entirely sure about this fix, as it's the idiomatic way for Remix to determine [dynamic segments](https://remix.run/docs/en/main/file-conventions/routes#dynamic-segments)

chohner:
Agreed, we def need the `$` route, which is used to dynamically show pages with arbitrary slugs from the CMS. Not only would this feature disappear - adding a new `index-context.tsx` file in the `/routes` folder also creates the `/index-context` route.

We could move the shared loader & component to a different file and import them into all routes, including `$.tsx`

pgurusinga:
I am unsure if this is related to the `$` of Remix, there might be an internal issue with the JS compiler 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Identifier_after_number

khallad2:
closed as the naming needs `$` according to [Chris comment](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1761#issuecomment-2678397673)

PR #1774, opened 2025-02-24
feat: markdown utiliy should not remove interpolated url

pgurusinga:
### Problem
The court website link is not working because the URL is interpolated with curly braces. Due to recent changes, the sanitizer removes the value of the `href` attribute.

### Proposed Solution
- Ignore anchor tags (`<a>`) where the `href` value contains curly braces.
- For external URLs containing curly braces, include an external link icon.


app/components/StandaloneLink.tsx
```diff
@@ -17,7 +17,9 @@ export const StandaloneLink = ({
   icon,
   className,
 }: StandaloneLinkProps) => {
-  const shouldOpenNewTab = isExternalUrl(url) || isFileDownloadUrl(url);
+  const interpolatedExternalUrl = url.startsWith("{") && url.endsWith("}");
```
chohner, line 20: could we re-use the same `PLACEHOLDER_REGEX` here and in the sanatizer?

chohner: also - by changing the order of the rendering pipeline, we loose insight into the URL content here. another reason to move the string replacements up in the chain 🙈

chohner (approved):

Spencer6497 (approved):

PR #1784, opened 2025-02-25
feat/responsiveness of header

judithmh:
solves issue where link to Gebärdensprache is cut off in header [ticket](https://app.asana.com/0/1207022303033556/1209248472068595)


app/components/PageHeader.tsx
```diff
@@ -18,6 +18,10 @@ type PageHeaderProps = {
   };
 };
 
+const decodeHTMLSoftHyphen = (html: string) => {
+  return html.replace(/&shy;/g, "\u00AD");
```
pgurusinga, line 22: why do we need this? by default, `&shy;` is already working and we use it in several places 
image

judithmh: But it was a good remark - it usually works because most of the time it's rich text. However, in some fields it's also a heading and doing research I figured that we have already a dependency for the decoding so I changed the function from the custom function to the one we are already using somewhere else

pgurusinga (approved):

PR #1785, opened 2025-02-26
Kontopfaendung updates

khallad2:
- remove guard double checks
- remove unit tests


app/domains/schulden/vorabcheck/xStateConfig.ts
```diff
@@ -47,9 +45,6 @@ export const kontopfaendungWegweiserXstateConfig = {
           },
           {
             target: "kontopfaendung",
-            guard: ({ context }) =>
-              context.hasKontopfaendung === hasKontopfaendung.Values.nein &&
```
khallad2, line 51: @Spencer6497 This approach ensures that 
- The code stays consistent and type-safe across your application. 
- If the underlying enum value ever changes, you'll only need to update the enum definition instead of hunting down every literal string occurrence. 
- Additionally, referencing the enum centralizes the source of truth for valid values, reducing the chance of typos and discrepancies

Spencer6497: I just want to reiterate that:
* the code is already consistent across the entire codebase and type-safe. if you try to change the comparison string value to a non-enum value, typescript will throw an error (See screenshot below). This introduces a new pattern and more LoC
* If the underlying enum changes, in either pattern, we'll have to hunt down and correct all of the now-invalid values. again, the code is already type-safe
* The source of truth is the enum in both cases. These aren't magic strings, they're string literals, if you try to set them to invalid values or make a typo, TypeScript will complain

https://github.com/user-attachments/assets/b5c89586-5330-483d-86a5-ff6b2f387d82

app/domains/schulden/vorabcheck/xStateConfig.ts
```diff
@@ -65,9 +60,6 @@ export const kontopfaendungWegweiserXstateConfig = {
           },
           {
             target: "glaeubiger",
-            guard: ({ context }) =>
-              context.hasPKonto === hasPKonto.Values.nein ||
-              context.hasPKonto === hasPKonto.Values.ja,
           },
```
Spencer6497, line 63: instead of an object with a target, it can be even simpler like so: 
```
SUBMIT: [
    {
        target: "p-konto-probleme",
        guard: ({ context }) =>
            context.hasPKonto === hasPKonto.Values.bank ||
            context.hasPKonto === hasPKonto.Values.nichtAktiv,
    },
    "glaeubiger"
]
```

Spencer6497 (approved):

PR #1789, opened 2025-02-26
refactor: remove field identifier and use inline items directly

pgurusinga:


app/components/summaryOverview/SummaryOverviewBoxArray.tsx
```diff
@@ -20,19 +19,14 @@ const SummaryOverviewBoxArray = ({ boxId, boxItems, title, stepId }: Props) => {
...
 
-  const arrayBoxPageFieldsWithoutArrayChar = boxItems.map((items) => ({
-    ...items,
-    field: items.field.split(arrayChar)[1],
-  }));
-
   return arrayObject.map((object, index) => (
     <SummaryOverviewBox
       // eslint-disable-next-line react/no-array-index-key
       key={boxId + index}
```
judithmh, line 25: key={`${boxId}-${title?.text ?? index}` } maybe we can use this here then we can get rid of the eslint-disable comment

frederike-ramin (commented):
Love it! Just once suggestion to make it even better in my view. Thank you!

app/components/summaryOverview/SummaryOverviewBoxArray.tsx
pgurusinga: thanks for suggestion but with title.text is not unique :(
this is based on our overview
image

judithmh (approved):

frederike-ramin (approved):
Yes, love it! 💯 (also great new name for the emptyValue thing 👌 )

PR #1795, opened 2025-02-27
feat(heading): html decode on parse

chohner:
Looking at https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1784/files, I realized we still have multiple places where we use `html-entities` in the front-end.

This PR moves them into the `schema.transform()`, resulting in a nice reduction in bundle size:

- before: 414.51 kB gzipped
- after: 387.62 kB gzipped


app/services/cms/models/__test__/stringWithHtmlEntities.test.ts
```diff
@@ -0,0 +1,7 @@
+import { stringWithHtmlEntities } from "../stringWithHtmlEntities";
+
+test("stringWithHtmlEntities should decode html entities", () => {
+  const stringWithEntities = "&lt;h1&gt;test&lt;/h1&gt;";
+  const actual = stringWithHtmlEntities.safeParse(stringWithEntities);
+  expect(actual).toEqual({ data: "<h1>test</h1>", success: true });
+});
```
pgurusinga, line 7: Can we add a test for a string with HTML entity currently we use? e.g `&shy;`  or `funktioniert&#x27;s`
image
 🙏

PR #1800, opened 2025-02-28
Kontopfaendung restructure routes

khallad2:
There is no Schulden page anymore so we need to
- change routes and breadcrumbs from `/schulden/kontopfaendung/wegweiser` to `/kontopfaendung/vorabcheck`
- enable clickable parent in breadcrumbs


chohner (approved):
lgtm!

PR #1802, opened 2025-02-28
ZOV: linear personal data

pgurusinga:
when klagende person is unfilled


app/domains/fluggastrechte/formular/persoenlicheDaten/flow.json
```diff
@@ -30,7 +33,10 @@
                 "target": "#prozessfuehrung.zeugen"
               }
             ],
-            "BACK": "#person.daten"
+            "BACK": {
+              "guard": "personDone",
```
judithmh, line 37: just out of interest: why is the guard here necessary?

app/domains/fluggastrechte/formular/persoenlicheDaten/guards.ts
```diff
@@ -8,6 +8,7 @@ export const persoenlichDatenGuards = {
     context: { pageData, weiterePersonen },
   }) => isValidArrayIndex(weiterePersonen, pageData),
   ...yesNoGuards("isWeiterePersonen"),
+  personDone,
   persoenlichenDatenDone: ({ context }) =>
     personDone({ context }) && weiterePersonenDone({ context }),
```
judithmh, line 13: do we still need the personDone({ context }) here?

pgurusinga: true, I removed it

app/domains/fluggastrechte/formular/persoenlicheDaten/flow.json
pgurusinga: I think it's not anymore

judithmh (approved):

PR #1807, opened 2025-03-04
fix(Logo missing for non-js users)

ingriddorioschulze:
This PR enables no-js users to have the logo rendered at the page and footer.

Findings:

- react-inlinesvg does not work for non-js users because it requires React to run
- returning a version for non-js users raised another bug: 

1. the browser load the image 2 times: first the `` tag and then the` <Svg> `from react-inlinesvg
2.  the first response (with the `` tag) is cached by the browser
3. the same response is used again for the second request (with the `<Svg>` tag)
4. they are not compatible and the CORS headers are missing
Screenshot 2025-03-04 at 07 14 09

My research links: https://github.com/tanem/react-svg/issues/197 and https://serverfault.com/a/856948

Solution:

- the solution with the `<noscript>` tag prevents the first request `` to happen when javascript enabled is
- if !svg an `` will be rendered
- if !jsAvailable an `` will be rendered inside a `<noscript>` tag to handle the Chrome bug


app/components/Image.tsx
```diff
@@ -9,23 +10,37 @@ export type ImageProps = Readonly<{
...
-    />
-  ) : (
-    <img {...props} src={url} alt={alternativeText ?? ""} />
-  );
+  if (isSvg) {
+    if (jsAvailable) {
+      return (
+        <Svg
+          {...props}
+          id="svg-image"
```
chohner, line 27: I think we should remove this `id`, since `Image` can be used multiple times per page

app/components/Image.tsx
```diff
@@ -9,23 +10,37 @@ export type ImageProps = Readonly<{
...
+        />
+      );
+    } else {
+      return (
+        <noscript>
+          <img {...props} src={url} alt={alternativeText ?? ""} />
+        </noscript>
+      );
+    }
+  } else {
```
chohner, line 41: Maybe 3 "early" returns are more readable?

```
if (!isSvg){
return <img ...>
}

if (!jsAvailable){
<noscript>...
}

return <Svg ...
```

app/components/Image.tsx
```diff
@@ -9,23 +10,37 @@ export type ImageProps = Readonly<{
...
+          id="svg-image"
+          src={url}
+          title={svgAltText}
+          role="img"
+          height="100%"
+        />
+      );
+    } else {
+      return (
+        <noscript>
```
chohner, line 36: oh, could you add a (short) code comment of why this is necessary? Could include links to external references etc

app/components/Image.tsx
```diff
@@ -9,23 +10,45 @@ export type ImageProps = Readonly<{
...
+      <Svg
+        {...props}
+        id="svg-image"
+        src={url}
+        title={svgAltText}
+        role="img"
+        height="100%"
+      />
+    );
+  };
```
chohner, line 33: I think this can just directly return at the end?

chohner: (also, there's still an `id` here

app/components/Image.tsx
```diff
@@ -9,23 +10,45 @@ export type ImageProps = Readonly<{
...
+
+  const renderImage = () => {
```
chohner, line 35: I don't think this needs to be a function? something like

```typescript
const ImageTag = <img ... />

if (!isSvg) return <ImageTag />
if (!jsAvailable) return <noscript><ImageTag /></noscript>
return <Svg ... />
```

frederike-ramin (commented):
Thanks for taking this on! I have to admit, I'm really confused why we even use this `react-inlinesvg` in the first place. Why not just render a normal `<svg>`? I'm sure I'm missing something here, but do you know the answer @ingriddorioschulze?

ingriddorioschulze:
> Thanks for taking this on! I have to admit, I'm really confused why we even use this `react-inlinesvg` in the first place. Why not just render a normal `<svg>`? I'm sure I'm missing something here, but do you know the answer @ingriddorioschulze?

Inline SVGs are much better to work with accessibility and react-inlinesvg library makes it easier to have inline svgs. The only problem we have at the moment is to display the logo for no-js users in a high contrast friendly way. For no-JS users we are rendering an  and not a inline svg.

app/components/Image.tsx
ingriddorioschulze: @chohner Sorry, Chris but the id has to stay. I added a comment in the code :)

app/components/Image.tsx
```diff
@@ -9,22 +11,65 @@ export type ImageProps = Readonly<{
...
+          <img
+            className={imageClasses}
+            {...props}
+            src={url}
+            alt={altText}
+            title={altText}
+          />
+        </div>
+      </noscript>
+    );
```
Spencer6497, line 60: since these are identical blocks (aside from the `noscript` tags), I would prefer something like a minimal wrapper component, like so: 
Bildschirmfoto 2025-03-06 um 08 22 45

Spencer6497 (approved):
great work!

app/components/Image.tsx
ingriddorioschulze: Thanks for the suggestion, @Spencer6497 ! I did it something like that before and I got a feedback that it might not be that readable. Could we keep like this for now or do you have strong opinions?

PR #1817, opened 2025-03-05
remove hardcoded SummaryPage

judithmh:
deletion of the hardcoded version of the summaryPage


aaschlote (commented):
if you want, you can remove these files:

api.airline.$iataCode.ts
delay-flight-airports.ts
step-over-airports.ts

They were created only to retrieve data from hard code version of the summary component, we won't need them anymore.

aaschlote (approved):
LGTM!

PR #1822, opened 2025-03-06
fix(Posthog Capture Text): add maskAllInputs to session record for maximum privacy

ingriddorioschulze:
This PR implements maskAllInputs to session recording in posthog init. It prevents sensitive data being shown on pages

### DIFF
--- a/app/components/cookieBanner/CookieBanner.tsx
+++ b/app/components/cookieBanner/CookieBanner.tsx
@@ -39,8 +39,9 @@ export function CookieBanner({
         api_host: POSTHOG_API_HOST,
 
         session_recording: {
-          // Masking text elements to prevent sensitive data being shown on pages
+          // Masking input and text elements to prevent sensitive data being shown on pages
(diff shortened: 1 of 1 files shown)


frederike-ramin (approved):
Fascinating. How did we do it before? Because I think the inputs were in fact masked?

josh-nowak:
@frederike-ramin Yes, input fields were masked before! PostHog does this by default. The two options interacted in a somewhat weird way: Setting the `maskTextSelector` apparently disabled the default input masking and required us to set `maskAllInputs` explicitly.

ingriddorioschulze:
> Fascinating. How did we do it before? Because I think the inputs were in fact masked?

There was no specification for session_recording on posthog.init. In the documentation says they mask input elements by default, if needed, you can set it to false and specify inputs to be masked.
Just masking the text might don't help with the inputs elements. With this approach we go for maximum privacy.

frederike-ramin:
Faaaascinating, thank you for the context!

PR #1838, opened 2025-03-13
Create accordion component in list item

khallad2:
**Behaviour:**
- list of closed details components (1..n)
- clicking on a headline reveals the previously hidden description
- with JS: max 1 open
- withoutJS: multiple can be open
- The accordion component itself contains 1 ... N AccordionItems
- Props:
  - Title: string (mandatory);
  - description: markdown (mandatory)

**Acceptance Criteria:**
    1. Add new Accordion component in app
    2. Add component to storybook
    3. Strapi: optional accordion in ListItem
    4. App: StrapiListItem schema is adapted


app/components/ListItem.tsx
```diff
@@ -27,24 +29,32 @@ const ListItem = ({
...
-            ))}
-          </ButtonContainer>
-        )}
-      </div>
-    </div>
+      {accordion && (
+        <div className="m-16">
+          <Accordion {...accordion} />
+        </div>
+      )}
```
chohner, line 56: why would this not be part of the div?

app/services/cms/models/StrapiListItem.ts
```diff
@@ -15,7 +16,7 @@ export const StrapiListItemSchema = z
       .nullable()
       .transform(omitNull)
       .optional(),
+    accordion: StrapiAccordionSchema.optional().nullable().transform(omitNull),
```
chohner, line 19: Strapi never returns `undefined`, so the `.optional()` can be removed

app/services/cms/models/StrapiContentComponent.ts
```diff
@@ -29,6 +30,7 @@ export const StrapiContentComponentSchema = z.union([
   StrapiDetailsSchema,
   StrapiUserFeedbackSchema,
   StrapiSummaryOverviewSectionSchema,
+  StrapiAccordionSchema,
```
chohner, line 33: technically we don't need to add this here, since we don't allow content editors to add this to arbitrary pages

app/services/cms/models/StrapiAccordion.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const StrapiAccordionSchema = z
+  .object({
+    items: z.array(StrapiAccordionItemSchema),
+  })
+  .merge(HasOptionalStrapiIdSchema)
+  .transform((cmsData) => ({
+    __component: "page.accordion" as const,
+    ...cmsData,
+  }));
```
chohner, line 17: I think this could just be one schema without a transform?

app/components/PageContent.tsx
```diff
@@ -78,6 +79,8 @@ function cmsToReact(strapiContent: StrapiContentComponent) {
       return <UserFeedback {...strapiContent} />;
     case "page.summary-overview-section":
       return <SummaryOverviewSection {...strapiContent} />;
+    case "page.accordion":
+      return <Accordion {...strapiContent} />;
```
chohner, line 83: Since we don't allow accordions to be placed on a page we wouldn't need to introduce them here

app/components/Accordion.tsx
```diff
@@ -0,0 +1,53 @@
...
+              <details>
+                <summary className="cursor-pointer text-lg font-medium py-4 px-2">
+                  {item.title}
+                </summary>
+                <div className="p-4">{item.description}</div>
+              </details>
+            </div>
+          ))}
+        </section>
+      )}
```
chohner, line 50: I think this can be much more DRY with just a change in what the `.map()` returns

chohner: in fact - why do we even need to have to differentiate here at all? the only JS we need should be closing other items on click? 🤔

app/components/ListItem.tsx
khallad2: because of design issues (trying to match the design on Figma) without updating the listItem styles.

app/services/cms/models/StrapiListItem.ts
khallad2: for some reason all e2e tests are failing when there is no .optional(), looking to it.

app/components/Accordion.tsx
```diff
@@ -0,0 +1,40 @@
+import { useState, useEffect } from "react";
+import AccordionItem, { AccordionItemProps } from "~/components/AccordionItem";
+
+export type AccordionProps = Readonly<{
+  items: Array<Omit<AccordionItemProps, "isOpen" | "onToggle">>;
```
Spencer6497, line 5: why do we explicitly need to omit these props?

app/components/Accordion.tsx
```diff
@@ -0,0 +1,40 @@
...
+  className?: string;
+}>;
+
+export default function Accordion({ items, className }: AccordionProps) {
+  const [jsEnabled, setJsEnabled] = useState<boolean>(false);
+  const [openIndex, setOpenIndex] = useState<number>(-1);
+  const hasItems = items && items.length > 0;
```
Spencer6497, line 12: is it even possible or desireable that we allow Accordions without any items?

app/components/Accordion.tsx
```diff
@@ -0,0 +1,40 @@
...
+                setOpenIndex((prev) => (prev === index ? -1 : index))
+              }
+              jsEnabled={jsEnabled}
+            />
+          ))}
+        </section>
+      )}
+    </>
+  );
+}
```
Spencer6497, line 40: do we need to wrap this in a `<noscript>` tag in the event of no-JS to prevent the problem with rendering flickering? I know we had to do something similar with `FilesUpload`

app/components/Accordion.tsx
khallad2: because these props are meant to be managed internally by the Accordion component rather than supplied from external data.

the parent Accordion component controls the open/close state of each AccordionItem, and those values (and the function to toggle them) should be computed based on internal state

- Consumers of the Accordion provide only the static data (like title and description) and not the interactive state, which is fully controlled by the Accordion.

- Using Omit<AccordionItemProps, "isOpen" | "onToggle"> enforces that the consumer doesn't accidentally pass in these stateful props. This prevents conflicts between externally provided values and the internal logic of the component.

- It guarantees that the Accordion component is responsible for handling which item is open (via its own openIndex state) and how toggling happens, leading to predictable, centralized behavior.

app/components/Accordion.tsx
khallad2: ListItem has optional attribute accordion which is also nullable so if the is no items in the accordion 

the listItem comes from Strapi 
```
{
  id: 1072,
  identifier: 'Item1',
  content: 'item 1',
  headline: null,
  accordion: null,
  buttons: []
}
```

Spencer6497: but in ListItem we're already conditionally rendering an `<Accordion />` if it exists... meaning _if_ an `accordion` exists, it _will_ have at least one `AccordionItem`

app/components/Accordion.tsx
Spencer6497: how would conflicts arise between the passed-in values if we're never using them later-on? we're only using `id`, `title` and `description`

khallad2: yeah you are right conflicts would arise if i am using them, i was thinking about it at the beginning but i was not sure what exactly can be better in this case, but since it's still in draft and not finished yet, it would be part of the cleaning before i open the PR for review.

app/components/Accordion.tsx
khallad2: Since the PR still in Draft it maybe confusing at the moment, but i expect to remove the conditional rendering of the Accordion bec it's not necessary anymore.

app/services/flow/server/buildFlowController.ts
```diff
@@ -157,7 +157,6 @@ function stepStates(
 
   const statesWithDoneFunctionOrSubstates = Object.values(
     stateNode.states ?? {},
-    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
   ).filter((state) => state.meta?.done || Object.keys(state.states).length > 0);
```
khallad2, line 160: this changes doesn't belong to the PR, it came from merging main into current branch, it was mandatory to fix in order to pass eslint.

app/components/Accordion.tsx
```diff
@@ -0,0 +1,30 @@
...
+  }, []);
+
+  return (
+    <section>
+      {items.map((item, index) => (
+        <AccordionItem
+          key={item.id ?? index}
+          title={item.title}
+          description={item.description}
+          isOpen={openIndex === index}
```
pgurusinga, line 23: I'm curious about what our requirement is. Let's say the design team wants to create a set of collapsible, and one of them should always be open by default without any user interaction (configurable via Strapi)

pgurusinga (commented):
nice one! also works in high-contrast modes!

Bildschirmfoto 2025-03-17 um 17 56 46
Bildschirmfoto 2025-03-17 um 17 56 38

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
+import KeyboardArrowDownIcon from "@digitalservicebund/icons/KeyboardArrowDown";
+import KeyboardArrowUpIcon from "@digitalservicebund/icons/KeyboardArrowUp";
+import RichText from "~/components/RichText";
+import { useTranslations } from "~/services/translations/translationsContext";
+
+export type AccordionItemProps = Readonly<{
+  title: string;
+  description: string;
+  id?: number;
```
chohner, line 9: why the id?

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+            className="ds-body-02-reg px-16 pt-16 pb-24 gap-y-32"
+            html={description}
+          />
+        )}
+      </div>
+    );
+  }
+
+  return (
+    <details key={id} className="group border-2 rounded border-blue-500">
```
chohner, line 57: I dont think we need to specify the key here

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default function AccordionItem({
+  title,
+  description,
+  isOpen,
+  onToggle,
+  jsEnabled,
+  id,
+}: AccordionItemProps) {
+  const { accordion } = useTranslations();
+  if (jsEnabled) {
```
chohner, line 24: I was really hoping we could only ship one `<details>` component and then close stuff with some extra JS. I will focus the review on the noJs part for now just in case thats still possible

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <button
+          type="button"
+          onClick={onToggle}
+          aria-expanded={isOpen}
+          className="w-full flex justify-between text-left cursor-pointer text-lg gap-x-8 bg-blue-100"
+        >
+          <span className="p-16 ds-label-01-bold">{title}</span>
+          <span className="p-16 text-blue-800 flex items-center">
+            {isOpen ? (
+              <>
```
chohner, line 36: the fragments seem unecessary?

app/components/Accordion.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default function Accordion({ items }: AccordionProps) {
+  const [jsEnabled, setJsEnabled] = useState<boolean>(false);
+  const [openIndex, setOpenIndex] = useState<number>(-1);
+
+  useEffect(() => {
+    setJsEnabled(true);
+  }, []);
+
+  return (
+    <section>
```
chohner, line 17: I'm not sure whether `<section>` is the right element here 🤔

chohner: additionally: I think the rounded border should exist on this element, not each element

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <span className="flex group-open:hidden">
+            <KeyboardArrowDownIcon /> {accordion.accordionItemShow}
+          </span>
+          <span className="hidden group-open:flex">
+            <KeyboardArrowUpIcon /> {accordion.accordionItemHide}
+          </span>
+        </span>
+      </summary>
+      <RichText
+        className="ds-body-02-reg px-16 pt-16 pb-24 gap-y-32"
```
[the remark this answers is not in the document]
khallad2: because it's mentioned in the design that the description has gap 32.

app/components/Accordion.tsx
khallad2: if the rounded borders exists only in the parent then we there is no border on every element.
this is to match the design.

app/components/__test__/Accordion.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+    const toggleButtons = screen.getAllByTestId("accordion-toggle");
+    const descriptions = screen.getAllByTestId("accordion-description");
+
+    descriptions.forEach((desc) => {
+      expect(desc.textContent).toBe("");
+    });
+
+    fireEvent.click(toggleButtons[0]);
+    expect(descriptions[0].textContent).toBe("Description 1");
+    expect(descriptions[1].textContent).toBe("");
```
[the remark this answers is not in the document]
khallad2: because the received value must be an HTMLElement or an SVGElement.

app/domains/fluggastrechte/services/pdf/sections/reason/factsOfCases/table/getConnectionDetails.ts
```diff
@@ -111,7 +111,6 @@ const getConnectionDetailsCancel = (
   return {
     info: `${announcementDescription}\n${replacementFlightDescription}`,
     timeTable: [
-      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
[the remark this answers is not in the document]
khallad2: this is in an old commit, it doesn't exist anymore.

app/components/ListItem.tsx
```diff
@@ -27,23 +29,31 @@ const ListItem = ({
...
-        {headline && <Heading {...headline} />}
-        {content && <RichText html={content} />}
-        {arrayIsNonEmpty(buttons) && (
-          <ButtonContainer className="mt-16">
-            {buttons.map((button) => (
-              <Button key={button.text ?? button.href} {...button} />
-            ))}
-          </ButtonContainer>
-        )}
+    <div>
```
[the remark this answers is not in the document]
khallad2: i think it's needed to structure the html and match the design.
putting it in the `<div id={identifier} className="flex flex-row gap-16">` will not match the design because
it's a row and the design will not match.

app/domains/fluggastrechte/services/pdf/sections/reason/factsOfCases/table/getConnectionDetails.ts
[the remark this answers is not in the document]
khallad2: this is in an old commit, it's updated with the merge.

app/components/AccordionItem.tsx
khallad2: This syntax is often considered more readable and declarative for grouping multiple elements without adding extra DOM nodes.

i thought itt might be more readable than 
```
<span className="p-16 text-blue-800 flex items-center">
            {isOpen ? (
              [<KeyboardArrowUpIcon key="up" />, accordion.accordionItemHide]
            ) : (
              [<KeyboardArrowDownIcon key="down"/> ,accordion.accordionItemShow]
            )}
          </span>
```

app/root.tsx
```diff
@@ -148,6 +149,7 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
       postSubmissionText: parseAndSanitizeMarkdown(
         translations.feedback["text-post-submission"],
       ),
+      accordionTranslation: translations.accordion,
```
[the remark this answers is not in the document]
khallad2: my reason of not using the accessibility translations, is that it's a new component with it's own scope.

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+import KeyboardArrowUpIcon from "@digitalservicebund/icons/KeyboardArrowUp";
+import RichText from "~/components/RichText";
+import { useTranslations } from "~/services/translations/translationsContext";
+
+export type AccordionItemProps = Readonly<{
+  title: string;
+  description: string;
+  id?: number;
+  isOpen?: boolean;
+  onToggle?: () => void;
```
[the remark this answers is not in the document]
khallad2: because `isOpen?: boolean;
  onToggle?: () => void;` come from the state not from Strapi, so it doesn't exist in the schema but exist in props

app/components/__test__/Accordion.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  it("passes jsEnabled as true to each AccordionItem after mount", () => {
+    render(<Accordion items={dummyItems} />);
+    const jsEnabledIndicators = screen.getAllByTestId("js-enabled");
+    jsEnabledIndicators.forEach((indicator) => {
+      expect(indicator).toHaveTextContent("true");
+    });
+  });
+
+  it("allows toggling so that only one AccordionItem is open at a time", () => {
```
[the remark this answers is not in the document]
khallad2: the logic of open and closing now is handled in Accordion

app/components/Accordion.tsx
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from "react";
+import AccordionItem, { AccordionItemProps } from "~/components/AccordionItem";
+
+export type AccordionProps = Readonly<{
+  items: AccordionItemProps[];
+}>;
+
+export default function Accordion({ items }: AccordionProps) {
+  const [jsEnabled, setJsEnabled] = useState<boolean>(false);
+  const [openIndex, setOpenIndex] = useState<number>(-1);
```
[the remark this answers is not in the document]
khallad2: the numeric state approach is better suited for scenarios where only one item can be expanded at a time.
A single number (or a sentinel value like -1 for “none open”) directly represents which item is open. This avoids the overhead of managing an array of booleans where only one item should be true.

if it's boolean would be something like the following which is more code and complex than using numeric.
```
const [openStates, setOpenStates] = useState<boolean[]>(items.map(() => false));
const handleToggle = (index: number) => {
    setOpenStates((prev) =>
      prev.map((isOpen, i) => (i === index ? !isOpen : false))
    );
  };
```

app/components/__test__/Accordion.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+import type { AccordionItemProps } from "~/components/AccordionItem";
+import Accordion from "../Accordion";
+
+const dummyItems = [
+  { title: "Item 1", description: "Description 1", id: 1 },
+  { title: "Item 2", description: "Description 2", id: 2 },
+  { title: "Item 3", description: "Description 3", id: 3 },
+];
+
+vi.mock("~/components/AccordionItem", () => ({
```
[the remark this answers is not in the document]
khallad2: but we need to test the toggling functionality.

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+  jsEnabled?: boolean;
```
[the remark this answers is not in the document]
khallad2: id is required now, but isOpen?: boolean; onToggle?: () => void; come from the state not from Strapi, so it doesn't exist in the schema but exist in props

app/components/Accordion.tsx
chohner: The rounded borders exist only on the outside, not each element
image

app/components/AccordionItem.tsx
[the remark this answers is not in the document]
aaschlote: yeah, but when these props will be `undefined`?

app/components/AccordionItem.tsx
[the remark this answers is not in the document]
aaschlote: yeah, but when these props will be `undefined`?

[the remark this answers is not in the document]
aaschlote: and here my question is about the `jsEnabled` and assume it'll aways have a value

app/components/__test__/Accordion.test.tsx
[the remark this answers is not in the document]
aaschlote: this doesn't sound correctly from my point of view, because in the code the `Accordion` just handle the states of the top layer when the item is opened or not, when the logic to open and closing still in `AccordionItem`

image


image

app/components/AccordionItem.tsx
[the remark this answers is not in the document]
khallad2: explained and discussed in a huddle.

app/components/AccordionItem.tsx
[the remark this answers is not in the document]
khallad2: explained and discussed in a huddle.

app/components/ListItem.tsx
```diff
@@ -27,23 +29,31 @@ const ListItem = ({
...
+            <ButtonContainer className="mt-16">
+              {buttons.map((button) => (
+                <Button key={button.text ?? button.href} {...button} />
+              ))}
+            </ButtonContainer>
+          )}
+        </div>
       </div>
+      {accordion && (
+        <div role="region" className="w-full ml-56 mt-16">
```
aaschlote, line 53: I'm seeing this sonar error in my VSCode

"Use <section aria-label=...>, or <section aria-labelledby=...> instead of the "region" role to ensure accessibility across all devices.sonarqube(typescript:S6819)"

Can you please check it?

app/components/__test__/AccordionItem.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+    video: {},
+    accessibility: {},
+    fileUpload: {},
+    accordion: {
+      accordionItemShow: "Einblenden",
+      accordionItemHide: "Ausblenden",
+    },
+  };
+
+  beforeEach(() => {
```
aaschlote, line 41: I believe this before and after each is not needed, because you are not changing the values of the mock for every test, if you only add this line:

``
vitestVi.mocked(useTranslations).mockReturnValue(mockTranslations);
``

it should be sufficient 😃

app/services/cms/models/StrapiAccordion.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from "zod";
+import { HasStrapiIdSchema } from "~/services/cms/models/HasStrapiId";
+import { buildRichTextValidation } from "~/services/validation/richtext";
+
+export const StrapiAccordionSchema = z.object({
```
aaschlote, line 5: let's add a small unit test for this schema here

app/domains/fluggastrechte/services/pdf/sections/reason/factsOfCases/table/getConnectionDetails.ts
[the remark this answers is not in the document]
aaschlote: just bring back the the eslint disable again so we don't see this warning again

app/components/AccordionItem.tsx
```diff
@@ -0,0 +1,76 @@
+import KeyboardArrowDownIcon from "@digitalservicebund/icons/KeyboardArrowDown";
+import KeyboardArrowUpIcon from "@digitalservicebund/icons/KeyboardArrowUp";
+import RichText from "~/components/RichText";
+import { useTranslations } from "~/services/translations/translationsContext";
+
+export type props = AccordionItemProps &
```
aaschlote, line 6: the type name should be capital and I don't see reason to export it as well, if you are only using for the test

app/components/ListItem.tsx
```diff
@@ -27,23 +29,31 @@ const ListItem = ({
...
-        {content && <RichText html={content} />}
-        {arrayIsNonEmpty(buttons) && (
-          <ButtonContainer className="mt-16">
-            {buttons.map((button) => (
-              <Button key={button.text ?? button.href} {...button} />
-            ))}
-          </ButtonContainer>
-        )}
+    <div>
+      <div id={identifier} className="flex flex-row gap-16">
```
chohner, line 36: The identifier should be moved to the outer div I think

pgurusinga:
I noticed only one border between each collapsible accordion in the design. Could we update this? I’m not sure if other engineers have mentioned this.

image
Bildschirmfoto 2025-03-17 um 17 56 46

app/domains/fluggastrechte/services/pdf/sections/reason/factsOfCases/table/getConnectionDetails.ts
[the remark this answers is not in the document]
khallad2: i think this should be handled out of this PR, also we shouldn't suppress the warning but we we should use  nullish coalescing operator ?? instead of logical or ||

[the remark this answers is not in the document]
aaschlote: on the main it's already disabled this line, for some reason you did a change that it removed this line in this PR.

Please check here in the [main](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/app/domains/fluggastrechte/services/pdf/sections/reason/factsOfCases/table/getConnectionDetails.ts#L114)

PR #1839, opened 2025-03-13
feat(FGR): add validation between stopover, start and end airports

pgurusinga:
- [x] Error message added to Strapi


aaschlote (approved):
Please fill the values for the start and end airport for the other tests. Right now these values on theses tests are undefined, but in the real life they will have some value. The tests results should be the same!

image


LGTM!

PR #1845, opened 2025-03-14
feat(List): move from isNumeric to variant

chohner:
List: move from boolean `isNumeric` to `variant` union

depends on https://github.com/digitalservicebund/a2j-rechtsantragstelle-strapi/pull/407


app/services/cms/models/StrapiList.ts
```diff
@@ -19,7 +19,7 @@ export const StrapiListSchema = z
     heading: StrapiHeadingOptionalSchema,
     subheading: StrapiRichTextOptionalSchema(listRenderer),
     items: z.array(StrapiListItemSchema),
-    isNumeric: z.boolean(),
+    variant: z.enum(["unordered", "numbered"]),
```
aaschlote, line 22: maybe you can make the default value as `unordered`, so you won't need to define default in the props or even make optional [here](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1845/files#diff-6b496380adcae17a6ef2910317e80a24fdd2b56592ea6fd76aef0bace38e0cd2R11)

PR #1847, opened 2025-03-14
(docs): Split README into /docs [skip ci]

chohner:
Splits the overly long main README into explicit documents inside `/docs`

- better focus on important topics in README
- friendlier face of our repo to the public
- hopefully easier to find documentation (can be found with fuzzy file-search)
- maybe additional push to keep documentation up-to-date?


aaschlote (commented):
thanks for the initiative, I would have two recommendations:

1. Maybe we can use the`kebab-case` for the naming of the files as we do for the adr files
2. Maybe we can group some files in one file, like `addingService`, `arrayLogic` and `component-rendering` as `development`. Maybe we can check grouping other files if they are related about one topic.

PR #1848, opened 2025-03-17
feat: PKH add flow page and Weitere Angaben section to pdf

ingriddorioschulze:
This PR implements the new flow page Weitere Angaben. The new flow item is implemented in the navbar after Persönliche Daten and before Abgabe. At this page, user can add a text to the weitere angaben. If a text is available, a Weitere Angaben section will added to the PDF Anhang.

Screenshot 2025-03-17 at 10 28 50
Screenshot 2025-03-17 at 10 29 04


app/domains/prozesskostenhilfe/formular/index.ts
```diff
@@ -149,8 +150,21 @@ export const prozesskostenhilfeFormular = {
...
+        meta: { done: prozesskostenhilfePersoenlicheDatenDone },
+        states: {
+          "weitere-angaben": {
+            on: {
+              BACK: "#persoenliche-daten.beruf",
+              SUBMIT: "#abgabe",
+            },
+          },
+        },
+      },
```
Spencer6497, line 167: is there a reason we double-nested `weitere-angaben` here instead of defining it flatly?

app/domains/prozesskostenhilfe/services/pdf/index.ts
```diff
@@ -66,6 +67,11 @@ const buildProzesskostenhilfePDFDocument: PDFDocumentBuilder<
   if (requiresBelege(userData)) {
     buildBelegeList({ doc, documentStruct, userData, translations });
   }
+
+  if (userData.weitereAngaben) {
+    createWeitereAngaben(doc, documentStruct, userData);
+  }
+
```
Spencer6497, line 74: so i know we talked about this this morning but I think now that we're not worried about a discrete page for Weitere Angaben (and it can be merged with the existing Anhang on the same page), I think we can remove this pattern in favor of something like `fillWeitereAngaben` like so:

Bildschirmfoto 2025-03-17 um 16 13 08

Spencer6497 (changes requested):
See note about changing the pattern, otherwise great work 🤩

app/domains/prozesskostenhilfe/services/pdf/index.ts
ingriddorioschulze: I tried this approach but the Weitere Angaben section is rendered before the Belege List. Fill functions are called before the BelegeList. As we talked, we are keeping it like this for now and I suggest a refactor at some point :)

Spencer6497 (approved):
:shipit:

PR #1863, opened 2025-03-18
Remove remix json

aaschlote:
Remove Remix json due the deprecation of the `json` and improve source code to use in the future React router.

image

Only for response without setting the header values


pgurusinga:
should we also start writing tests here? 🤔

aaschlote:
> should we also start writing tests here? 🤔

what tests?

pgurusinga:
@aaschlote We’re working on updating the route but don't have (route) tests in place yet, I think Khallad mentioned this as a suggestion during our refactoring discussions. or do you think all changes here are safe enough?

aaschlote:
> @aaschlote We’re working on updating the route but don't have (route) tests in place yet, I think Khallad mentioned this as a suggestion during our refactoring discussions. or do you think all changes here are safe enough?

I believe that for these kinds of changes, unit tests won't effectively determine if the modifications are working, as they typically mock the library I replaced. To properly verify the changes, we need to run the application without mocking the library—in this case, by executing E2E tests. From what I’ve observed, most of the routes I modified already have E2E tests in place.

pgurusinga (approved):

PR #1869, opened 2025-03-19
Fix some html issues reported by W3 HTML Validator

aaschlote:
### Issues

- Element [meta](https://html.spec.whatwg.org/multipage/#the-meta-element) is missing one or more of the following attributes: content, property.
  - Solution: Add a random content recommended [here](https://github.com/darkreader/darkreader/pull/3902#issuecomment-1003728731)

image

- CSS: Deprecated media feature max-device-width
  - Solution: Use max-width instead of deprecated max-device-width

image

- Redundante roles
  - Solution: Remove roles where it's not necessary

image

- The aria-describedby attribute must point to an element in the same document.
  - Solution: Add aria-describedby to button and link when navigation is done

image


app/root.tsx
```diff
@@ -212,7 +212,10 @@ function App() {
         {description && <meta name="description" content={description} />}
         <meta property="og:title" content={ogTitle ?? title} />
         <meta property="og:description" content={description} />
-        <meta name="darkreader-lock" />
+        <meta
+          name="darkreader-lock"
+          content="1e82dc17-d02f-4566-8487-ae413a504055"
```
pgurusinga, line 217: how did you generate this uuid? is it possible to use the CSRF token instead?

aaschlote: I used this [website](https://www.uuidgenerator.net/). 

I can use it, but the CSRF token is not generated on all pages. For content pages, it's missing, so I would need to generate it at the root level, which doesn't seem ideal. Otherwise, I'll generate it only for this content meta.

PR #1879, opened 2025-03-21
feat: add new categorized footer

pgurusinga:


app/services/cms/models/StrapiFooter.ts
```diff
@@ -9,7 +11,12 @@ export const StrapiFooterSchema = z
   .object({
     image: StrapiImageOptionalSchema,
     paragraphs: z.array(StrapiParagraphSchema),
-    links: z.array(StrapiLinkSchema),
+    categorizedLinks: z.array(
```
aaschlote, line 14: we can add the option `.nonempty()` to force the array not be empty!

app/services/cms/models/StrapiFooter.ts
```diff
@@ -9,7 +11,12 @@ export const StrapiFooterSchema = z
...
+      z
+        .object({ title: z.string(), links: z.array(StrapiLinkSchema) })
```
aaschlote, line 16: same here, we can add the option `.nonempty()`

app/services/cms/models/StrapiFooter.ts
```diff
@@ -1,5 +1,7 @@
 import omit from "lodash/omit";
```
aaschlote, line 1: Can we add an unit test for this file?

app/components/Footer.tsx
```diff
@@ -10,19 +10,35 @@ type LinkProps = {
...
 
+type CategorizedLinkProps = {
+  id: number;
+  title: string;
+  links: LinkProps[];
+};
+
 export type FooterProps = Readonly<{
-  links?: LinkProps[];
+  categorizedLinks?: CategorizedLinkProps[];
```
aaschlote, line 20: why it's optional? I guess we can make it required, right?

app/components/Footer.tsx
```diff
@@ -57,20 +73,17 @@ export default function Footer({
...
-              <li key={link.url} className="leading-snug ds-label-03-reg">
-                <StandaloneLink
-                  text={link.text ?? ""}
-                  url={link.url}
-                  className="pb-6" // adding extra space to avoid bug in safari
-                />
-              </li>
-            ))}
-          </ul>
+          {categorizedLinks?.map((category) => (
```
aaschlote, line 79: let's remove `?` and make the `categorizedLinks` as required

app/components/Footer.tsx
```diff
@@ -57,20 +73,17 @@ export default function Footer({
...
+            <div key={category.id}>
+              <p className="ds-label-03-bold mb-7">{category.title}</p>
```
aaschlote, line 81: maybe it worth to check with Marion later if this should be a paragraph

app/components/Footer.tsx
```diff
@@ -57,20 +73,17 @@ export default function Footer({
...
+              <ul className="list-none pl-2 space-y-10">
```
aaschlote, line 82: I guess we should also check with Marion if this ok to have multiple `ul` html elements, because in this [ticket](https://app.asana.com/0/1206022902169650/1207820849591120) we make only one `ul`

app/components/Footer.tsx
```diff
@@ -10,19 +10,35 @@ type LinkProps = {
   text?: string;
 };
 
+type CategorizedLinkProps = {
```
aaschlote, line 13: make the type as ReadOnly :)

PR #1880, opened 2025-03-21
fix(Textarea): proposal to change stringOptionalSchema 

ingriddorioschulze:
This PR proposes a change in stringOptionalSchema to match browser and zod behavior on counting whitespaces and newlines

### DIFF
--- a/app/services/validation/stringOptional.ts
+++ b/app/services/validation/stringOptional.ts
@@ -3,6 +3,16 @@ import { TEXTAREA_CHAR_LIMIT } from "./inputlimits";
 
 export const stringOptionalSchema = z
   .string()
-  .trim()
-  .max(TEXTAREA_CHAR_LIMIT, { message: "max" })
+  //processing the string before validation.
+  .transform((string) => {
+    // remove possible whitespaces
+    return string.trim();
+  })
+  // using refine instead of max() to show validation message according to the string length
+  .refine(
+    (string) =>
+      // using replace to remove newlines
+      string.replace(/\r\n|\r|\n/g, "").length <= TEXTAREA_CHAR_LIMIT,
+    { message: "max" },
+  )
   .optional();


app/services/validation/stringOptional.ts
```diff
@@ -3,6 +3,16 @@ import { TEXTAREA_CHAR_LIMIT } from "./inputlimits";
 
 export const stringOptionalSchema = z
   .string()
-  .trim()
-  .max(TEXTAREA_CHAR_LIMIT, { message: "max" })
+  //processing the string before validation.
+  .transform((string) => {
+    // remove possible whitespaces
+    return string.trim();
+  })
```
chohner, line 10: I think you can just call `.trim()` on the `z.string()`. I think the comment could be removed, especially since `trim` only refers to the beginning and end, not all whitespace

app/services/validation/stringOptional.ts
```diff
@@ -3,6 +3,16 @@ import { TEXTAREA_CHAR_LIMIT } from "./inputlimits";
...
+  // using refine instead of max() to show validation message according to the string length
+  .refine(
+    (string) =>
+      // using replace to remove newlines
+      string.replace(/\r\n|\r|\n/g, "").length <= TEXTAREA_CHAR_LIMIT,
```
chohner, line 15: I think this could be a function, ie `stripAllWhitespace()`?

chohner: also, in regex you can just use `\s+` to match all whitespace, instead of listing the specific types

app/services/validation/stringOptional.ts
```diff
@@ -3,6 +3,16 @@ import { TEXTAREA_CHAR_LIMIT } from "./inputlimits";
...
 export const stringOptionalSchema = z
   .string()
-  .trim()
-  .max(TEXTAREA_CHAR_LIMIT, { message: "max" })
+  //processing the string before validation.
+  .transform((string) => {
+    // remove possible whitespaces
+    return string.trim();
+  })
+  // using refine instead of max() to show validation message according to the string length
```
chohner, line 11: I think this comment should be more explicit about the intent:

`can't use max(), as we want to count text length similar to the html input fields maxLength property (which ignores whitespace)`

app/services/validation/stringOptional.ts
```diff
@@ -3,6 +3,16 @@ import { TEXTAREA_CHAR_LIMIT } from "./inputlimits";
 
 export const stringOptionalSchema = z
   .string()
-  .trim()
-  .max(TEXTAREA_CHAR_LIMIT, { message: "max" })
+  //processing the string before validation.
```
chohner, line 6: I don't think this comment gives much additional context

app/services/validation/stringOptional.ts
```diff
@@ -3,6 +3,16 @@ import { TEXTAREA_CHAR_LIMIT } from "./inputlimits";
...
+  .transform((string) => {
+    // remove possible whitespaces
+    return string.trim();
+  })
+  // using refine instead of max() to show validation message according to the string length
+  .refine(
+    (string) =>
+      // using replace to remove newlines
+      string.replace(/\r\n|\r|\n/g, "").length <= TEXTAREA_CHAR_LIMIT,
+    { message: "max" },
```
chohner, line 16: One big issue with this approach: Users could save unlimited amounts of whitespace to our infra (because its not counted against the limit).

So from a security perspective, we should rather think about whether text inputs' `maxLength` property is the right thing to use...

PR #1888, opened 2025-03-24
feat(print-this-page-functionality): print result page checklist

ingriddorioschulze:
This PR implements the printing the content of a result page.


app/routes/shared/components/VorabcheckPage.tsx
```diff
@@ -27,6 +39,12 @@ export function VorabcheckPage() {
                 className="ds-stack ds-stack-16"
                 fullScreen={false}
               />
+              <Button
+                iconLeft={<FileDownloadIcon />}
+                text="Anleitung Herunterladen"
+                size="large"
+                href="/kontopfaendung/wegweiser/sozialleistungen-einmalzahlung?print"
+              />
```
chohner, line 47: the button shouldn't be in the code but added from strapi :)

app/util/url.ts
```diff
@@ -5,7 +5,7 @@ export function isExternalUrl(url: string) {
 }
 
 export function isFileDownloadUrl(url: string) {
-  return url.endsWith("/download/pdf");
+  return url.endsWith("/download/pdf") || url.endsWith("?print");
```
chohner, line 8: unit test?

app/routes/shared/components/VorabcheckPage.tsx
```diff
@@ -14,7 +17,16 @@ export function VorabcheckPage() {
...
     buttonNavigationProps,
+    shouldPrint,
   } = useLoaderData<typeof loader>();
+
+  useEffect(() => {
+    if (shouldPrint) {
+      window.print();
+      window.close();
+    }
+  }, [shouldPrint]);
```
chohner, line 28: Placing this here would mean `?print` only works on vorabcheck pages. I suppose this needs to work everywhere and should therefore be added to `root.ts`

app/routes/shared/vorabcheck.server.ts
```diff
@@ -40,6 +40,8 @@ export const loader = async ({
   const { flowId, stepId } = parsePathname(pathname);
   const cookieHeader = request.headers.get("Cookie");
 
+  const shouldPrintParam = searchParams.get("print") !== null;
```
chohner, line 43: see above - instead of restricting this to vorabchecks, lets move it into `root.ts`

app/routes/shared/components/VorabcheckPage.tsx
ingriddorioschulze: Yes, I am aware. It was just for testing purposes

app/util/url.ts
ingriddorioschulze: UT will follow :)

app/routes/shared/components/VorabcheckPage.tsx
ingriddorioschulze: True!

app/root.tsx
```diff
@@ -81,9 +81,11 @@ export const meta: MetaFunction<RootLoader> = () => {
 export type RootLoader = typeof loader;
 
 export const loader = async ({ request, context }: LoaderFunctionArgs) => {
-  const { pathname } = new URL(request.url);
+  const { pathname, searchParams } = new URL(request.url);
   const cookieHeader = request.headers.get("Cookie");
 
+  const shouldPrintParam = searchParams.get("print") !== null;
```
Spencer6497, line 87: nit: would simply name this `shouldPrint`, or better yet, define it inline in the return as it's not used elsewhere

Spencer6497 (approved):
great work 🙌

app/util/__test__/url.test.ts
```diff
@@ -14,6 +14,12 @@ describe("isFileDownloadUrl", () => {
   });
 });
 
+describe("isPrintUrl", () => {
```
chohner, line 17: I think this should be part of the previous `describe("isFileDownloadUrl"...` block

app/util/__test__/url.test.ts
```diff
@@ -14,6 +14,12 @@ describe("isFileDownloadUrl", () => {
...
+  it("returns true for print urls", () => {
```
chohner, line 18: "checks ?print query param"?

app/util/__test__/url.test.ts
```diff
@@ -14,6 +14,12 @@ describe("isFileDownloadUrl", () => {
...
+    expect(isFileDownloadUrl("?print")).toBeTruthy();
```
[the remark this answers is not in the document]
chohner: 1. agreed. maybe we should rather a relative URL (`/test/url?print`)
2. It's either `download/pdf` or `?print` - I don't think this function should be smarter

PR #1900, opened 2025-03-26
fix: adjust Sidenav menu width

khallad2:
(bug): Sidenav should be 248 px ([Ticket](https://app.asana.com/0/1203259475859667/1209657702141639/f))

Before

After


aaschlote (approved):
nice fixing, LGTM!

PR #1906, opened 2025-03-27
Implement logger with pino http

aaschlote:
### Motivation

In our Grafana system monitoring, we are currently unable to filter or create panels to analyze which paths are most frequently requested and which return error status codes. The project [digitalcheck-ditto](https://github.com/digitalservicebund/digitalcheck-dito) has its own request logging implementation, allowing them to track request paths and status codes within their service.

### Solution
Use `pino-http` dependency and integrated as middleware in the express server to log our request and response.

image


app/services/logging/createPinoHttpLogger.ts
```diff
@@ -0,0 +1,37 @@
+import { pinoHttp } from "pino-http";
+
+export const createPinoHttpLogger = () => {
+  return pinoHttp({
+    autoLogging: {
+      ignore: (req) =>
```
khallad2, line 6: Instead of chaining multiple `.startsWith` checks, you could define an array of ignored path prefixes and then use a concise check.

For example:
```
const ignoredPaths = ["/assets/", "/app/", "/@", "/node_modules/", "/data/"];
ignore: (req) =>
  ignoredPaths.some(prefix => (req.url ?? "").startsWith(prefix)),
  ```

Comment (aaschlote) [reply app/services/logging/createPinoHttpLogger.ts:6]: Yeap sure, I'm still playing around with this dependency. I want to present the idea first in dev exchange before making changes that it goes to production!

File: app/services/logging/createPinoHttpLogger.ts [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+    },
+    serializers: {
+      req: (req) => ({
+        method: req.method,
+        url: req.url,
+        userAgent: req.headers["user-agent"],
+      }),
+      res: (res) => ({
+        statusCode: res.statusCode,
+      }),
```
khallad2, line 34: suggestion if needed:  
adding error serializer can help log additional error details which can be useful during debugging.
```err: (err) => err ? { message: err.message, stack: err.stack, etc... } : err,```

File: app/services/logging/createPinoHttpLogger.ts [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+        return "warn";
+      } else if (res.statusCode >= 500 || err) {
+        return "error";
+      } else if (res.statusCode >= 300 && res.statusCode < 400) {
+        return "silent";
+      }
+      return "info";
+    },
+    transport: {
+      target: "pino-pretty", // Makes logs readable (optional)
```
khallad2, line 24: The `pino-pretty transport` is great for development, but it may not be ideal for production due to performance considerations. 
if you plan using it on Production please check trade-offs: 
- `Formatting Overhead` 
- `Increased CPU Usage`
- `Scalability`:if logs are consumed by log aggregators or monitoring tools it will expect structured JSON  

if you plan using it on Development
 Consider conditionally applying the transport based on an environment variable something like:

```
transport: process.env.NODE_ENV === 'development'
  ? { target: "pino-pretty" }
  : undefined,
```

app/services/logging/createPinoHttpLogger.ts
khallad2: nice work anyway :tada:

app/services/logging/createPinoHttpLogger.ts
aaschlote: actually I want to remove the `pino-pretty`, I was just checking locally. By default on local developments it should be disable, but sometimes we need it. I'll remove it later!

judithmh:
LGTM :) Only one question: As far as I understand the logs will be saved in Grafana -  could it be a privacy concern to save the host-name info? and do you know where the pid comes from?

judithmh:
> Thanks for pointing it, the host-name and the pid are information about the machine is running the application, in this case it would be the pods in K8s. But I removed from the remote ip address who is doing the request, in this case the user ip address.
> 
> So it shouldn't have any problem about the data protection 😃

ah true that makes sense! thanks for the explanation and clarification :)

judithmh (approved):

chohner (approved):
lgtm

PR #1912, opened 2025-03-28
feat(Conditional Rendering): Add Strapi visibility field

Spencer6497:
This PR is a Proof-of-Concept for implementing a visibility field in Strapi, in which content editors can simply put the name of a condition, e.g. "hasArbeit", and have the frontend evaluate it and decide whether or not to show the component.

On the strapi side, this would be simply adding an optional text field to every hideable component, I've named it "isVisible" here.

On the app side, we've replaced the `interpolateSerializableObject()` function with a 1:1 drop-in replacement, `recursivelyReplaceStrings()` that foregoes the deserialization/serialization and instead deeply replaces string values with their desired stringReplacement counterparts. It also checks for the aforementioned visibility field, and simply filters the object out if it evaluates to false.


chohner:
Instead of pushing this responsibility into each component - could we do this inside the `loader`? ie filtering out any component who's `shouldDisplay` evaluates to `true`?

app/components/Accordion.tsx
```diff
@@ -8,24 +10,29 @@ export type AccordionProps = Readonly<{
 export default function Accordion({ items }: AccordionProps) {
   const [jsEnabled, setJsEnabled] = useState<boolean>(false);
   const [openIndex, setOpenIndex] = useState<number>(-1);
+  const { replacements } = useLoaderData() as { replacements: Replacements };
```
khallad2, line 13: Potenial bug: 
- If the loader function fails to provide replacements or returns it in an unexpected format, you could encounter runtime errors.
Solution:
- Instead of using a type assertion with `as`, maybe you can define a type for the loader data and use generics with useLoaderData. so you catch errors on development if any and apply type safety.
something like 
```
type LoaderData = { replacements: Replacements; }
const { replacements } = useLoaderData<LoaderData>();
```

aaschlote (commented):
I wrote a few comments, but I don't think it's a good idea to expose all the string replacements from the server to the client

app/util/recursivelyReplaceStrings.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Deeply replaces values with stringReplacements, removing content that shouldn't be visible
+ */
+export function recursivelyReplaceStrings<T>(
+  content: T,
+  stringReplacements?: Replacements,
+): T {
+  if (!stringReplacements || !content) return content;
+  if (typeof content === "string") {
+    return mustache.render(content, stringReplacements) as T;
```
chohner, line 14: could we make this into its tiny helper `applyStringReplacement` function? because there are multiple places where we don't want the complexity of the big function

chohner: also, I'm a little surprised this needs a cast

app/services/flow/formular/buildFormularServerTranslations.ts
```diff
@@ -48,7 +48,7 @@ function interpolateTranslations(
     return translation;
   }
 
-  return interpolateSerializableObject(
+  return recursivelyReplaceStrings(
```
chohner, line 51: would just need `applyStringReplacement()`

app/routes/beratungshilfe.zustaendiges-gericht.auswahl.$PLZ.tsx
```diff
@@ -30,10 +30,10 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
     fetchMeta({ filterValue }),
   ]);
 
-  const resultListHeading = fillTemplate({
-    template: common.resultListHeading,
-    replacements: { postcode: zipCode ?? "" },
-  });
+  const resultListHeading = recursivelyReplaceStrings(
```
chohner, line 33: would just need `applyStringReplacement()`

app/components/arraySummary/ArraySummaryDataItems.tsx
```diff
@@ -38,7 +38,7 @@ const ArraySummaryDataItems = ({
     return null;
   }
 
-  const heading = interpolateSerializableObject(
+  const heading = recursivelyReplaceStrings(
```
chohner, line 41: would just need `applyStringReplacement()`

chohner (approved):
nice, I think this looks pretty good!

PR #1919, opened 2025-04-01
fix: special chars issue

pgurusinga:
https://app.asana.com/0/1207022303033556/1209855503144765/f


app/components/inputs/autoSuggestInput/customComponents/FormatOptionLabel.tsx
```diff
@@ -1,6 +1,10 @@
 import type { FormatOptionLabelMeta } from "react-select";
 import type { DataListOptions } from "~/services/dataListOptions/getDataListOptions";
 
+function escapeSpecialCharacters(string: string) {
+  return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
aaschlote, line 5: the error only happens with parentheses, so we don't need to scape all the special characters besides the parentheses.

Please remember to rename the function!

aaschlote (approved):
LGTM!

PR #1924, opened 2025-04-02
Fix dropdown icon high contrast mode

aaschlote:
### Before

image

### After

image


app/components/inputs/Select.tsx
```diff
@@ -29,7 +29,7 @@ const Select = ({
   const { error, getInputProps } = useField(name, { formId });
 
   const selectClassName = classNames(
-    "ds-select",
+    "ds-select forced-color-adjust-none",
```
pgurusinga, line 32: Is this clarified with Marion whether we want to disable or have a select with HCM meaning? The dropdown should have a background colour matching the body background colour with an inverted line colour.

aaschlote: Nope, but the only ACs I got from the ticket were:

All selects have a visible outline in HCM ✅
All selects have a visible dropdown icon in HCM ✅

I haven't seen any mentioned about the background color matching. But I can talk with her once she's available!

pgurusinga (approved):

app/components/inputs/__test__/Select.test.tsx
```diff
@@ -86,5 +86,18 @@ describe("Select", () => {
...
+        <Select
+          name="select"
+          options={[]}
+          label="Test Label"
+          formId="formId"
+        />,
+      );
+      const element = screen.getByRole("combobox");
+      expect(element).toHaveClass("forced-color-adjust-none");
+    });
```
chohner, line 101: I have to admit - I dont understand why we need to test static classnames...

aaschlote: I just want to make sure if this className is in my element and if someone will remove it, it should think twice 👀

PR #1931, opened 2025-04-03
Enable single fetch remix

aaschlote:
### Problem
We should move remix to react router as remix has been integrated inside the react router environment. Before doing it, we need to enable the single-fetch feature flag, because by default react router will use it.

### Solution
Enable feature flag "single-fetch" following this [documentation](https://remix.run/docs/en/main/start/future-flags#v3_singlefetch) and adapt all the functions to work with this feature.

Source: https://reactrouter.com/upgrading/remix


tsconfig.json
```diff
@@ -11,16 +11,22 @@
...
     "incremental": true,
     "skipLibCheck": true,
-    "lib": ["DOM", "DOM.Iterable", "ES2022"],
+    "lib": ["DOM", "DOM.Iterable", "ESNext"],
     "isolatedModules": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "jsx": "react-jsx",
-    "module": "ES2022",
+    "module": "ESNext",
```
chohner, line 19: I think this [could/should](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html) be `nodenext`?

tsconfig.json
```diff
@@ -11,16 +11,22 @@
     "preserveSymlinks": true,
     "incremental": true,
     "skipLibCheck": true,
-    "lib": ["DOM", "DOM.Iterable", "ES2022"],
+    "lib": ["DOM", "DOM.Iterable", "ESNext"],
```
chohner, line 14: any reason we need to add `ESNext` here? I'm worried this might lead to less support of older browsers (unless we start shipping poly-fills).

app/routes/action.send-rating.ts
```diff
@@ -1,5 +1,5 @@
 import type { ActionFunctionArgs, Session, SessionData } from "@remix-run/node";
-import { json, redirect } from "@remix-run/node";
+import { data as remixData, redirect } from "@remix-run/node";
```
chohner, line 2: I would prefer to use the original `data()` name and rename any potentially existing variables

app/routes/__test__/action.send-rating.test.ts
```diff
@@ -37,7 +42,13 @@ describe("/action/send-rating route", () => {
       `http://localhost:3000/action/send-rating?url=${ratingPath}&js=false`,
       options,
     );
-    const response = await action({ request, params: {}, context: {} });
+
+    const response = (await action({
+      request,
+      params: {},
+      context: {},
+    })) as Response;
```
chohner, line 50: I'm surprised this (and following) casts are necessary - did the return value of `action()` change? (not a blocker, since this is only in tests...)

tsconfig.json
aaschlote: The [documentation](https://remix.run/docs/en/main/start/future-flags#adopt-the-vite-plugin) mentioned to use `ESNext`, I can use it, but not sure what it would be the side effects.

tsconfig.json
aaschlote: I just followed the documentation, maybe I can remove it, but I'm not 100% sure!

app/routes/__test__/action.send-rating.test.ts
aaschlote: it looks like with this feature flag enable not it returns a `DataWithResponseInit` (due `data()`) or `ValidationErrorResponseData`. The value didn't change, somehow typescript was complaining about it and I was forced to cast it.

image

Spencer6497 (approved):
FINALLY 🎉

tsconfig.json
pgurusinga: This is just the tsconfig. I think we are fine with `esnext`, since Chris updated the Vite config to compile the code to esnext 
refer to: https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/main/vite.config.ts#L41

pgurusinga (approved):

tsconfig.json
aaschlote: I was checking this option again and I found out it:

``TypeScript includes a default set of type definitions for built-in JS APIs (like Math), as well as type definitions for things found in browser environments (like document)``

Maybe I don't need this option at all, because it's only related browser configuration.

I'll remove it!

app/routes/__test__/action.send-rating.test.ts
```diff
@@ -26,8 +27,12 @@ describe("/action/send-rating route", () => {
...
     );
     const response = await action({ request, params: {}, context: {} });
-    expect(response.status).toEqual(200);
-    expect(response.ok).toBeTruthy();
+
+    if ("data" in response) {
+      expect(response.data.success).toEqual(true);
+    } else {
+      throw new Error("Response does not contain 'data' property");
+    }
```
chohner, line 35: I'm confused why we throw an error now in the test?

app/routes/__test__/action.send-rating.test.ts
chohner: ok, strange. I think we might be doing something slightly not-so-correct then but lets figure that out later. maybe write a small TODO at the top of the file to revisit later?

aaschlote: I'll do it

app/routes/__test__/action.send-rating.test.ts
aaschlote: When it runs the test to get the response data is assigned to an object element `data`

image

But typescript doesn't see it and I'm forced to create this condition to throw an exception  in case it doesn't not exist

image

PR #1936, opened 2025-04-04
feat(a11y): enable focus on error for radio

pgurusinga:
https://app.asana.com/0/1207022303033556/1208646940266128

## TODO
- [ ] needs Marion clarification whether this is for all components that shows error or just Radio
- [ ] Remove live-region

<h2>Before</h2>
before-output

<h2>After</h2>
output

### DIFF
--- a/app/components/inputs/InputError.tsx
+++ b/app/components/inputs/InputError.tsx
@@ -1,16 +1,23 @@
 import ErrorOutline from "@digitalservicebund/icons/ErrorOutline";
 import type { PropsWithChildren } from "react";
+import { forwardRef } from "react";
 
 type InputErrorProps = PropsWithChildren<{
   readonly id: string;
+  readonly tabIndex?: number;
 }>;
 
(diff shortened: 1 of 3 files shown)


chohner:
Why do we want to focus on the non-interactive error message?

pgurusinga:
@chohner this is ticket https://app.asana.com/0/1207022303033556/1208646940266128


> The current pattern we use for error management is not being announced to SR users, because the error appears only when the focus is leaving the radio group empty, but the aria-describedby is associated with this field: without focus it cannot be pass on to users. Therefore, we need to move the focus to the radio button group after submitting the form, so that the users will hear the error message.

chohner:
> Therefore, we need to move the focus to the radio button group after submitting the form, so that the users will hear the error message.

The conclusion talks about focusing the radio group, not the error message

pgurusinga:
ah true, why am I always think that I need to focus on error message 🫠 , thanks for the hint

pgurusinga:
will create a new PR with less noise

PR #1939, opened 2025-04-07
feat(Radio): focus component when error message appears

pgurusinga:
## Problem
Radio input is not in a focus state when validation fails on submission

## Proposed changes
- Use remix-validated-form API `handleReceiveFocus` (reference: https://www.remix-validated-form.io/reference/use-field#handleReceiveFocus)
- Pass ref to Radio
- Focus on the first radio input https://www.w3.org/WAI/ARIA/apg/patterns/radio/, 
	- `handleReceiveFocus` said the first invalid input will receive focus, but in reality, it will focus on the last part, so we need extra logic

## Before
before-output 

## After
after-output


aaschlote (approved):
LGTM!

PR #1945, opened 2025-04-09
Move to new version of validation form

aaschlote:
### Problem
While working on [this PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1943), we discovered that upgrading to React Router 7 requires us to also adopt the latest version of the remix-validated-form library (now known as rvf).

### Solution
Since the latest major version of rvf introduces significant breaking changes, we decided to upgrade incrementally. As a first step, we're moving to version [v6](https://github.com/airjp73/rvf/releases/tag/rvf-v6.0.0), which still supports remix, allowing us to minimize the scope of changes before the full transition to React Router 7.

### Changes

- Installed `@rvf/zod` and `@rvf/remix`
- Removed `@remix-validated-form/with-zod` and `remix-validated-form`
- Refactored all input components to support the new useField API
- Adjusted default values for nested fields, following the [migration guide](https://www.rvf-js.io/migration#other-breaking-changes)
- Updated the `Checkbox` component to handle checked and unchecked states
- Added a workaround for `AutoSuggestionInput` to handle validation on onChange and onBlur when errors are present
- Wrapped input stories in Storybook with `FormProvider` context

### Upcoming Tasks

- Once we migrate to React Router 7, we'll replace @rvf/remix with @rvf/react-router, as outlined in this [release](https://github.com/airjp73/rvf/releases/tag/v7). The transition should be straightforward since both packages expose the same APIs.
- We should open an issue regarding the workaround in AutoSuggestionInput, as calling field.validate() seems to trigger validation on all form fields, not just the targeted one.


package.json
```diff
@@ -50,10 +50,11 @@
     "@digitalservicebund/icons": "^2.0.1",
     "@mjackson/form-data-parser": "^0.7.0",
     "@pdf-lib/fontkit": "^1.1.1",
+    "@rvf/remix": "^6.3.0",
+    "@rvf/zod": "^7.0.2",
```
pgurusinga, line 54: should this be aligned with version 6 first?

aaschlote: we could, but I have seen any changes related the source code from the version 6 to 7 for this dependency.

image

app/components/filesUpload/__test__/FilesUpload.test.tsx
```diff
@@ -35,18 +35,39 @@ vi.mock("~/services/translations/translationsContext", () => ({
...
+  getErrorMock.mockReturnValue(error);
+};
+
+const renderFilesUpload = ({ ...args }: Partial<FilesUploadProps> = {}) =>
+  render(
+    <RouterProvider
+      router={createMemoryRouter([
+        {
+          path: "/",
+          element: <FilesUpload name={fieldName} formId="formId" {...args} />,
```
pgurusinga, line 63: I saw in the code, we don't use formId, why pass it here?

app/components/filesUpload/FilesUpload.tsx
```diff
@@ -29,7 +26,6 @@ export type FilesUploadProps = {
 const FilesUpload = ({
   name,
   title,
-  formId,
```
pgurusinga, line 32: could you remove from the props type?

app/components/inputs/__test__/Textarea.test.tsx
```diff
@@ -34,16 +33,9 @@ describe("Textarea component", () => {
...
-        path: "",
-        Component: () => (
-          <Textarea name={componentName} label="Test Label" formId="formId" />
-        ),
-      },
-    ]);
-
-    render(<RemixStub />);
+    render(
+      <Textarea name={componentName} label="Test Label" formId="formId" />,
```
pgurusinga, line 37: we don't need formId right?

app/components/inputs/__test__/Textarea.test.tsx
```diff
@@ -60,21 +52,14 @@ describe("Textarea component", () => {
...
-      },
-    ]);
-
-    render(<RemixStub />);
+    render(
+      <Textarea
+        name="test-textarea"
+        label="Test Label"
+        description="Test Description"
+        formId="formId"
```
pgurusinga, line 60: also here formId?

pgurusinga: basically we could remove formId since it's not used in the useField()

aaschlote: this is true, I forgot to remove it. I'll do it right now!

app/services/validation/useStringField.ts
```diff
@@ -1,12 +0,0 @@
-import { useField } from "remix-validated-form";
-
-export function useStringField(
-  name: string,
-  options?: Parameters<typeof useField>[1],
-) {
-  const { defaultValue, ...props } = useField(name, options);
```
pgurusinga, line 7: do you know what this hook was created? maybe @chohner can answer?

.storybook/RFCFormerProvider.tsx
```diff
@@ -0,0 +1,16 @@
...
+type Props = {
+  children: ReactNode;
+};
+
+export const RFCFormerProvider = ({ children }: Props) => {
+  const form = useForm({
+    validator: withZod(z.object({ name: z.string().optional() })),
+  });
+
+  return <FormProvider scope={form.scope()}> {children}</FormProvider>;
```
pgurusinga, line 15:
```suggestion
  return <FormProvider scope={form.scope()}>{children}</FormProvider>;
```

pgurusinga (approved):

chohner (approved):
tried locally 👍🏼

PR #1946, opened 2025-04-09
feat(PKH): add files upload page to PKH

ingriddorioschulze:
This PR implements a files upload page to the PKH flow


frederike-ramin (commented):
Nice, correct. Only things missing, I think:
- on "ende" we need a conditional BACK link to match the new SUBMIT link
- This you can test with a flow test in the according `testcases` file
- Adapt the e2e tests to just click "next" on the dokumente page (so the pipeline is green again)

PR #1949, opened 2025-04-10
add string replacements for meta information

judithmh:


judithmh:
> should we apply this change also in the `formular.server.ts` and `result.server.ts`?

good question - we were also thinking about it - so far we have no cases where we need string Replacements for formular and result but I can also add this already just in case

aaschlote (approved):
LGTM!

PR #1950, opened 2025-04-10
refactor(PostHog): Consolidate PostHog instances

Spencer6497:
This PR attempts to consolidate a lot of the spidery PostHog logic & instantiation we have throughout the app. It does so by (primarily) using the nodejs backend client as the single source of truth, passing it through the loader to the frontend and providing it to consumers through Context.

Notable exception to this: the PostHog NodeJS SDK doesn't support surveys yet, so for this feature we still have to fallback to the javascript web sdk.

- **tweak PosthogProvider to expose cookie header to consumers, implement in CookieBanner as proof-of-concept**
- **implement PosthogProvider for UserFeedback component**
- **remove instantiation of posthog-node client inside of feature flag logic**
- **adapt unit test**


app/components/userFeedback/reportProblem/ReportProblem.tsx
```diff
@@ -1,29 +1,18 @@
...
-      if (POSTHOG_API_KEY) {
-        posthog.init(POSTHOG_API_KEY, {
-          api_host: POSTHOG_API_HOST,
-          loaded: () => {
-            setSurvey(fetchSurvey());
-          },
-        });
-      }
-    }
     setSurvey(fetchSurvey());
```
aaschlote, line 16: do you still need to keep it in a `useState`? I guess you can move out from here, or am I missing something?

app/services/analytics/posthogHelpers.ts
```diff
@@ -1,11 +1,22 @@
...
+export function fetchSurvey(surveyId: string) {
   let survey: Survey | undefined;
   posthog.getSurveys((surveys) => {
-    survey = surveys.find((survey) => survey.id === feedbackSurveyId);
+    survey = surveys.find((survey) => survey.id === surveyId);
   });
   return survey;
 }
+
+export function idFromCookie(cookie: string | null) {
```
aaschlote, line 13: should we add an unit test here?

app/services/analytics/posthogHelpers.ts
```diff
@@ -1,11 +1,22 @@
...
+  // Note: can't use cookie.parse(): https://github.com/remix-run/remix/discussions/5198
+  // Returns ENVIRONMENT if posthog's distinct_id can't be extracted
```
aaschlote, line 15: why do we return the `ENVIRONMENT` if we can get it? should we return empty?

app/components/cookieBanner/CookieBanner.tsx
```diff
@@ -26,51 +26,37 @@ export function CookieBanner({
...
-        api_host: POSTHOG_API_HOST,
+    const captureConsent = async () => {
+      if (!hasTrackingConsent && posthog?.optOut) {
+        await posthog.optOut();
+      } else if (hasTrackingConsent && posthog?.optIn) {
+        await posthog.optIn();
+      }
+    };
 
-        session_recording: {
```
aaschlote, line 41: are we setting this configuration somewhere? We had a critical bug because we didn't setup the posthog config correct. You can find the changes in this [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/1822).

Maybe it worth to check it before merging your changes!

aaschlote (commented):
Nice changes, I guess we need just to check about the tracking of the page before we merge these changes.

app/services/analytics/posthogHelpers.ts
Spencer6497: I think it's just so the PostHog `distinctId` will be more explicit when we look at it; i.e. the event will be tagged with 'local' or 'production' instead of `undefined`

app/components/cookieBanner/CookieBanner.tsx
Spencer6497: so these aren't even initialization options for the NodeJS SDK, presumably because the NodeJS SDK doesn't support event Session Replay. However, I will add them to where we're initializing the JS one

app/services/analytics/posthogHelpers.ts
pgurusinga: can we rename it to `client-[ENVIRONMENT]` ? as you probably notice, we have posthogDistinctId for feature flag set as `backend`

pgurusinga:
@Spencer6497, my last comment: how would we test this? I think we should extensively test this to avoid missing tracking data.

app/services/analytics/__test__/posthogHelpers.test.ts
```diff
@@ -0,0 +1,32 @@
...
+}));
+
+describe("posthogHelpers", () => {
+  describe("idFromCookie", () => {
+    it("should return ENVIRONMENT if POSTHOG_API_KEY is unavailable", () => {
+      expect(idFromCookie("cookieString")).toBe("client-local");
+    });
+
+    it("should return ENVIRONMENT if the posthog cookie's distinct_id is undefined", () => {
+      mockAPIKey = "mockAPIKey";
```
aaschlote, line 22: I wouldn't mock using a global variable, for two reasons:

1. When we have a big unit test file, we might get difficult to understand what the variable does, specially if it's global.
2. If you change the order of this test, you will see this test fails.

You can check how are mocking in the test `app/services/errorPages/__test__/throw404.test.ts`, so for every test you mock the values for your test.

A small improvement you also can do it, it's following the AAA pattern (Arrange, Act and Assert). 

So it would be something like:


```
it("should return the posthog cookie's distinct_id", () => {
     // Arrange
      mockAPIKey = "mockAPIKey";
      phCookieString = '{"distinct_id": "mockDistinctId"}';

      //Act
      const actual = idFromCookie("cookieString");

      //Assert
      expect(actual).toBe("mockDistinctId");
    });
```

You can check more in this article [here](https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80).

Spencer6497:
> @Spencer6497, my last comment: how would we test this? I think we should extensively test this to avoid missing tracking data.

After merging I was going to monitor PostHog to see if any of our metrics/indicators rapidly drop off; not sure how else we would test regressions?

app/services/analytics/__test__/posthogHelpers.test.ts
Spencer6497: oh I didn't realize I could use `vi.mocked()`. thanks, will update

PR #1951, opened 2025-04-10
Ignore sentry errors

aaschlote:
### Problem
Since enabling the singleFetch feature in Remix, we’ve been encountering intermittent errors such as `Error in input stream` and `Load failed`. These issues only affect a small subset of users, and unfortunately, we haven’t been able to reliably reproduce them in development or staging environments. Additionally, Sentry continues to trigger alerts for these errors, even though they have already been archived.

### Solution
To ensure Sentry alerts reflect actionable and traceable issues, we propose filtering out these specific, non-reproducible errors from our monitoring pipeline. This will help us focus on real problems and reduce alert fatigue.


app/entry.client.tsx
```diff
@@ -4,6 +4,9 @@ import { startTransition, StrictMode, useEffect } from "react";
 import { hydrateRoot } from "react-dom/client";
 import { config } from "~/services/env/web";
 
+// Ignore a few common errors that are not useful to track
+const SENTRY_IGNORE_ERRORS = ["Error in input stream", "Load failed"];
```
pgurusinga, line 8: what about this error? https://github.com/remix-run/remix/issues/10214

aaschlote: it seems some people mentioned the issue has been resolved once they moved to react router 7 and react 19. And as this issue has been a few times 16 times in the last seven days I would say we don't need to but in the list.

Or do we see it makes sense to dd to the list?

chohner:
I have to say - I dont know how comfortable I feel disabling sentry errors in the application. Do we expect to disable this filter again once moving to react router?

aaschlote:
> I have to say - I dont know how comfortable I feel disabling sentry errors in the application. Do we expect to disable this filter again once moving to react router?

I guess I will keep this PR open until we update to react router.

aaschlote:
This problem will be solved in this [PR](https://github.com/remix-run/react-router/pull/12945), so I'll merge it and remove once they fixed in the react router side.

PR #1962, opened 2025-04-14
feat: make checkbox focused on error

pgurusinga:


app/components/inputs/Checkbox.tsx
```diff
@@ -49,7 +49,8 @@ const Checkbox = ({
           className={className}
           aria-describedby={field.error() ? errorId : undefined}
           onClick={() => setRenderHiddenField(!renderHiddenField)}
-          required={required}
+          aria-required={required}
```
pgurusinga, line 52: Note: This is important! If required, browser input validation will validate (which we don't want); I don't know why it didn’t work before.

app/components/inputs/__test__/Checkbox.test.tsx
```diff
@@ -64,9 +84,43 @@ describe("Checkbox", () => {
     expect(errorMessage).toBeInTheDocument();
   });
 
-  it("renders the checkbox as required", () => {
-    render(<Checkbox name="checkbox-name" label="Checkbox Label" required />);
+  it("sets aria-required attribute based on required prop", () => {
```
aaschlote, line 87: maybe we can have a test to check the another way around (required props as false should not have the attribute with the value true)

aaschlote (approved):
besides my comment, LGTM!

PR #1967, opened 2025-04-15
Fix: TileGroup should focus on first element on error

pgurusinga:
## Issue  
The first element of the tile group does not receive focus.  

## Proposal  
- Implement forwardRef  
- Ensure focus on the first element  
- Enhance testing procedures

## After
tile-output


aaschlote (approved):
LGTM!

PR #1975, opened 2025-04-17
feat(BerH): attach user's uploaded files to Antrag PDF

ingriddorioschulze:
1. This PR implements user uploaded files to Antrag PDF as follows:

- When abgabeArt is online, user's uploaded files will be attached to the Antrag PDF (Antrag Formular, Anhang, Nächsten Schritte Online and Uploaded Files) 
- When abgabeArt is ausdrucken, there will be no files upload and therefore the PDF will be rendered as before (Antrag Formular, Anhang and Nächsten Schritte)

2. This PR refactors s3 logic passing sessionId instead the cookieHeader to the functions so http layer doesn't leak into s3 layer


app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @param userData The user data containing saved file keys
+ * @param cookieHeader The cookie header for the request for authentication
+ * @param flowId The flow ID for the request
+ * @returns The resulting PDF document as a Uint8Array
+ */
+export async function embedUserFilesToPdf(
+  mainPdfBuffer: Uint8Array,
+  userData: BeratungshilfeFormularContext,
+  cookieHeader: string | null,
+  flowId: FlowId,
```
aaschlote, line 39: I don't think you need the `flowId` as parameter, this function is already on the scope for the `beratungshilfe`, so it makes this parameter redundant.

app/services/externalDataStorage/userFileS3Helpers.ts
```diff
@@ -67,3 +72,54 @@ export async function deleteUserFileFromS3(
...
+      return await response.Body.arrayBuffer().then((buffer) => {
+        return Buffer.from(buffer);
+      });
+    }
+    throw new Error("Response body is not a readable stream");
+  } catch (error) {
+    const errorDescription =
+      error instanceof Error ? error.message : "Unknown error";
+
+    sendSentryMessage(
```
aaschlote, line 104: This error is not necessary it throws already an exception, so it will be capture from Sentry in the `entry.server.tsx`.

app/services/externalDataStorage/userFileS3Helpers.ts
```diff
@@ -67,3 +72,54 @@ export async function deleteUserFileFromS3(
...
+        Bucket: config().S3_DATA_STORAGE_BUCKET_NAME,
+        Key: getObjectKey(sessionId, flowId, savedFileKey),
+      }),
+    );
+    if (!response.Body) {
+      throw new Error("Response body is undefined");
+    }
+    if (response.Body instanceof Readable) {
+      return await streamToBuffer(response.Body);
+    } else if (response.Body instanceof Blob) {
```
aaschlote, line 94: do you need to make an `else if`?

I guess you can eliminate it with another if condition.

I also would recommend to you the reading of this [article](https://medium.com/swlh/stop-using-if-else-statements-f4d2323e6e4).

app/services/externalDataStorage/__test__/userFileS3Helpers.test.ts
```diff
@@ -119,4 +123,38 @@ describe("userFileS3Helpers", () => {
       expect(mockS3Client.send).toBeCalled();
     });
   });
+
+  describe("downloadUserFileFromS3", () => {
```
aaschlote, line 127: I would expect a few more tests:

- When the `Response.Body` is instance of `Blob`
- When an error happens

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,65 @@
...
+import { type BeratungshilfeFormularContext } from "~/domains/beratungshilfe/formular";
+import { type FlowId } from "~/domains/flowIds";
+import { downloadUserFile } from "~/services/upload/fileUploadHelpers.server";
+
+/**
+ * Extracts all saved file keys from userData.
+ * @param userData user data which can contain saved file keys in the form of arrays of objects
+ * @returns an array of all saved file keys
+ */
+export const extractSavedFileKeys = (
```
aaschlote, line 11: Are you using this function outside of this file? If you are `export` it only for tests, this is a bad pattern and we should avoid it. 

If you need to test it, it's better to move to a new file, otherwise exporting it only to test is not good.

Check this [article](https://medium.com/decathlondigital/should-i-test-private-methods-direclty-c48f4fa7bb4d) and this [one](https://dev.to/canro91/this-is-why-we-dont-test-private-methods-28ef) for understand better it.

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,65 @@
...
+  userData: BeratungshilfeFormularContext,
+) => {
+  const allKeys: string[] = [];
+  for (const value of Object.values(userData)) {
```
aaschlote, line 15: You can improve this code like this:

```
Object.values(userData)
  .filter(Array.isArray)
  .flatMap(files =>
    files
      .filter(file => file.savedFileKey)
      .map(file => file.savedFileKey!)
  )
```

With this you avoid if statements and const variables.

aaschlote (commented):
I wrote a few comments, just ping me in case you need a better clarification!

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,65 @@
...
+      userPdfFile.getPageIndices(),
+    );
+
+    // Add the copied pages to the main PDF document
+    for (const copiedPdfFilePage of copiedPdfFilePages) {
+      mainPdfDocument.addPage(copiedPdfFilePage);
+    }
+  }
+  const finalPdf = await mainPdfDocument.save();
+  return finalPdf;
```
chohner, line 64: I think you can `return mainPdfDocument.save()`, the caller will already need to await it

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
chohner: I think if instead of `BeratungshilfeFormularContext` this uses generic user data this could be used for multiple flows, instead of only beratungshilfe. but maybe thats for the future..

app/services/upload/fileUploadHelpers.server.ts
```diff
@@ -100,18 +101,25 @@ export async function deleteUserFile(
   return false;
 }
 
+export async function downloadUserFile(
```
chohner, line 104: I'm a bit confused by this function.. `downloadUserFileFromS3` already does everything this function does?

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
aaschlote: ok, but for now it's not generic and the methods is on the scope for the `beratungshilfe`, so having this parameter is redundant.

app/services/externalDataStorage/userFileS3Helpers.ts
ingriddorioschulze: Ok, I did that. 
I can sadly not read the article. It is for members only.

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
ingriddorioschulze: Cool, thanks for suggesting!

app/services/upload/fileUploadHelpers.server.ts
ingriddorioschulze: You are completely right! I think I got lost on my own abstractions 😆

app/domains/beratungshilfe/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
ingriddorioschulze: I like the idea of using the Generic Context and changed my implementation. If I use a generic context I will need the flowId, or it is still redundant?

aaschlote: It only be redundant to have the parameter if it's inside the scope of the domain `beratungshilfe`, so in case you want to have it generic, move outside of the `domain` folder 😃

ingriddorioschulze: Cool, makes more sense!

app/domains/beratungshilfe/services/pdf/index.ts
```diff
@@ -55,6 +57,8 @@ const buildBeratungshilfePDFDocument: PDFDocumentBuilder<
 
 export async function beratungshilfePdfFromUserdata(
   userData: BeratungshilfeFormularContext,
+  cookieHeader: string | null,
+  flowId: FlowId,
```
aaschlote, line 61: as we discussed yesterday, we can remove this parameter, it's on the "beratunshilfe" scope and we don't need a parameter for the `flowId` here.

app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -19,11 +19,12 @@ type PdfFlowContexts =
   | FluggastrechteFlugdatenContext
   | ProzesskostenhilfeFormularContext;
 
-export type PdfConfig = PdfFlowContexts extends infer T
+type PdfConfig = PdfFlowContexts extends infer T
   ? T extends PdfFlowContexts
     ? {
         pdfFunction: (
           userData: T,
+          cookieHeader: string | null,
```
chohner, line 27: I think this would be much simplified if it where `userSessionId: string | undefined`

app/services/externalDataStorage/userFileS3Helpers.ts
```diff
@@ -67,3 +72,52 @@ export async function deleteUserFileFromS3(
     throw error;
   }
 }
+export async function downloadUserFileFromS3(
+  cookieHeader: string | null,
```
chohner, line 76: I would prefer to separate our S3 interface from cookies etc, so lets pass the `sessionId` here instead.

Also, shouldn't never be `null` or `undefined`, right?

app/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Embeds user uploaded files into a PDF document.
+ * @param mainPdfBuffer The PDF document as a Uint8Array
+ * @param userData The user data containing saved file keys
+ * @param cookieHeader The cookie header for the request for authentication
+ * @param flowId The flow ID for the request
+ * @returns The resulting PDF document as a Uint8Array
+ */
+export async function embedUserFilesToPdf(
```
chohner, line 28: I think `attach` is probably a better verb, instead of `embed`?

aaschlote, line 28: why the name of the function is different from the file?

aaschlote (approved):
besides my comment, LGTM!

app/services/pdf/userUploadedFilesToPdf/embedUserUploadedFilesToPdf.ts
ingriddorioschulze: Probably it was the exactly moment my brain got distracted. Thanks for pointing out!

app/services/externalDataStorage/userFileS3Helpers.ts
ingriddorioschulze: Ok, yes, we could do this separation. Is that ok if I merge this PR and then we work on the refactor?

app/routes/shared/pdfDownloadLoader.ts
ingriddorioschulze: Ok, yes, we could do this separation. Is that ok if I merge this PR and then we work on the refactor?

Spencer6497 (approved):

app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -33,14 +37,17 @@ export type PdfConfig = PdfFlowContexts extends infer T
...
+      userData: BeratungshilfeFormularContext,
+      sessionId: string,
+    ) => await beratungshilfePdfFromUserdata(userData, sessionId),
     filenameFunction: () =>
       `Antrag_Beratungshilfe_${pdfDateFormat(today())}.pdf`,
   },
   "/prozesskostenhilfe/formular": {
     pdfFunction: async (
       userData: ProzesskostenhilfeFormularContext,
+      _sessionId: string,
```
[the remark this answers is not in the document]
ingriddorioschulze: We are not using sessionId on this scope. I am guessing it is from an earlier iteration. It is a required param here and that's why it is being passed with the underscore: 

`type PdfConfig = PdfFlowContexts extends infer T
  ? T extends PdfFlowContexts
    ? {
        pdfFunction: (
          userData: T,
          sessionId: string,
          translations?: Translations,
        ) => Promise<Uint8Array>;
        filenameFunction: () => string;
      }
    : never
  : never;`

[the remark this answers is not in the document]
chohner: The issue is that there are now two optional params (`sessionId` & `translations`). To do this cleanly we should probably move from to object parameters
```
{
	userData: X;
	sessionId?: string;
	translations: Translations
}
```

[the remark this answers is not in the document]
ingriddorioschulze: We can check that after the merge since it is no blocker :)

PR #2014, opened 2025-04-28
feat(Wegweiser): combine variants to be able to use isVisible to display accordion

ingriddorioschulze:
This PR implements three new variants that are a combination of other variants. The idea was to group variants is to make it possible to use isVisible function to display the accordion when one of the conditions is true.


app/domains/kontopfaendung/wegweiser/stringReplacements.ts
```diff
@@ -181,3 +181,90 @@ export const getKinderStrings = (userData: KontopfaendungWegweiserContext) => {
...
+  const { hasEinmalSozialleistung } = getEinmalSozialleistungStrings(userData);
+  const { hasNachzahlungSozialUnter500 } =
+    getNachzahlungSozialUnter500Strings(userData);
+  const { hasNachzahlungArbeitUnter500 } =
+    getNachzahlungArbeitUnter500Strings(userData);
+  const { hasBuergergeld } = getBuergergeldStrings(userData);
+  const { hasGrundsicherung } = getGrundsicherungStrings(userData);
+  const { hasAsylbewerberleistung } = getAsylbewerberleistungStrings(userData);
+  const hasPflegegeldSelbst =
+    getPflegegeldSelbstStrings(userData).hasPflegegeldSelbst;
```
chohner, line 204: this could be simplified a lot if these only return a boolean:

```typescript
const hasErhoehungsbetrag = (userData) =>  userData.kinderUnterhalt === "yes" ||
      userData.kinderWohnenZusammen === "ja" ||
      userData.partnerUnterhalt === "yes" ||
      userData.partnerWohnenZusammen === "yes";

const hasKindergeld = (userData) => userData.sozialleistungenUmstaende?.kindergeld === CheckboxValue.on


const stringReplacements = (userData) => {
	hasErhoehungsbetrag: hasErhoehungsbetrag(userData),
	hasKindergeld: hasKindergeld(userData),
	...
	needsSchuldnerberatung: hasErhoehungsbetrag(userData) || hasKindergeld(userData) || ...

}
```

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
```diff
@@ -181,3 +181,90 @@ export const getKinderStrings = (userData: KontopfaendungWegweiserContext) => {
...
+      hasErhoehungsbetrag,
+      hasKindergeld,
+      hasEinmalSozialleistung,
+      hasNachzahlungSozialUnter500,
+      hasNachzahlungArbeitUnter500,
+      hasBuergergeld,
+      hasGrundsicherung,
+      hasAsylbewerberleistung,
+      hasPflegegeldSelbst,
+    },
```
chohner, line 216: IIRC, the values should all be combined using OR (`||` in javascript) to produce a single boolean value

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
```diff
@@ -181,3 +181,90 @@ export const getKinderStrings = (userData: KontopfaendungWegweiserContext) => {
     hasKinder: userData.hasKinder === "yes",
   };
 };
+/**
+ * Returns an object with a single key `schuldnerberatungsstelleIsVisible` which is an object of boolean values.
+ * The keys of this object are the names of the conditions that make the Schuldnerberatungsstelle visible.
+ * The values of this object are the results of the corresponding string replacement functions.
+ * The Schuldnerberatungsstelle is visible if any of the conditions are true.
+ */
```
chohner, line 189: I agree with the current API some explanation might be needed, however comments are usually reserved for bugs or surprising things that aren't obvious from the code. Here we can probably resolve this by making the functions more reusable

app/domains/kontopfaendung/wegweiser/__test__/stringReplacements.test.ts
```diff
@@ -271,4 +274,162 @@ describe("stringReplacements", () => {
       });
     });
   });
+  describe("getSchuldnerberatungsstelleStrings", () => {
```
chohner, line 277: Personal opinion: Don't think we need to extensively test these boolean combinations (or rather I think the time is better spend other places).
The logic itself is relatively trivial and its too easy to hide bugs inside this testing code (because its reproducing exactly what we wrote in the implementation)

chohner (changes requested):

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
ingriddorioschulze: You are completely right!

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
ingriddorioschulze: Ok, I see your point. I mostly wanted to document this combining variable functions because we did not have this before.

app/domains/kontopfaendung/wegweiser/__test__/stringReplacements.test.ts
ingriddorioschulze: Totally get your point. My idea was to keep track of the combined variables. In case some change happens, we still  have a chance to get it with the tests. I will keep an eye to balance that better next time :)

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
```diff
@@ -181,3 +181,66 @@ export const getKinderStrings = (userData: KontopfaendungWegweiserContext) => {
     hasKinder: userData.hasKinder === "yes",
   };
 };
+export const getSchuldnerberatungsstelleStrings = (
```
chohner, line 184: these still need to be added to `wegweiser/index.ts` I think?

ingriddorioschulze: Totally true! I forgot this part 🙈

chohner (approved):
lgtm

PR #2017, opened 2025-04-28
Proposal: move translation into component

chohner:
Following discussion on reducing the amount of strings from the CMS:

Removing the string lookup for `Ja` and `Nein` simplifies component, tests and scaffolding around it..


Spencer6497 (approved):
crazy how much cruft there was

aaschlote (commented):
I’d prefer not to hardcode the text directly in the source code. Instead, using a tool like [react-i18next](https://react.i18next.com/) would be a better approach. That way, if the text ever needs to be updated (even if it's infrequent), we can simply modify a configuration file rather than the code itself.

This also enables non-engineers to make content changes without needing developer involvement. Additionally, keeping translations separated by feature or domain helps maintain cleaner, more modular code that's easier to manage and evolve over time.

PR #2023, opened 2025-04-29
add fonts for self hosting

judithmh:
- Added google noto to our code base for self hosting the font and included it into our stylesheet


app/styles.css
```diff
@@ -289,3 +289,10 @@ details:not(:has(.summary-content:focus-visible)) {
 .auto-suggest-input-disabled {
   box-shadow: none;
 }
+
+@font-face {
+  font-family: "Fira Sans";
+  font-style: normal;
+  font-weight: 400;
+  src: url(/fonts/fira-sans/google_noto.woff2) format("woff2");
```
chohner, line 297: noto? i think this should be fira? also why the subfolder?

judithmh: you are right! I made the subfolder because they were becoming many fonts and to have some structure for the future ;) but I guess since it is just this one font added for now I will remove the folder again :)

ingriddorioschulze (approved):

aaschlote (approved):
LGTM!

You also can update `app/services/security/cspHeader.server.tsx` in the property `style-src`, we don't need to enable the `fonts.googleapis.com` anymore.

judithmh:
> LGTM!
> 
> You also can update `app/services/security/cspHeader.server.tsx` in the property `style-src`, we don't need to enable the `fonts.googleapis.com` anymore.

Nice thanks! removed it from style and font src :)

PR #2030, opened 2025-04-30
bug(PKH): allow PDF generation with special characters 

ingriddorioschulze:
This PR fixes the PDF generation with special characters:
- Register fontkit with PDFDocument in the printNameInSignatureFormField to allow pdf to be generated with special characters like "ń"

### DIFF
--- a/app/domains/prozesskostenhilfe/services/pdf/printNameInSignatureFormField.ts
+++ b/app/domains/prozesskostenhilfe/services/pdf/printNameInSignatureFormField.ts
@@ -1,15 +1,26 @@
+import fs from "fs";
+import fontkit from "@pdf-lib/fontkit";
 import type { PDFDocument } from "pdf-lib";
 import type { ProzesskostenhilfeFormularContext } from "../../formular/context";
 
-export function printNameInSignatureFormField(
+export async function printNameInSignatureFormField(
   pdfDoc: PDFDocument,
   userData: ProzesskostenhilfeFormularContext,
 ) {
   if (userData.versandArt === "digital") {
+    // registering fontkit with PDFDocument to encode and embed Unicode characters like "ń", "ś", "ł"
+    // without fontkit, pdf-lib will use simple encoding (WinAnsi) and will throw "WinAnsi cannot encode
+    // "ń" when it does not know how to encode
+    pdfDoc.registerFontkit(fontkit);
+    const fontBytes = fs.readFileSync(
+      "public/fonts/BundesSansWeb-Regular.woff",
+    );
+    const font = await pdfDoc.embedFont(fontBytes, { subset: true });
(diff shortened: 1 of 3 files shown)


judithmh (approved):

app/domains/prozesskostenhilfe/services/pdf/printNameInSignatureFormField.ts
```diff
@@ -1,15 +1,26 @@
...
-export function printNameInSignatureFormField(
+export async function printNameInSignatureFormField(
   pdfDoc: PDFDocument,
   userData: ProzesskostenhilfeFormularContext,
 ) {
   if (userData.versandArt === "digital") {
+    // registering fontkit with PDFDocument to encode and embed Unicode characters like "ń", "ś", "ł"
+    // without fontkit, pdf-lib will use simple encoding (WinAnsi) and will throw "WinAnsi cannot encode
+    // "ń" when it does not know how to encode
+    pdfDoc.registerFontkit(fontkit);
```
aaschlote, line 14: in the function `fillPdf` is already doing it, should you call the method `embedFont` there as well?

ingriddorioschulze: I guess it id already being done here:

` const customPdfFormFont = await pdfDoc.embedFont(bundesSansCondensed, {
    features: { liga: false },
  });`
  
  Is that what you mean?

aaschlote: Yeah, but can you do there as well? There is embed the `BundesSansCond-DTP-Regular.otf`, maybe it's missing to add the `public/fonts/BundesSansWeb-Regular.woff`.

If you do there, it's a global change, so it won't need to apply for the other parts of the code

ingriddorioschulze:
> Should we also do the same for [BerH](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/9d7afaa9b6ef24c5a52003122abf506829dfa0d5/app/domains/beratungshilfe/services/pdf/pdfForm/header.ts#L27)?

@Spencer6497  in BerH is being done [here](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/9ab5f8308862a2997f2fcc65a8e6528bcea83c27/app/services/pdf/fillPdf.server.ts#L56)

The difference between both implementations are:

PHK uses `printNameInSignatureFormField` function to sign the PDF with name + nachname

BerH uses the fillFunction and the signing happens in the [fillFooter ](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/9ab5f8308862a2997f2fcc65a8e6528bcea83c27/app/domains/beratungshilfe/services/pdf/pdfForm/footer.ts#L24)

PR #2031, opened 2025-04-30
Rename component header to hero

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-52

### DIFF
--- a/app/services/cms/models/StrapiHeading.ts
+++ b/app/services/cms/models/StrapiHeading.ts
@@ -24,9 +24,6 @@ export const StrapiHeadingSchema = z
       "ds-body-01-reg",
       "ds-body-02-reg",
     ]),
-    className: z.string().optional(),
-    dataTestid: z.string().optional(),
-    tagId: z.string().optional(),
   })
   .merge(HasOptionalStrapiIdSchema)
   .nullable()
--- a/app/services/cms/models/__test__/StrapiHero.test.ts
+++ b/app/services/cms/models/__test__/StrapiHero.test.ts
@@ -1,0 +1,71 @@
+import { StrapiHeroSchema } from "../StrapiHero";
+
+describe("StrapiHero", () => {
+  it("should return false given undefined container", () => {
+    const undefinedContainer = {
+      heading: {
+        text: "someText",
+      },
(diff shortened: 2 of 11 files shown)


app/services/cms/models/StrapiHeading.ts
```diff
@@ -24,9 +24,6 @@ export const StrapiHeadingSchema = z
       "ds-body-01-reg",
       "ds-body-02-reg",
     ]),
-    className: z.string().optional(),
-    dataTestid: z.string().optional(),
-    tagId: z.string().optional(),
```
judithmh, line 29: out of curiosity: why did you remove className, dataTestid, tagId?

aaschlote: they are not needed, just cleaning up the source code 🧹

judithmh (approved):

app/services/cms/models/__test__/StrapiHero.test.ts
```diff
@@ -0,0 +1,71 @@
+import { StrapiHeroSchema } from "../StrapiHero";
+
+describe("StrapiHero", () => {
```
chohner, line 3: imo this is just testing `zod.parse` and not really needed.

chohner (approved):
lgtm but I think we shouldn't test the zod parsing

app/services/cms/models/__test__/StrapiHero.test.ts
aaschlote: The purpose of testing this file is to verify the behavior of the validation logic when handling unknown data using zod. Since zod is an external dependency, future updates could potentially introduce changes that break our current validation logic.

We’ve previously encountered a related issue with the InfoBoxComponent — specifically, the detailsSummary attribute was not being transformed into details, which caused unintended behavior. You can find more context on that in [this Slack thread](https://digitalservicebund.slack.com/archives/C04MC0KTHJA/p1739460767410709). A basic unit test could have caught that problem before it reached production.

In short, even though this test is simple, it adds valuable protection to the platform by helping prevent unexpected regressions in the future.

PR #2038, opened 2025-05-02
chore(ci): minimize workflow permissions [skip ci]

chohner:
- defaults to `package: read` by default
- granularly expands permission on job-scope


.github/workflows/ci-pipeline.yml
```diff
@@ -63,7 +66,6 @@ jobs:
     runs-on: ubuntu-latest
     needs: [code-quality, verify-local-e2e]
     permissions:
-      contents: read
       id-token: write # for cosign w/ keyless signing
       packages: write # for updating cosign attestation
       security-events: write
```
Spencer6497, line 71: should `security-events: write` be removed here like as above?

Spencer6497 (approved):

.github/workflows/ci-pipeline.yml
aaschlote: I guess we need to run the `code scanning` after the docker image is created. Check it [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token)

.github/workflows/npm-checks.yml
```diff
@@ -24,5 +24,4 @@ jobs:
         if: ${{ github.actor != 'dependabot[bot]' && matrix.npmCMD == 'test:coverage' }}
         uses: SonarSource/sonarqube-scan-action@aa494459d7c39c106cc77b166de8b4250a32bb97 # v5.1.0
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Needed to get PR information, if any
```
aaschlote, line 27: don't we need it?

.github/workflows/scan.yml
```diff
@@ -45,8 +45,8 @@ jobs:
           ]
     permissions:
       contents: read
-      id-token: write # for cosign w/ keyless signing
-      packages: write # for updating cosign attestation
+      id-token: read # for cosign w/ keyless signing
```
aaschlote, line 48: why did you change to `read`? I haven't seen any mention to be `read`, only to `write`. Check it [here](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#updating-your-actions-for-oidc).

.github/workflows/scan.yml
```diff
@@ -45,8 +45,8 @@ jobs:
...
+      packages: read # for updating cosign attestation
```
judithmh, line 49: can we really change this to read here if cosign attestation needs to be added?

chohner: good question. I wasn't sure what the `packages: write` was used for. is there some documentation somewhere?

Spencer6497: [yes](https://github.com/marketplace/actions/cosign-installer#usage) it appears to be necessary

PR #2044, opened 2025-05-06
Move to express 5

aaschlote:
### Problem
`Remix` didn't support express 5, but since we moved to `React-Router`, we can upgrade our application to run the `Express 5`

### Solution
Upgrade our application to run the Express 5

### Changes

- Update dependencies `express` and `@types/express`
- Add a name `splat` for wildcard *. Check this [link](https://expressjs.com/en/guide/migrating-5.html#path-syntax)
- Handle errors in `Express 5` as event. Check this [link](https://expressjs.com/en/guide/migrating-5.html#app.listen)

Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-92


app/expressApp.ts
```diff
@@ -57,7 +58,8 @@ export const expressApp = (
   // Everything else (like favicon.ico) is cached for an hour
   // You may want to be more aggressive with this caching.
   app.use(express.static("build/client", { maxAge: "1h" }));
-  app.all("*", reactRouterHandler);
+  // Express 5 must have the wildcard * with name. Check https://expressjs.com/en/guide/migrating-5.html#path-syntax
+  app.all("/{*splat}", reactRouterHandler);
```
chohner, line 62: why `{splat}` here and `*splat` above?

app/expressApp.ts
```diff
@@ -17,7 +17,7 @@ export const expressApp = (
   viteDevServer: ViteDevServer,
 ) => {
   const redisClient = getRedisInstance();
-  const reactRouterHandler = createRequestHandler({ build }) as RequestHandler; // express 4 doesn't handle returned promises
+  const reactRouterHandler = createRequestHandler({ build }) as RequestHandler; // express 5 doesn't handle returned promises
```
chohner, line 20: so sad that this still isn't matching in types 🥲

chohner (approved):
works for me!

app/expressApp.ts
aaschlote: I guess we don't need it, I copied somewhere from github thread issue. I'll change it! 

Thanks for mention it!

PR #2052, opened 2025-05-06
feat(BerH): attach relevant documents only to PDF

ingriddorioschulze:
This PR implements the attachment of only relevant documents to final PDF. 

The solution covers the following case:
- user selects "Bürgergeld" and uploads files for "BürgergeldBeweis"
- user goes back and selects "Asylbewerberleistungen" and uploads files for "AsylbewerberleistungenBeweis"
- PDF is generated with both "BürgergeldBeweis" and "AsylbewerberleistungenBeweis"


app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/__test__/attachUserUploadedFilesToPdf.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      mockMainPDFBuffer,
+      mockedUserData,
+      sessionId,
+      flowId,
+    );
+
+    const resultPdf = await PDFDocument.load(resultBuffer);
+
+    expect(resultPdf.getPageCount()).toBe(2);
+    expect(resultBuffer).toBeInstanceOf(Uint8Array);
```
chohner, line 59: this is already in the typing system and shouldn't need testing. also - we dont REALLY care that this is an `Uint8Array` :D

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/attachUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,138 @@
...
+        evidence.map((fileMeta) =>
+          downloadUserFileFromS3(
+            sessionId,
+            flowId,
+            fileMeta.savedFileKey ?? "",
+          ),
+        ),
+      );
+    }
+  };
```
chohner, line 55: Not mandatory but I think separating this into its own file would be cleaner (especially since this is modifying `relevantDocumentsPromises` inline)

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/attachUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,138 @@
...
+  addRelevantDocuments(
+    leistungen.hasNoSozialleistung,
+    userData.keineLeistungenBeweis,
+  );
+  addRelevantDocuments(
+    leistungen.hasGrundsicherung,
+    userData.grundsicherungBeweis,
+  );
+  addRelevantDocuments(
+    geldAnlagen.hasLebensversicherung ?? false,
```
chohner, line 81: I think adding adjusting `addRelevantDocuments` to take `condition?: boolean` might be easier than explicitly passing `false`

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/attachUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,138 @@
...
+  for (const file of relevantDocuments) {
+    const userPdfFile = await PDFDocument.load(file);
+    const copiedPdfFilePages = await mainPdfDocument.copyPages(
+      userPdfFile,
+      userPdfFile.getPageIndices(),
+    );
+    for (const copiedPdfFilePage of copiedPdfFilePages) {
+      mainPdfDocument.addPage(copiedPdfFilePage);
+    }
+  }
```
chohner, line 136: this is the part where I would try to re-use `appendPagesToPdf()`. Then you don't need to test the attaching-files functionality anymore..

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/__test__/attachUserUploadedFilesToPdf.test.ts
```diff
@@ -0,0 +1,190 @@
...
+import { CheckboxValue } from "~/components/inputs/Checkbox";
+import { type BeratungshilfeFormularContext } from "~/domains/beratungshilfe/formular";
+import { downloadUserFileFromS3 } from "~/services/externalDataStorage/userFileS3Helpers";
+import { attachUserUploadedFilesToPdf } from "../attachUserUploadedFilesToPdf";
+
+vi.mock("~/services/externalDataStorage/userFileS3Helpers", () => ({
+  downloadUserFileFromS3: vi.fn(),
+}));
+
+describe("attachUserUploadedFilesToPdf", () => {
```
chohner, line 12: looking at the test, this function does too much - it both contains service logic (attaching two pdfs) and business logic (when should which pdf be attached). be re-using the existing `appendPagesToPdf()` function we could simplify the scope to be only about BH scope

chohner (approved):
my comments are all mostly stylistic. no blockers, I think the functionality should work (without having tried it)

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/attachUserUploadedFilesToPdf.ts
```diff
@@ -0,0 +1,138 @@
+import { PDFDocument } from "pdf-lib";
```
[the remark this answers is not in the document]
ingriddorioschulze: Yes, that's fine. I will opt for this one `services/pdf/attachUserUploadedFilesToPdf/index.ts`

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/__test__/attachUserUploadedFilesToPdf.test.ts
ingriddorioschulze: Makes sense!!

app/domains/beratungshilfe/services/pdf/attachUserUploadedFilesToPdf/attachUserUploadedFilesToPdf.ts
ingriddorioschulze: Yes, it works using nicely `appendPagesToPdf()`

PR #2069, opened 2025-05-12
bug(FilesUpload): display file size 

ingriddorioschulze:
This PR proposes a fix in the conversion of the displayed fileSizes. 
The issue was that fileSizes were being displayed as 0MB when the file size was too small. The extended logic does the conversion and shows bytes, kilobytes or megabytes. 

Examples: 

**1. Very small files (smaller than 1 KB)** -> 500 bytes will be displayed as "500 B"

Screenshot 2025-05-12 at 14 00 40

**2. Small files (between 1 KB and 1MB)** -> 19718 bytes will be displayed as "19,3 KB"

Screenshot 2025-05-12 at 14 01 04

**3. Medium/larger files (bigger or equal to 1 MB)** -> 2500000 bytes will be displayed as "2,4 MB"

Screenshot 2025-05-12 at 14 00 07


app/services/upload/__test__/formatFileSizeToString.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  it("should convert kilobytes for small file (1KB - 1MB) sizes correctly", () => {
+    const bytesCount = 19718;
+    expect(formatFileSizeToString(bytesCount)).toBe("19,3 KB");
+  });
+
+  it("should convert convert megabytes for large file (>= 1 MB) sizes correctly", () => {
+    const bytesCount = 2500000;
+    expect(formatFileSizeToString(bytesCount)).toBe("2,4 MB");
+  });
+  it("should convert zero bytes correctly", () => {
```
aaschlote, line 18: can we add some tests when the parameter of this function is a negative value or `NaN`? I would expect the result should be `0 B`.

app/components/filesUpload/FileUploadInfo.tsx
```diff
@@ -38,7 +38,7 @@ export const FileUploadInfo = ({
           {file.filename}
         </p>
         <p className="ds-body-01-reg text-gray-900">
-          {bytesToMegabytesString(file.fileSize ?? 0)}
+          {formatFileSizeToString(file.fileSize ?? 0)}
```
aaschlote, line 41: I guess we don't need anymore the default 0 here as the `fileSize` is not more `undefined` or `null`.

aaschlote (approved):
I wrote two small comments, but LGTM!

app/components/filesUpload/FileUploadInfo.tsx
ingriddorioschulze: Yes, good catch! Thanks :)

Spencer6497 (approved):
LGTM 🌟

PR #2086, opened 2025-05-16
feat(PKH AB Test): add component to display AB test version

ingriddorioschulze:
This PR implements an icon and a string to the form when the flow is prozesskostenhilfe/formular and the feature flag conversion-rate-pkh-flow has the variant key test. When the feature flag has the variant key control the flow should be as before.
Using the multiple variant feature flag key helps us to check which experiment variant the user has been assigned to. In case of a problem with the experiment or feature flag, users will keep seeing the usual flow.

### DIFF
--- a/app/components/AbTest/EstimatedTimeAbTest.tsx
+++ b/app/components/AbTest/EstimatedTimeAbTest.tsx
@@ -1,0 +1,38 @@
+import TimerOutlinedIcon from "@digitalservicebund/icons/TimerOutlined";
+import { useFeatureFlagVariantKey } from "posthog-js/react";
+import { useLocation } from "react-router";
+
+export function EstimatedTimeAbTest() {
+  /**
+   * Hook to determine whether to display the estimated time string and icon.
+   *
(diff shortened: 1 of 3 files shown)


app/components/abTest/EstimatedTimeAbTest.tsx
```diff
@@ -0,0 +1,35 @@
+import TimerOutlinedIcon from "@digitalservicebund/icons/TimerOutlined";
+import { useFeatureFlagVariantKey } from "posthog-js/react";
+import { useLocation } from "react-router";
+import { config } from "~/services/env/web";
+
+export function EstimatedTimeAbTest() {
+  function useDisplayEstimatedTimeAbTest(): boolean {
+    if (!config().POSTHOG_API_KEY) {
```
aaschlote, line 8: can we move this under the `useLocation` so we can reduce at least one ``eslint-disable``?

aaschlote (approved):
besides my comment, LGTM!

m0dh4x (approved):
LGTM

PR #2107, opened 2025-05-21
Add properties for svg in the fill black

aaschlote:
Before:
image

After:
image

### DIFF
--- a/app/styles.css
+++ b/app/styles.css
@@ -38,6 +38,10 @@ svg {
     stroke: white;
     stroke-width: 0.75;
   }
+  .svg-image path[fill="black"] {
+    stroke: white;
+    stroke-width: 1.25;
+  }
 }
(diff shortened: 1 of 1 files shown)


app/styles.css
```diff
@@ -38,6 +38,10 @@ svg {
     stroke: white;
     stroke-width: 0.75;
   }
+  .svg-image path[fill="black"] {
```
pgurusinga, line 41: This implementation is working for our case but breaks the BMJV icon in forced colors preference light mode

### Dark
image

### Light
image

pgurusinga (approved):

pgurusinga (approved):

app/styles.css
aaschlote: fixed, closing it!

PR #2108, opened 2025-05-21
add aria-describedby to customInput for AutoSuggest

judithmh:


aaschlote:
@judithmh why did you create this PR? For fixing what problem?

PR #2112, opened 2025-05-22
refactor(userData): Rename context to user data in wegweiser from context to userData

ingriddorioschulze:
This PR refactors the instances of context to userData in wegweiser


chohner (approved):
lgtm, except the import order

PR #2117, opened 2025-05-22
feat: auto suggest is focusable on disabled state

pgurusinga:
As we discussed in the daily, we will implement focusable as it is, and for screen reader users, we will fix it with a different approach (e.g FieldSet)


aaschlote (approved):
LGTM!

PR #2123, opened 2025-05-23
chore(deps): pin `posthog-js` to investigate pageView regression

ingriddorioschulze:
This PR changes the posthog-js version to a lower one to understand if the newer version might have affected the mobile page views


chohner (approved):
context for the future: page views dropped dramatically for mobile devices (factor 3x). The change seems to coincide with this version change

PR #2124, opened 2025-05-23
Automatically send user focus on the first h1

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-35

### DIFF
--- a/app/components/useFocusHeadingTag.tsx
+++ b/app/components/useFocusHeadingTag.tsx
@@ -1,0 +1,17 @@
+import { useEffect } from "react";
+
+const HEADING_ONE_TAG = "h1";
+
+export const useFocusHeadingTag = () => {
+  useEffect(() => {
+    const focusTimer = setTimeout(() => {
+      const headingOneTag = document.querySelectorAll(HEADING_ONE_TAG);
+
+      if (headingOneTag.length > 0) {
+        headingOneTag[0].setAttribute("tabindex", "-1");
+        headingOneTag[0].focus();
+      }
+    }, 0);
+    return () => clearTimeout(focusTimer);
+  }, []);
+};
--- a/app/components/__test__/useFocusHeadingTag.test.tsx
+++ b/app/components/__test__/useFocusHeadingTag.test.tsx
@@ -1,0 +1,38 @@
+import { render, waitFor } from "@testing-library/react";
+import { useFocusHeadingTag } from "../useFocusHeadingTag";
+
+const TEXT_HEADING_LABEL = "Test Heading";
+
+function DummyComponentWithH1() {
+  useFocusHeadingTag();
+  return <h1>{TEXT_HEADING_LABEL}</h1>;
+}
+
+function DummyComponentWithoutH1() {
+  useFocusHeadingTag();
+  return <h2>{TEXT_HEADING_LABEL}</h2>;
+}
+
+describe("useFocusHeadingTag", () => {
+  it("should set tabindex and focus in case h1 element is present", async () => {
+    const { getByText } = render(<DummyComponentWithH1 />);
+
+    const element = getByText(TEXT_HEADING_LABEL);
+
+    await waitFor(() => {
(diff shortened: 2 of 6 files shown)


app/components/useFocusHeadingTag.tsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from "react";
+
+const HEADING_ONE_TAG = "h1";
+
+export const useFocusHeadingTag = () => {
+  useEffect(() => {
+    const focusTimer = setTimeout(() => {
+      const headingOneTag = document.querySelectorAll(HEADING_ONE_TAG);
```
chohner, line 8: if we always want the first one we could just use `querySelector()`?

app/components/useFocusHeadingTag.tsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from "react";
+
+const HEADING_ONE_TAG = "h1";
+
+export const useFocusHeadingTag = () => {
```
chohner, line 5: maybe `useFocusFirstH1()` makes it a bit more explicit?

aaschlote: true, I'll rename it!

app/components/useFocusHeadingTag.tsx
aaschlote: `querySelector` returns either a single HTMLHeadingElement or null, while `querySelectorAll` returns a list of HTMLHeadingElement instances. If there are no <h1> elements on the page, querySelectorAll simply returns an empty list. I chose to use querySelectorAll as a matter of personal preference.

app/components/__test__/useFocusFirstH1.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const { getByText } = render(<DummyComponentWithoutH1 />);
+
+    const element = getByText(TEXT_HEADING_LABEL);
+
+    await waitFor(() => {
+      expect(element.getAttribute("tabindex")).not.toBe("-1");
+      expect(element).not.toHaveFocus();
+    });
+  });
+});
```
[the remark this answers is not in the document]
aaschlote: the requirement is only to focus in `h1`, so I assume in case it doesn't have `h1` doesn't focus in anything. But in case the page doesn't have `h1` however with `h2` present, the page have a11y issue.

app/components/useFocusHeadingTag.tsx
m0dh4x: i would support @chohner. think it makes the code a bit cleaner cause you can get rid of the length check and the array access. especially in case of "there should be only one h1 at this page (a11y)". :)

m0dh4x (approved):
beside my comment on querSelector vs querySelectorAll discussion. LGTM :)

app/components/__test__/useFocusFirstH1.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    });
+  });
+
+  it("should not set tabindex and focus in case h1 element is not present", async () => {
+    const { getByText } = render(<DummyComponentWithoutH1 />);
+
+    const element = getByText(TEXT_HEADING_LABEL);
+
+    await waitFor(() => {
+      expect(element.getAttribute("tabindex")).not.toBe("-1");
```
pgurusinga, line 34: That will pass if getAttribute returns `null`, `''`, or even `'0'`, based your code you wanna do something like this no?
`expect(element).not.toHaveAttribute('tabindex');`

app/components/useFocusHeadingTag.tsx
aaschlote: Alright, I'll change it!

app/components/__test__/useFocusFirstH1.test.tsx
aaschlote: it'll pass, I can change to change the condition you mentioned it. Thanks!

pgurusinga (approved):

PR #2125, opened 2025-05-23
refactor(Report Problem API)

ingriddorioschulze:
This PR refactors the Report Problem logic to get the surveys from the api

TODOS:
- return just necessary json
- create post request and action


app/routes/api-posthog-survey-list.ts
```diff
@@ -0,0 +1,53 @@
...
+    id: survey.id,
+    questions: survey.questions,
+  };
+}
+
+export async function action({ request }: ActionFunctionArgs) {
+  try {
+    const body = await request.json();
+    const { surveyId, responses } = body;
+    posthog.capture("survey sent", {
```
aaschlote, line 39: can you check if this really works? Because you are using the `posthog` from the `posthog-js` in the server and this has been not initialised in this scope of the system. Maybe we can send the capture to the event using the API as well.

PR #2126, opened 2025-05-26
feat: add optional button to the Hero component

pgurusinga:
https://digitalservicebund.atlassian.net/browse/ZOV-11

### DIFF
--- a/app/components/Button.tsx
+++ b/app/components/Button.tsx
@@ -44,6 +44,7 @@ function Button({
       "ds-button-with-icon": iconLeft ?? iconRight,
       "ds-button-with-icon-only": (iconLeft ?? iconRight) && !children,
       "ds-button-full-width": fullWidth,
+      "is-disabled": props.disabled,
     },
     "contrast-more:border-4 forced-colors:border-4 border-solid contrast-more:border-black", // TODO: move into angie?
     props.className,
@@ -74,7 +75,7 @@ function Button({
     return (
       <a
         {...props}
-        href={href}
+        href={props.disabled ? undefined : href}
         className={buttonClasses}
         onKeyDown={onKeyDown}
         {...opts}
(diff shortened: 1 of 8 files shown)


app/services/cms/models/__test__/StrapiHero.test.ts
```diff
@@ -38,4 +38,73 @@ describe("StrapiHero", () => {
       __component: "page.hero",
     });
   });
+
+  it("should transform null buttons to undefined", () => {
```
chohner, line 42: I feel like we are testing zod library code here. Instead of testing the schema parsing I think component testing the rendering of the buttons would be more worthwhile imo

app/services/cms/models/StrapiHero.ts
```diff
@@ -10,6 +11,11 @@ export const StrapiHeroSchema = z
     heading: StrapiHeadingSchema,
     content: StrapiParagraphSchema.nullable().transform(omitNull),
     outerBackground: StrapiBackgroundSchema.nullable(),
+    buttons: z
```
aaschlote, line 14: it's not mentioned in the ticket, but I guess it should have the option to select only one button for this component. What about to do like this?

pgurusinga: In the ticket, it is mentioned implicitly, `let's people add buttons in Strapi.` as in plural

image

aaschlote (approved):
besides my comment, LGTM!

app/components/Hero.tsx
```diff
@@ -1,18 +1,30 @@
...
   return (
     <div className="container !pt-24 !pb-24 ds-stack ds-stack-16">
       <Heading {...heading} />
       {content && (
         <RichText className="ds-heading-03-reg" html={content.html} />
       )}
+      {arrayIsNonEmpty(buttons) && (
+        <ButtonContainer>
+          {buttons.map((button) => (
+            <Button key={button.text ?? button.href} {...button} />
```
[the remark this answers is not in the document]
pgurusinga: I'd keep it as it is right now, since we are using text most of the time, and the button in that button container component scope, I am sure that would be unique

app/services/cms/models/StrapiHero.ts
pgurusinga: I just clarified with @m0dh4x, and it's one single button, though there might be a possibility Hero could contain two or fewer, but since I was not there in the refinement, let's keep it as one button

m0dh4x (approved):
lgtm

PR #2137, opened 2025-05-27
Refactor get rid of ts enums

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-67


chohner:
nice! could you add an eslint rule for banning enums in the future?
see https://github.com/typescript-eslint/typescript-eslint/issues/561#issuecomment-496664453

PR #2139, opened 2025-05-27
Accessibility/sr for error for autosuggest

judithmh:
The fieldname and error were not read because of missing aria-describedby in the CustomInput. This solves the issue


app/components/inputs/autoSuggestInput/customComponents/CustomInput.tsx
```diff
@@ -1,22 +1,38 @@
...
-      }
-    }}
-    {...props}
-    maxLength={INPUT_CHAR_LIMIT}
-    aria-required={props.selectProps.className?.includes(
-      "auto-suggest-input-required",
-    )}
-  />
-);
+type CustomSelectProps = Props<
```
aaschlote, line 10: maybe we can add a comment here that they will add soon the `aria-describedby` in the props from this [PR](https://github.com/JedWatson/react-select/pull/6016) and we can get rid of this custom types here.

pgurusinga (approved):

PR #2158, opened 2025-06-02
(refactor): initialize posthog-js in new `useAnalytics()` hook

chohner:
- clearly separates frontend and backend posthog instances
- single point of initialization of posthog-js
- single `useAnalytics` hook that provides both `hasTrackingConsent` as well as `posthogInstance` (if available)


chohner:
closes https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/2125 and https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/2103

app/root.tsx
```diff
@@ -147,6 +150,15 @@ function App() {
   const matches = useMatches();
   const { breadcrumbs, title, ogTitle, description } = metaFromMatches(matches);
   const nonce = useNonce();
+  const [posthogClient, setPosthogClient] = useState<PostHog | undefined>();
+  useEffect(() => {
+    initPosthog(hasTrackingConsent)
+      .then((client) => {
+        setPosthogClient(client);
+      })
+      .catch(logError);
```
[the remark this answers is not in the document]
chohner: well - `.init()` might fail internally. but more importantly, eslint forces us to handle exceptions which I find quite reasonable in this case

app/root.tsx
```diff
@@ -147,6 +150,15 @@ function App() {
   const matches = useMatches();
   const { breadcrumbs, title, ogTitle, description } = metaFromMatches(matches);
   const nonce = useNonce();
+  const [posthogClient, setPosthogClient] = useState<PostHog | undefined>();
+  useEffect(() => {
```
aaschlote, line 154: I guess I'm already mentioned, but I would prefer to have this `useEffect` as custom hook in a different file, with this we can:

1. Decoupling this logic and keep the code cleaner 
2. Incorporate all the logic from `initPosthog` without call from a separated function.
3. Avoid to many `useEffects` calls in the root.tsx
4. We can easily add unit tests

app/services/analytics/fetchSurveys.ts
```diff
@@ -0,0 +1,9 @@
+import type { PostHog, Survey } from "posthog-js";
+
+export const fetchSurvey = (surveyId: string, posthogClient?: PostHog) => {
```
aaschlote, line 3: can we add an unit test for this function?

app/services/analytics/useAnalytics.tsx
```diff
@@ -0,0 +1,12 @@
+import { type PostHog } from "posthog-js";
+import { createContext, useContext } from "react";
+
+export const AnalyticsContext = createContext<{
```
aaschlote, line 4: I would prefer to have this context inside of the `components` folder due the `.tsx`. All the `.tsx` should be inside the `components` due the config we have to run the `vitest`. I know we don't have tests for this hook and we don't need, but I feel keeping `.tsx` outside of the `components` folder a bit odd.

app/services/analytics/useAnalytics.tsx
```diff
@@ -0,0 +1,12 @@
...
+  posthogClient: PostHog | undefined;
+  hasTrackingConsent: boolean | undefined;
+}>({
+  posthogClient: undefined,
+  hasTrackingConsent: undefined,
```
aaschlote, line 9: can't we init as `false`? I get confuse sometimes with the consent being `false`, `true` and `undefined`.

aaschlote (commented):
Nice code, I just feel we can do a improve the source code using a good patterns and add more unit tests to test the correct behaviour of the functionality.

app/services/analytics/useAnalytics.tsx
chohner: actually I have no idea why this is a tsx 🤔

app/components/reportProblem/Survey.tsx
```diff
@@ -43,8 +45,8 @@ export const PosthogSurvey = ({
   );
 
   const onFeedbackSubmitted = () => {
-    if (responses) {
-      posthog.capture("survey sent", {
+    if (responses && posthogClient) {
+      posthogClient.capture("survey sent", {
```
pgurusinga, line 49: I was experimenting with this survey and noticed something interesting. Is it intentional that you can submit an empty response after selecting and unselecting an option? If it's not by design, I think the issue might be that the initial state is set to undefined. However, when a selection is made, the app creates an object that looks like this:
```js
{
	$survey_response_uuid: ['Sonstiges']
}
```
and when you selected

```js
{
	$survey_response_uuid: ['']
}
```
and in the if block evaluates this as true

tile-output

app/components/cookieBanner/CookieBanner.tsx
```diff
@@ -23,49 +21,23 @@ type CookieBannerContentProps = {
...
-      });
-    }
-
-    if (posthogLoaded) {
-      if (!hasTrackingConsent && posthog.has_opted_in_capturing()) {
-        posthog.opt_out_capturing();
-      } else if (hasTrackingConsent && posthog.has_opted_out_capturing()) {
-        posthog.opt_in_capturing();
+    if (posthogClient) {
+      if (!hasTrackingConsent && posthogClient.has_opted_in_capturing()) {
```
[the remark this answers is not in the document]
chohner: Just to confirm - this is indeed called on manual opt-out after accepting the cookies

app/components/reportProblem/__test__/ReportProblem.test.tsx
```diff
@@ -1,60 +1,38 @@
...
-  useFeedbackTranslations: () => ({
-    "report-problem": reportProblem,
-    cancel: cancel,
-    "submit-problem": submitProblem,
-    "open-feedback-placeholder": "Beschreibung des Problems....",
-  }),
-}));
+describe("ReportProblem", () => {
+  it("should trigger the Survey popup", () => {
+    vi.mock("~/services/analytics/fetchSurveys");
```
aaschlote, line 9: same as here, you should mock the files outside the tests, better outside of the describe scope. If you move this test as last or just run the other tests as single, they will fail.

app/services/analytics/gdprCookie.server.ts
```diff
@@ -39,7 +39,7 @@ async function createTrackingCookie({
   cookie[acceptCookiesFieldName] =
     consent === undefined ? undefined : stringifiedConsentValue;
   sendCustomAnalyticsEvent({
-    eventName: "cookie consent given",
+    eventName: "clicked cookie banner",
```
aaschlote, line 42: just to be sure, this was already aligned with the Product or UX? Not sure if they are using this custom tracking for any report.

app/components/reportProblem/__test__/ReportProblem.test.tsx
```diff
@@ -1,58 +1,40 @@
...
-    });
-    const { getByRole } = render(<ReportProblem />);
-    const reportButton = getByRole("button");
-    expect(reportButton).toBeInTheDocument();
-    expect(reportButton.textContent).toBe(` ${reportProblem} `);
-    expect(reportButton.id).toBe("survey-button");
+  vi.mock("~/services/analytics/fetchSurveys");
+  vi.mock("~/services/analytics/useAnalytics");
+
+  vi.mocked(fetchSurvey).mockReturnValue({
```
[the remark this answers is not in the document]
chohner: `vi.mock()` is hoisted to the top of the file anyways. I will move them up just to avoid confusion

app/services/analytics/gdprCookie.server.ts
chohner: good eye and yes, this was aligned. but I'll pull this out just to make this more obvious

PR #2159, opened 2025-06-03
feat(Navigation Menu): Always expand when user reaches "Validation Site"

Spencer6497:
This PR auto-expands all subflows of the sidebar navigation menu if the user reaches the end of the flow with incomplete responses, for user clarity

- **add validationPage property to xStateConfig meta**
- **add logic to handle expanding subflows upon reaching a validationPage**


pgurusinga:
@Spencer6497 Was this logic aligned with the ZOV design? This is the first time I am hearing about it.

app/domains/beratungshilfe/formular/abgabe/xstateConfig.ts
```diff
@@ -14,6 +14,7 @@ export const abgabeXstateConfig = async (backDestination: string) => {
     states: {
       ueberpruefung: {
         on: { BACK: backDestination },
+        meta: { validationPage: true },
```
aaschlote, line 17: Maybe we could rename it to something like expandNavigationComponent? Since you're using it on a few validation pages, it might seem limited in scope — but in reality, this configuration can be applied to any page in the flow. Ultimately, its main purpose is to expand the navigation component, so a more descriptive name would help clarify its intent.

app/components/navigation/NavItem.tsx
```diff
@@ -39,7 +40,7 @@ export function NavItem({
   const isDisabled = stateIsDisabled(state);
   const isCurrent = stateIsCurrent(state);
   const isDone = stateIsDone(state);
-  const collapse = useCollapse({ defaultExpanded: isCurrent });
+  const collapse = useCollapse({ defaultExpanded: expanded ?? isCurrent });
```
chohner, line 43: This mixing of controlling and internal state might get a bit tricky. If the external `expanded` ever is `false`, it overwrites the `isCurrent` behaviour

app/components/navigation/types.ts
```diff
@@ -17,6 +18,7 @@ export type NavItem = {
   destination: string;
   label: string;
   state: NavState;
+  expanded?: boolean;
```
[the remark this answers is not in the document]
chohner: I think `NavItem` can either be a single menu item or contain a submenu. if it is a single item the prop `expanded` doesn't make any sense and shouldn't be set I think. for a future refactor, single items and submenus could be split but for now I think `undefined` might be ok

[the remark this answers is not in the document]
chohner: however, it could maybe be renamed to `forceExpanded`? because it overwrites the internal expansion logic from the outside

app/domains/beratungshilfe/formular/abgabe/xstateConfig.ts
aaschlote: but for this `special treatment` are you going to use the same config? So the config will apply more than one action? Like expand the navigation and more things?

chohner: Yes, thats how we specified it. But indeed renaming this flag once we add more things might be the less confusing.

aaschlote: I see your point, but I still think the naming might be confusing. When I see a configuration named `isValidationPage`, it sounds like it's responsible for validating the page. However, what it actually does is expand the navigation component — which isn't immediately obvious from the name.

app/components/navigation/types.ts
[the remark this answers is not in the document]
aaschlote: fair enough, I'm closing this thread!

PR #2168, opened 2025-06-04
[Fix] PKH and BerH page over-scroll

melgaafary:
Upload button is taking way more width than it needs

### DIFF
--- a/app/components/inputs/__test__/FileInput.test.tsx
+++ b/app/components/inputs/__test__/FileInput.test.tsx
@@ -43,7 +43,7 @@ describe("FileInput", () => {
     const input = getByTestId("file-upload-input-belege[0]");
     expect(input).toBeInTheDocument();
     expect(input).toHaveClass(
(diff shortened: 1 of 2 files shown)


Spencer6497 (approved):
great work! 🚢 it

app/components/inputs/__test__/FileInput.test.tsx
```diff
@@ -43,7 +43,7 @@ describe("FileInput", () => {
     const input = getByTestId("file-upload-input-belege[0]");
     expect(input).toBeInTheDocument();
     expect(input).toHaveClass(
-      "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary w-full",
+      "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary w-fit",
```
chohner, line 46: also - I would vote for not testing static CSS classes since they dont prevent any failure and just create double work 🙈

PR #2169, opened 2025-06-04
ZOV: Vorabcheck — Simplify guards

pgurusinga:
### READ ME 
I keep some yes/no guards because either they are both used or one answer is used multiple times.


aaschlote (approved):
LGTM!

PR #2173, opened 2025-06-05
chore(content validation): separate data collections [skip ci]

chohner:
moved for better discoverability


data/validation/allowedWebsites.server.ts
```diff
@@ -0,0 +1,66 @@
+// used by scripts/verifyExternalLinks
+import { bucketUrl } from "~/services/cms/bucketUrl";
+
+export const allowedWebsites = [
```
aaschlote, line 4: should we add this file in `data` path? Because all the files from this path will go to docker image and I guess we don't need the websites there.

PR #2174, opened 2025-06-05
Force to focus in the input when it has an error

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-214

### DIFF
--- a/app/components/inputs/autoSuggestInput/customComponents/CustomInput.tsx
+++ b/app/components/inputs/autoSuggestInput/customComponents/CustomInput.tsx
@@ -1,3 +1,5 @@
+import { useField } from "@rvf/react-router";
+import { useEffect } from "react";
 import {
   components,
   type InputProps,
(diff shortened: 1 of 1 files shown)


chohner:
what happens when there are multiple fields with errors?

pgurusinga:
@chohner 
If the submitted data contains errors, it is helpful to focus on the first <input> element that has an error based on this W3 [link](https://www.w3.org/WAI/tutorials/forms/notifications/?utm_source=chatgpt.com#listing-errors:~:text=If%20the%20submitted%20data%20contains%20errors%2C%20it%20is%20convenient%20to%20set%20the%20focus%20to%20the%20first%20%3Cinput%3E%20element%20that%20contains%20an%20error.)

and this library already does it

aaschlote (commented):
LGTM (I can't approve it, I created the PR 😅)

PR #2193, opened 2025-06-11
[Feature] add new strapi template variable hasErhöhung

melgaafary:
adds a new variable that fulfills these conditions :

Supervariable is TRUE, if one of the following variables is TRUE
  {{#hasErhoehungsbetrag}}
  {{#hasKindergeld}}
  {{#hasEinmalSozialleistung}}
  {{#hasNachzahlungSozialUnter500}}
  {{#hasNachzahlungArbeitUnter500}}
  {{#hasBuergergeld}}
  {{#hasGrundsicherung}}
  {{#hasAsylbewerberleistung}}
  {{#hasPflegegeldSelbst}}
  {{#hasPflegegeldFremd}}
  {{#hasArbeitsentgeltEinmalig}}
  {{#hasNachzahlungArbeitMehr500}}
  {{#hasNachzahlungSozialMehr500}}
  {{#isSelbststaendig}}
  {{#hasWohngeld}}


app/domains/kontopfaendung/wegweiser/stringReplacements.ts
```diff
@@ -220,3 +220,49 @@ export const getInfoZumPKontoStrings = (
     isPrivilegierteForderungUnterhalt;
   return { infoZumPKontoIsVisible };
 };
+
+export const getHasErhöhungStrings = (
```
chohner, line 224: (I know you're just following the existing pattern)

In a follow-up PR we should move from returning objects (which makes re-using these very annoying) to returning a `bool`. The property name can simply be specified in `kontopfaendungWegweiser.stringReplacements`

melgaafary: yea was thinking about it as well. Will do it in a follow up PR, because Torsten needs this

PR #2197, opened 2025-06-12
ZOV: Add CSRF Protection to FGR Loader

pgurusinga:
## Problem
Our airports endpoint provides information about thousands of airports, and any HTTP request can be made using tools like `curl`, which can lead to high resource consumption.

## Changes
- CSRF token validation has been integrated into the `createDataListLoader` through `validateCsrfSessionFormless`, ensuring that only users who have previously visited the page can access the endpoint.
- Requests without valid CSRF tokens will receive a 403 Forbidden response
- Added error logging for failed CSRF validation attempts


aaschlote (approved):
LGTM!

PR #2205, opened 2025-06-13
ZOV: Add handling for additional cases in various functions

pgurusinga:
Fixing ESlint warning caused by the switch exhaustiveness check rule


app/components/PageContent.tsx
```diff
@@ -52,6 +52,7 @@ function wrapInBackground(
 }
 
 function cmsToReact(strapiContent: StrapiContentComponent) {
+  // eslint-disable-next-line @typescript-eslint/switch-exhaustiveness-check
```
pgurusinga, line 55: The issue is caused by `page.array-summary`, which is filtered by the `PageContent` function

pgurusinga: I will make it fall through same as the default

aaschlote (approved):
LGTM!

PR #2207, opened 2025-06-13
[Refactor] FileInput refactor and bug fix

melgaafary:
Current: When you hover over file upload component the cursor doesnt turn into cursor

Also refactored out some unused divs and an issue when resizing the screen and the file name is too long it would break the whole container down

### DIFF
--- a/app/components/filesUpload/FileUploadInfo.tsx
+++ b/app/components/filesUpload/FileUploadInfo.tsx
@@ -32,12 +32,10 @@ export const FileUploadInfo = ({
 
   return (
     <div className={classes} data-testid={`file-upload-info-${inputName}`}>
-      <div className="max-w-full flex justify-between items-center">
+      <div className="max-w-full grid grid-cols-[24px_1fr_auto] gap-x-12 items-center">
         <InsertFileIcon className="shrink-0 fill-gray-900" aria-hidden="true" />
-        <p className="ds-body-01-reg text-black mr-8 ml-10 truncate">
-          {file.filename}
-        </p>
-        <p className="ds-body-01-reg text-gray-900">
+        <p className="ds-body-01-reg text-black truncate">{file.filename}</p>
+        <p className="ds-body-01-reg text-gray-900 text-16">
           {formatFileSizeToString(file.fileSize)}
         </p>
       </div>
--- a/app/components/inputs/FileInput.tsx
+++ b/app/components/inputs/FileInput.tsx
@@ -29,33 +29,31 @@ export const FileInput = ({
   const errorId = `${name}-error`;
 
   const inputClasses = classNames(
-    "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary file:ds-button-large w-fit",
-    {
-      "w-0.1 h-0.1 opacity-0 overflow-hidden absolute z-0 cursor-pointer":
-        jsAvailable,
-    },
+    jsAvailable
+      ? "hidden"
+      : "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary file:ds-button-large w-fit file:cursor-pointer",
   );
 
   const FileInput = (
     <input
-      name={jsAvailable ? undefined : name}
-      onChange={(event) => {
-        const file = event.target.files?.[0];
-        void onFileUpload(name, file);
-      }}
+      id={name}
+      name={name}
       type="file"
       accept=".pdf"
       data-testid={`file-upload-input-${name}`}
       aria-invalid={error !== undefined}
       aria-errormessage={error && errorId}
       className={inputClasses}
-      {...(selectedFile ? { value: selectedFile.filename } : {})}
+      onChange={(event) => {
+        const file = event.target.files?.[0];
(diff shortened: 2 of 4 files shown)


app/components/inputs/FileInput.tsx
```diff
@@ -29,33 +29,31 @@ export const FileInput = ({
...
 
   const inputClasses = classNames(
-    "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary file:ds-button-large w-fit",
-    {
-      "w-0.1 h-0.1 opacity-0 overflow-hidden absolute z-0 cursor-pointer":
-        jsAvailable,
-    },
+    jsAvailable
+      ? "hidden"
+      : "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary file:ds-button-large w-fit file:cursor-pointer",
```
chohner, line 34: I think the `classname` way to write this would be:

```typescript
classNames(
{
  "hidden": jsAvailable,
  "body-01....": !jsAvailable
})
```

however, I think simply adding `hidden` on `jsAvailable` might do the same?

```typescript
classNames("body-01....",
{
  "hidden": jsAvailable,
})
```

app/components/inputs/FileInput.tsx
```diff
@@ -29,33 +29,31 @@ export const FileInput = ({
...
   );
 
   const FileInput = (
     <input
-      name={jsAvailable ? undefined : name}
-      onChange={(event) => {
-        const file = event.target.files?.[0];
-        void onFileUpload(name, file);
-      }}
+      id={name}
```
chohner, line 39: I dont think we need the `id` here?

app/components/inputs/FileInput.tsx
chohner: also, I think a short comment why this is hidden on `jsAvailable` might be helpful for the future

app/components/inputs/FileInput.tsx
```diff
@@ -71,37 +69,46 @@ export const FileInput = ({
           deleteButtonLabel={translations.fileUpload.delete.de}
           hasError={!!error}
         />
+        {error && (
+          <InputError id={errorId}>
+            {errorMessages?.find((err) => err.code === error)?.text ?? error}
+          </InputError>
+        )}
```
chohner, line 76: I'm slightly surprised that `InputError` belongs to `FileUploadInfo` block - I think before it was always part of the component?

chohner (approved):
no blockers, just a few nitpicks :)
oh and for visual changes, quick before/after screenshots are super helpful

app/components/inputs/FileInput.tsx
```diff
@@ -29,33 +29,31 @@ export const FileInput = ({
...
-      }}
+      id={name}
+      name={name}
       type="file"
       accept=".pdf"
       data-testid={`file-upload-input-${name}`}
       aria-invalid={error !== undefined}
       aria-errormessage={error && errorId}
       className={inputClasses}
-      {...(selectedFile ? { value: selectedFile.filename } : {})}
```
Spencer6497, line 52: this strange syntax is actually required to fix a bug where users weren't able to upload a file, delete it, and re-upload it. See [this bug](https://app.asana.com/1/1146679785927162/project/1203259475859667/task/1210439986750987?focus=true) and the following screen recording: 

https://github.com/user-attachments/assets/ec8caf50-82e4-4641-88e5-2d5b23f23067

Spencer6497 (commented):
A few things I found:
* JS validation upon "Weiter" button click is no longer working, i.e. required, empty file uploads aren't being validated. Nothing seems to happen in the UI.
* Attempting to click the "Upload" button in No-JS mode for an empty file upload results in a 500 server error, instead of displaying the errored UI alongside the file upload component (see screen recording below) 
 
https://github.com/user-attachments/assets/f84c27d8-ec4b-4c53-a493-35aee92bf42d

* Incorrect error string is returned when the user attempts to upload an invalid file for the optional file upload
Bildschirmfoto 2025-06-16 um 11 29 30

app/components/filesUpload/FileUploadInfo.tsx
```diff
@@ -32,12 +32,10 @@ export const FileUploadInfo = ({
...
     <div className={classes} data-testid={`file-upload-info-${inputName}`}>
-      <div className="max-w-full flex justify-between items-center">
+      <div className="max-w-full grid grid-cols-[24px_1fr_auto] gap-x-12 items-center">
         <InsertFileIcon className="shrink-0 fill-gray-900" aria-hidden="true" />
-        <p className="ds-body-01-reg text-black mr-8 ml-10 truncate">
-          {file.filename}
-        </p>
-        <p className="ds-body-01-reg text-gray-900">
+        <p className="ds-body-01-reg text-black truncate">{file.filename}</p>
+        <p className="ds-body-01-reg text-gray-900 text-16">
```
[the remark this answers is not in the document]
melgaafary: so the font size in design is 16px, but in the code its 18. that's where text-16 is coming from. I will replace it!

PR #2218, opened 2025-06-17
bug(Kontopfandung): p-konto page flow

ingriddorioschulze:
This PR fixes the bug in the p-konto page flow. User selects yes or no on `p-konto` page and goes to `wegweiser/zwischenseite-unterhalt` page


Spencer6497 (approved):
could you add some flow tests?:)

PR #2221, opened 2025-06-18
feat(PKH): Vereinfachte Erklaerung [Pt. 2]

Spencer6497:
This PR completes the PKH Vereinfachte Erklaerung flow, complete with PDF logic and tests.

- **rename unterhaltLeisteIch -> unterhaltLeisteIchAnderePerson**
- **groundwork for Vereinfachte Erklaerung**
- **always skip RSV when unterhalt is paid to another person**
- **PR fixes**
- **rename empfaenger enum option to their english equivalents for clarity**
- **Correctly attach anhang in the case of a non-digital submission**
- **tweak Angehoerige pdf filler function to set up for VE, add Hinweis page**
- **test fixes**
- **add zahlender Unterhalt section and minderjaehrig question**
- **add Geburtsdatum field**
- **add flow test cases**
- **build out flow through Kind - Einnahmen**
- **fix user file upload bug**
- **build out flow through vermoegen**
- **add comprehensive flow tests**
- **add guard tests**
- **complete VE flow**
- **finish doneFunction, add unit tests**
- **add stringReplacement for conditional display of InlineNotice, add qualifiesForVE guard for PDF logic**
- **remove excess unit tests**
- **correctly wire up VE to RSV, use reusable kinderSchema and move unterhaltsSumme into it to avoid context collision, add flow transition test cases, fix bug with unterhalt pdf filler, add entered child to pdf child array**
- **add VE to Anhang when applicable**


app/domains/shared/services/pdf/unterhaltHelpers.ts
```diff
@@ -29,6 +31,20 @@ export function getFillUnterhalt<
     const attachment: AttachmentEntries = [];
     const zahltPartnerUnterhalt = userData.partnerUnterhaltsSumme !== undefined;
     const kinder = userData.kinder ?? [];
+    const pkhUserData = userData as ProzesskostenhilfeFormularUserData;
+    // Need to add the child added as a part of Vereinfachte Erklärung to the kinder array
+    if (
+      empfaengerIsChild({ context: pkhUserData }) &&
+      pkhUserData.child !== undefined
+    ) {
+      kinder.unshift({
```
melgaafary, line 40: I generally like to avoid mutations, since we use userData directly that could cause bugs elsewhere.

Im not sure if we do this here intentionally or not. But just putting my 2 cents, no need to change anything

app/domains/prozesskostenhilfe/services/pdf/util.ts
```diff
@@ -31,6 +31,23 @@ export const getTotalMonthlyFinancialEntries = (
...
+  financialEntry: FinancialEntry,
+) => {
+  const betragNumber = Number(
+    financialEntry.betrag.replaceAll(".", "").replace(",", "."),
+  );
+  switch (financialEntry.zahlungsfrequenz) {
+    case "monthly":
+      return betragNumber;
+    case "quarterly":
+      return betragNumber / 3;
```
melgaafary, line 44: I could be not seeing something here but shouldnt this be /4?

app/domains/prozesskostenhilfe/services/pdf/util.ts
```diff
@@ -31,6 +31,23 @@ export const getTotalMonthlyFinancialEntries = (
       maximumFractionDigits: 2,
     });
 
+export const getFinancialEntryMonthlyAverage = (
+  financialEntry: FinancialEntry,
+) => {
+  const betragNumber = Number(
```
melgaafary, line 37: should you also have !isNan guard here? or do we have validations elsewhere?

ingriddorioschulze (approved):
Great work! :)

app/domains/prozesskostenhilfe/services/pdf/util.ts
Spencer6497: yeah the validation would have already happened during the flow, and since this function is only reachable from the PDF, the user wouldn't have been able to make it here without a valid number

app/domains/shared/services/pdf/unterhaltHelpers.ts
chohner: I think this might warrant a code comment 😇

PR #2222, opened 2025-06-18
[Refactor] FileInput refactor and bug fixes

melgaafary:
Bug 1: When you hover over file upload component the cursor doesnt turn into cursor
Bug 2: When resizing the screen with a big file name the whole section breaks 
Bug 3: Can't upload multiple files in a row - more than 2

Also refactored  the component to be more semantic

### DIFF
--- a/app/components/inputs/FileInput.tsx
+++ b/app/components/inputs/FileInput.tsx
@@ -30,13 +30,14 @@ export const FileInput = ({
 
   const inputClasses = classNames(
     jsAvailable
-      ? "w-[100%] h-[100%] left-0 top-0 opacity-0 overflow-hidden absolute z-0 cursor-pointer"
+      ? "w-0 h-0 opacity-0 overflow-hidden absolute z-0 cursor-pointer"
       : "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary file:ds-button-large w-fit file:cursor-pointer",
   );
 
   const FileInput = (
     <input
       name={jsAvailable ? undefined : name}
+      id={`${name}-input`}
       type="file"
       accept=".pdf"
       value={selectedFile?.savedFileKey || undefined}
@@ -74,7 +75,7 @@ export const FileInput = ({
           {jsAvailable ? (
             <>
               <label
-                htmlFor={name}
+                htmlFor={`${name}-input`}
                 className="relative inline-flex items-center ds-button ds-button-tertiary ds-button-large cursor-pointer w-fit"
               >
                 {FileInput}
--- a/app/components/inputs/FileInput.tsx
+++ b/app/components/inputs/FileInput.tsx
@@ -29,32 +29,30 @@ export const FileInput = ({
   const errorId = `${name}-error`;
 
   const inputClasses = classNames(
-    "body-01-reg m-8 ml-0 file:ds-button file:ds-button-tertiary file:ds-button-large w-fit",
-    {
-      "w-0.1 h-0.1 opacity-0 overflow-hidden absolute z-0 cursor-pointer":
-        jsAvailable,
-    },
+    jsAvailable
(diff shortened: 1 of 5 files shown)


ingriddorioschulze (approved):
Cool! Great job :)

Spencer6497 (approved):
Excellent work! 🤩 there's still a bug when, in the no-JS case, a user clicks "Upload" without a file selected and the server sends back a non-prettified 500 (instead of a validation error), but aside from that great job!

app/components/inputs/FileInput.tsx
```diff
@@ -29,32 +29,31 @@ export const FileInput = ({
...
   );
 
   const FileInput = (
     <input
       name={jsAvailable ? undefined : name}
-      onChange={(event) => {
-        const file = event.target.files?.[0];
-        void onFileUpload(name, file);
-      }}
+      id={`${name}-input`}
```
chohner, line 40: maybe we could extract the `inputId` into the function body to re-use in the `label`?

app/components/inputs/FileInput.tsx
```diff
@@ -29,32 +29,31 @@ export const FileInput = ({
...
       type="file"
       accept=".pdf"
+      value={selectedFile?.savedFileKey || undefined}
```
chohner, line 43: I would agree with gh here (`Prefer using nullish coalescing operator`)

app/components/inputs/FileInput.tsx
```diff
@@ -29,32 +29,31 @@ export const FileInput = ({
...
       data-testid={`file-upload-input-${name}`}
```
chohner, line 44: do we need a `test-id` if we have a real id?

chohner (approved):
have not tried it but the cleanup looks great!

app/services/upload/fileUploadHelpers.server.ts
```diff
@@ -38,6 +42,20 @@ export async function uploadUserFile(
   const inputName = formAction.split(".")[1];
   const { fieldName, inputIndex } = splitFieldName(inputName);
   const file = await parseFileFromFormData(request, inputName);
+
+  if (!file) {
+    return {
+      validationError: validationError(
+        {
+          fieldErrors: {
+            [formAction.split(".")[1]]: "fileRequired",
```
Spencer6497, line 51: could we save `"fileRequired"` somewhere in a global const?

app/services/upload/fileUploadHelpers.server.ts
```diff
@@ -50,11 +68,6 @@ export async function uploadUserFile(
    */
   const fileArrayBuffer = await file.arrayBuffer();
   const fileMeta = createFileMeta(file, fileArrayBuffer);
-
-  /**
-   * Need to scope the context, otherwise we validate against the entire context,
-   * of which we only have partial data at this point
-   */
```
Spencer6497, line 57: why did we remove this comment?

Spencer6497 (approved):
thank god it's over 🫡 👍

app/components/inputs/FileInput.tsx
```diff
@@ -7,6 +7,7 @@ import { splitFieldName } from "~/services/upload/splitFieldName";
 import { type PDFFileMetadata } from "~/util/file/pdfFileSchema";
```
aaschlote, line 7: can we move this file to the path `app/components/inputs/filesUpload`?

aaschlote: ok, but can we move the `FileInput.tsx` to `app/components/inputs/filesUpload`?

melgaafary: you mean move the `app/components/filesUpload` folder to `app/components/inputs/filesUpload` and move `FileInput.tsx ` there too?

because right now `filesUpload` folder isnt `inputs`

PR #2223, opened 2025-06-18
Refactor formular server 2

aaschlote:
### Changes

- Fix: Get correct accessibility translations for `arial-label` FlowNavigation
- Move buttons translations to back and next to source code
- Move functions `getUserDataAndFlow`, `getPageAndFlowDataFromPathname`, `getUserPrunedDataFromPathname` and `validateStepIdFlow` to `service/flow`
- Create function `retrieveContentData` to process all the content data from CMS and return function `getContentData`
- Create function `getContentData` to return all the content data necessary to deliver to client-side 

Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-310


app/routes/shared/formular/contentData/getContentData.ts
```diff
@@ -0,0 +1,119 @@
...
+  migrationData: UserData | undefined,
+  stringTranslations: Translations,
+  translations: Record<string, Translations>,
+  userDataWithPageData: UserData & {
+    pageData: {
+      arrayIndexes: number[];
+    };
+  },
+) => {
+  return {
```
pgurusinga, line 30: is there a reason why all the methods are grouped into a single returned object? I get that it keeps things organized, but we don't want to have this below in each file no?
```js
vi.mock("this module", () => {
 arraySummaryData: vi.fn()
...
 getButtonNavigationProps: vi.fn()
}
```

app/routes/shared/formular.server.ts
```diff
@@ -60,91 +47,33 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
...
-
-  const arrayCategories = formPageContent.pre_form
-    .filter(isStrapiArraySummary)
-    .map((strapiSummary) => strapiSummary.category);
-
-  const arraySummaryData = getArraySummaryData(
-    arrayCategories,
-    flowController.getRootMeta()?.arrays,
+  const contentData = await buildContentData(
+    request,
```
pgurusinga, line 51: let's not pass around request object, just send what is needed. this function requires `request.url` and `request.headers` only, where `getPageAndFlowDataFromRequest` needs path parameter from request.url

app/routes/shared/formular.server.ts
```diff
@@ -60,91 +47,33 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
...
     userDataWithPageData,
   );
 
-  const migrationData = await getMigrationData(
+  const stringTranslations = contentData.getStringTranslations();
+  const translations = contentData.getTranslations();
+  const cmsContent = contentData.getCMSContent();
+  const formElements = contentData.getFormElements();
+  const meta = contentData.getMeta(currentFlow);
+  const migration = contentData.getMigration(currentFlow);
```
pgurusinga, line 61: is getMigration actually part of content? is flow-related business logic, right? if so, I think it should not use word  `contentData`

app/routes/shared/formular.server.ts
aaschlote: I liked the idea, I changed it!

app/routes/shared/formular/contentData/getContentData.ts
aaschlote: this was just the first draft, I think a few functions I'll extract to another files and have a test there.

app/services/flow/formular/contentData/buildFormsElements.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const buildFormsElements = (cmsContent: CMSContent) => {
+  return cmsContent.formContent.map((strapiFormElement) => {
+    if (
+      isStrapiSelectComponent(strapiFormElement) &&
+      strapiFormElement.label === null &&
+      cmsContent.heading !== undefined
+    )
+      strapiFormElement.altLabel = cmsContent.heading;
+    return strapiFormElement;
```
judithmh, line 12: Maybe  check first if !isStrapiSelect -> return element else check for label and heading and return el with new label and heading

app/services/flow/formular/contentData/getContentData.ts
```diff
@@ -0,0 +1,100 @@
...
+      flowController: ReturnType<typeof buildFlowController>,
+    ) => {
+      return getArraySummaryData(
+        arrayCategories,
+        flowController.getRootMeta()?.arrays,
+        userDataWithPageData,
+      );
+    },
+    getFormElements: () => {
+      return buildFormsElements(cmsContent);
```
judithmh, line 41: maybe renaming to buildFormElements

app/services/flow/formular/contentData/getContentData.ts
```diff
@@ -0,0 +1,122 @@
...
+    getButtonNavigationProps: (
+      flowController: ReturnType<typeof buildFlowController>,
+      pathname: string,
+      stepId: string,
+      arrayIndexes: number[] | undefined = [],
+    ) => {
+      const backDestination = flowController.getPrevious(stepId);
+
+      const backDestinationWithArrayIndexes =
+        backDestination && arrayIndexes
```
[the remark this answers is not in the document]
aaschlote: yeah, this is legacy. I'll change it. Thanks for pointing it!

app/services/flow/getUserPrunedDataFromPathname.ts
```diff
@@ -1,11 +1,13 @@
...
 import { pruneIrrelevantData } from "~/services/flow/pruner";
-import { getSessionData } from "~/services/session.server";
-import { getPageAndFlowDataFromRequest } from "./getPageAndFlowDataFromRequest";
+import { type CookieHeader, getSessionData } from "~/services/session.server";
+import { getPageAndFlowDataFromPathname } from "./getPageAndFlowDataFromPathname";
 
-export const getUserPrunedDataFromRequest = async (request: Request) => {
-  const { flowId, arrayIndexes } = getPageAndFlowDataFromRequest(request);
-  const cookieHeader = request.headers.get("Cookie");
+export const getUserPrunedDataFromPathname = async (
```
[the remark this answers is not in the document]
aaschlote: yeah, I can do it. Thanks for the suggestion!

Spencer6497 (approved):
so much better now 🤩 btw you should win an award for unit testing 🥇

app/services/flow/__test__/getPageAndFlowDataFromPathname.test.ts
```diff
@@ -0,0 +1,13 @@
+import { getPageAndFlowDataFromPathname } from "../getPageAndFlowDataFromPathname";
+
+describe("getPageAndFlowDataFromPathname", () => {
+  it("should extract flowId, stepId, and arrayIndexes from request URL", () => {
+    const mockPathname = "/fluggastrechte/formular/stepId1/2";
```
pgurusinga, line 5: has malformed URL been tested? such as `/fluggastrechte/formular/////\\//stepId1/2`

pgurusinga (approved):
I think the loader is much better now ✅

app/services/flow/__test__/getPageAndFlowDataFromPathname.test.ts
aaschlote: I can add this test in `parsePathname`, because there does all the logic of the extraction. I created the function `getPageAndFlowDataFromPathname` just to get all the `flow` and `page` info that it's needed.

app/routes/shared/formular.server.ts
```diff
@@ -48,110 +34,42 @@ export const loader = async ({ params, request }: LoaderFunctionArgs) => {
...
-
-  const migrationData = await getMigrationData(
+  const translations = contentData.getTranslations();
+  const navItems = contentData.getNavItems(flowController, stepId);
+  const cmsContent = contentData.getCMSContent();
+  const formElements = contentData.getFormElements();
+  const meta = contentData.getMeta();
+  const arraySummaryData = contentData.arraySummaryData(flowController);
+  const stepData = contentData.getStepData();
+  const buttonNavigationProps = contentData.getButtonNavigation(
```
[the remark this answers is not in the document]
aaschlote: I was considering implementing it, but after retrieving the actual data from the CMS, several additional steps were required. These include generating button navigation, extracting meta content by comparing the parent and current metadata, creating the summary data array, and building the form data—which includes some logic for accessibility in the radio group, among other things.

To keep the main function lightweight, I broke the logic down into smaller, decoupled functions. This makes it easier to retrieve the necessary data, write tests, and significantly reduces the complexity we had before.

app/services/flow/getPageAndFlowDataFromPathname.ts
```diff
@@ -1,8 +1,7 @@
 import { parsePathname } from "~/domains/flowIds";
 import { flows } from "~/domains/flows.server";
 
-export const getPageAndFlowDataFromRequest = (request: Request) => {
-  const { pathname } = new URL(request.url);
+export const getPageAndFlowDataFromPathname = (pathname: string) => {
```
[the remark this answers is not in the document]
aaschlote: I was considering that, but the `currentFlow` value is needed in several places within `formular.server`. I wanted a function that could return it directly, without having to repeatedly import `flows` from `~/domains/flows.server`.

Also, since `parsePathname` is already being used on the front end—in the `schemaForFieldNames` function, I decided to create a separate function for this purpose. Otherwise, it could lead to conflicts or issues.

aaschlote:
> have to say - this is passing what I can realistically review.
> 
> overall, I am not sure whether moving functions into nested function helps with the complexity of the long loader, especially if a massive object of functions is returned.
> 
> I think one function per transmitted datapoint is a pretty good rule of thumb and helps with modularity and testing. I fear I now need to dig through 2-3 functions just to find out where data is generated for my route.
> 
> love the additional tests though!

I think we can revisit the nested function structure later, but the main goal of this PR was to decouple the logic from `formular.server`. This allows us to propose a clearer and more focused function design, although I believe `retrieveContentData` currently does a bit more than its name suggests.

Additionally, this change makes it easier to write tests without requiring excessive mocking or dealing with a large number of parameters.

The sub-functions within getContentData don’t fully depend on the data from formular.server.ts (aside from the `CMSContent` type), so there’s potential to reuse them in other loaders like `result.server` or `vorabcheck.server`.

PR #2228, opened 2025-06-19
feat(Accordion): open details on print

chohner:
When printing, users should also get the content of all accordions.
this PR:
- adds a new `useShouldPrint` hook, to be used in components in case they need to implement special behavior for printing
- opens all accordion `details` in case that hook returns `true`
- adds `renderWithRouter()` testing util, for rendering a component on a specific route


app/services/useShouldPrint.ts
```diff
@@ -0,0 +1,3 @@
+import { useSearchParams } from "react-router";
+// useSearchParams() returns [searchParams, setSearchParams], we only want to check the collection
+export const useShouldPrint = () => useSearchParams()[0].has("print");
```
aaschlote, line 3: can we move this to the `component` folder and rename to `.tsx`? Because the `useSearchParams` is a hook using dom elements. Check it [here](https://github.com/remix-run/react-router/blob/37c7022110a60bb15fb4af832f4e4f1f8d209307/packages/react-router/lib/dom/lib.tsx#L1397).

Can you add tests as well?

aaschlote: yeah, it's a custom hook, but the `useSearchParams` uses `JSX`. If you add tests for this custom hook, you might see that it fails.

app/components/__test__/renderWithRouter.tsx
```diff
@@ -0,0 +1,5 @@
+import { render } from "@testing-library/react";
+import { MemoryRouter } from "react-router";
+
+export const renderWithRouter = (reactNode: React.ReactNode, route = "/") =>
```
pgurusinga, line 4: @aaschlote Do we no longer have the `createRemixStub` function in React Router v7? :O ref: https://remix.run/docs/en/main/utils/create-remix-stub

pgurusinga (approved):

PR #2230, opened 2025-06-20
feat(Report Problem): enable report problem feature 

ingriddorioschulze:
This PR enables report a problem feature on `/kontopfaendung/wegweiser`, `/beratungshilfe/vorabcheck` and `/beratungshilfe/zustaendiges-gericht` flows. 

Screenshot 2025-06-20 at 13 17 46
Screenshot 2025-06-20 at 13 17 33
Screenshot 2025-06-20 at 13 17 17


app/routes/shared/components/VorabcheckPage.tsx
```diff
@@ -15,13 +15,18 @@ export function VorabcheckPage() {
...
+  const showReportProblem =
+    flowId === "/beratungshilfe/vorabcheck" ||
+    flowId === "/kontopfaendung/wegweiser";
+
   useFocusFirstH1();
 
   return (
-    <Background backgroundColor="blue">
-      <div className="min-w-[100vw]">
+    <div className="min-w-[100vw] flex flex-col bg-blue-100">
```
chohner, line 28: I'm surprised why `<Background>` isn't working here... I would prefer to make it fit more used-cases rather than removing it and spreading tailwind background colors

ingriddorioschulze: I also don't get it. I tried to keep it but somehow the report problem button is no visible.

app/routes/shared/vorabcheck.server.ts
```diff
@@ -123,6 +123,7 @@ export const loader = async ({
       meta,
       progressProps,
       buttonNavigationProps,
+      flowId,
```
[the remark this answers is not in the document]
ingriddorioschulze: What would be the difference between passing flowId or shouldShowReportProblem? (Just out of curiosity)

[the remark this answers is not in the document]
chohner: There are several reasons for me:

a) It moves this (static) decision into the server, which is where I would look for this. At some point, this might even live in the root loader to be shared across all routes
b) It simplifies the client code and is (slightly) better for performance (less data transmission, less client code, less chance to re-render)
c) it simplifies testing (bool conditions are simpler to test than text conditions)
d) `flowId` could already be retrieved in the front-end from the pathname, so passing that information seems redundant

pgurusinga:
@ingriddorioschulze  

Quick question about the **zuständiges Amtsgericht** flow: Why does it only appear on the first page? What happens in these cases:  
#### 1. When the correct Amtsgericht is found: `/beratungshilfe/zustaendiges-gericht/ergebnis/14059`
image

#### 2. When an address suggestion is shown: `/beratungshilfe/zustaendiges-gericht/auswahl/20457`
image

app/routes/shared/showReportProblem.ts
```diff
@@ -0,0 +1,10 @@
+import { type FlowId } from "~/domains/flowIds";
+
+export function shouldShowReportProblem(flowId: FlowId): boolean {
```
aaschlote, line 3: maybe we can move this file to `app/services/flow`, because the `app/routes/shared` is related more react-router things and this looks like more flow logic.

app/routes/shared/__test__/reportProblem.test.tsx
```diff
@@ -0,0 +1,176 @@
+// @vitest-environment jsdom
```
aaschlote, line 1: how hard it would be if instead of the `reportProblem.test.tsx` if we test the `FormFlowPage.tsx` and `VorabcheckPage.tsx` components?

PR #2241, opened 2025-06-24
refactor(Tiles): move transform into schema

chohner:
just came across this slightly complex schema and cleaned up the transformation a little bit.

I would also propose to delete the `StrapiTile.test.ts` since it now only tests uninteresting zod library code

### DIFF
--- a/app/components/inputs/tile/TileRadio.tsx
+++ b/app/components/inputs/tile/TileRadio.tsx
@@ -91,7 +91,9 @@ function TileRadio({
             <TileTag tagDescription={tagDescription} />
           </div>
           <span className="ds-label-01-bold">{title}</span>
-          {description && <RichText html={description} />}
+          {description && (
+            <RichText className="ds-subhead" html={description} />
+          )}
         </div>
       </label>
     </div>
--- a/app/services/validation/__test__/richtext.test.ts
+++ b/app/services/validation/__test__/richtext.test.ts
@@ -1,6 +1,5 @@
 import { type SafeParseError } from "zod";
 import { listRenderer } from "~/services/cms/models/StrapiList";
-import { createTileRenderer } from "~/services/cms/models/StrapiTile";
 import { buildRichTextValidation } from "~/services/validation/richtext";
 
 describe("richtext validation", () => {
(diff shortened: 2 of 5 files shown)


app/components/inputs/tile/TileRadio.tsx
```diff
@@ -90,8 +89,14 @@ function TileRadio({
             )}
             <TileTag tagDescription={tagDescription} />
           </div>
-          <span className="ds-label-01-bold">{title}</span>
-          {description && <RichText html={description} />}
+          <div>
+            <p className="ds-label-01-bold">{title}</p>
+            {description && (
+              <p id={descriptionId} className="ds-subhead ds-body-01-reg">
+                {description}
```
pgurusinga, line 96: Why do you remove the rich text? Would you change the Strapi too?

app/services/validation/__test__/richtext.test.ts
```diff
@@ -46,17 +45,6 @@ describe("richtext validation", () => {
   });
 
   describe("custom renderers", () => {
-    test("uses custom tile renderer", () => {
-      const paragraphText = "Some paragraph";
-      const tileRenderer = createTileRenderer("some-value");
-      expect(
-        buildRichTextValidation(tileRenderer).safeParse(paragraphText),
```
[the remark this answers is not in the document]
chohner: was this used by tiles? maybe makes sense to do that in a separate commit?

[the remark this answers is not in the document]
aaschlote: it was added only for the tiles, as we don't use the richtext anymore, we can get rid of it as well.

app/components/inputs/tile/TileRadio.tsx
aaschlote: Let's at least make the PR for Strapi before to merge this changes, so we don't forget it.

chohner: does anyone know what happens if I just change the field type to string? will the data disappear? 😢

aaschlote: In the App or in Strapi? In the App, it's ok, but if you add a markdown in the content, it'll print a markdown.

pgurusinga: luckily only 2 places with tiles

PR #2245, opened 2025-06-25
Refactor formular action upload files

aaschlote:
### Changes

- Create function `handleFileUpload` to handle upload and delete files. Add tests
- Change functions `uploadUserFile` and `buildFileUploadError` to return the type `ValidationErrorResponseData` when an error is presented instead of the function `validationError`
- Remove props `deleteButtonLabel` of the component `FileUploadInfo` and call inside the component the translation

### Disclaimer
It was not possible to create a new router to handle files due the following reasons:

- The `FileUpload` component is wrapped in a `form` element, and since nesting form elements is not allowed in HTML, we can’t place another `form` inside it.
- Due to restrictions for users without JavaScript, we can only trigger a new route using either a `form` submission or an `href` link. However, using a `form` is not viable because of the nesting limitation mentioned above. Using an `href` isn’t straightforward either, since the router interprets href navigation as a loader call, whereas the file upload process requires an action. Moreover, files can’t be submitted via `href`, as there's no mechanism to include file data in such requests.

Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-310

### DIFF
--- a/app/components/TableOfContents.tsx
+++ b/app/components/TableOfContents.tsx
@@ -30,13 +30,13 @@ const TableOfContents = ({
             {links.map((link) => (
               <li key={link.text ?? link.url}>
                 <StandaloneLink
-                  className="visited:text-black !text-black"
+                  className="visited:text-black !text-black flex"
                   url={link.url}
                   text={link.text ?? ""}
                   icon={
-                    <ArrowDownward className="h-[1em] w-[1em] inline-block" />
+                    <ArrowDownward className="h-[1em] w-[1em] shrink-0 my-[0.25em]" />
                   }
-                ></StandaloneLink>
+                />
               </li>
             ))}
           </ul>


judithmh (approved):

app/services/upload/buildFileUploadError.ts
```diff
@@ -9,21 +12,18 @@ import { type UserData } from "~/domains/userData";
...
-      ),
-    },
-    validationResult.submittedData,
-  );
+): ValidationErrorResponseData {
+  return {
+    fieldErrors: Object.fromEntries(
+      Object.entries(validationResult.error?.fieldErrors ?? {})
+        .map(([key, val]) => [key.split(".")[0], val])
+        .filter(([key]) => key === inputName),
```
pgurusinga, line 25: Non-blocking one, but can we make this more readable? since Vite will optimise the code, we could do something like this:

```ts
const fieldErrors = validationResult.error?.fieldErrors ?? {};
const filteredFieldErrors = filterFieldErrorsByInputName(fieldErrors, inputName);

  return {
    fieldErrors: filteredFieldErrors,
    repopulateFields: validationResult.submittedData,
  };
  ```

File: app/services/flow/formular/fileUpload/handleFileUpload.server.ts [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+    [fieldName]: (userData[fieldName] as ArrayData).filter(
+      (_, index) => index !== inputIndex,
+    ),
+  };
+}
+
+const actionFile = (formAction: string) => {
+  if (formAction.startsWith("fileUpload")) return "upload";
+  if (formAction.startsWith("deleteFile")) return "delete";
+  return "unknown";
```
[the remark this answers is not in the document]
aaschlote: do you mean rename the function name or the current value `unknown`?

app/services/upload/buildFileUploadError.ts
aaschlote: I can do it, but what do you mean `Vite will optimise the code`?

pgurusinga: Vite will compile and minify the code

pgurusinga (approved):

PR #2260, opened 2025-06-27
Fix array item deletion no JS redirection

aaschlote:
### Problem

- User without JS are deleting an array item and redirect to the action url page

### Changes

- Add a hidden input to see if the user has JS enabled
- Redirect to the pathname in case the user has not JS enabled

### DIFF
--- a/app/routes/action.delete-array-item.ts
+++ b/app/routes/action.delete-array-item.ts
@@ -1,4 +1,4 @@
-import type { ActionFunctionArgs } from "react-router";
+import { redirect, type ActionFunctionArgs } from "react-router";
 import { logWarning } from "~/services/logging";
 import { validatedSession } from "~/services/security/csrf/validatedSession.server";
 import { getSessionManager } from "~/services/session.server";
@@ -21,7 +21,7 @@ export const action = async ({ request }: ActionFunctionArgs) => {
     return new Response(resultFormData.error.message, { status: 422 });
   }
 
-  const { arrayName, index, flowId } = resultFormData.value;
+  const { arrayName, index, flowId, pathname } = resultFormData.value;
 
   const { getSession, commitSession } = getSessionManager(flowId);
   const cookieHeader = request.headers.get("Cookie");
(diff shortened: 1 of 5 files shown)


app/services/session.server/arrayDeletion.ts
```diff
@@ -33,7 +34,7 @@ export function getArrayDataFromFormData(formData: FormData): Result<
 
   const { flowId } = parsePathname(pathname as string);
 
-  return Result.ok({ arrayName, index, flowId });
+  return Result.ok({ arrayName, index, flowId, pathname: pathname as string });
```
chohner, line 37: why do you need a cast here?

app/routes/action.delete-array-item.ts
```diff
@@ -21,7 +21,7 @@ export const action = async ({ request }: ActionFunctionArgs) => {
     return new Response(resultFormData.error.message, { status: 422 });
   }
 
-  const { arrayName, index, flowId } = resultFormData.value;
+  const { arrayName, index, flowId, pathname } = resultFormData.value;
```
chohner, line 24: is it possible to get the pathname from the request?

app/routes/action.delete-array-item.ts
```diff
@@ -33,5 +33,13 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   }
 
   const headers = { "Set-Cookie": await commitSession(flowSession) };
-  return new Response("success", { status: 200, headers });
+
+  const clientJavaScriptAvailable = formData.get("_jsEnabled") ?? "true";
```
chohner, line 37: defaulting to `true` here looks a bit strange.

maybe its simpler to check the opposite?

```typescript
if (formData.get("_jsEnabled") === "no"){
  return redirect(pathname, { headers })
}
return new Response("success", { status: 200, headers });
```

chohner (approved):

app/routes/action.delete-array-item.ts
aaschlote: I can't, the request url is from the action and not the page

PR #2266, opened 2025-06-30
bug(Kontopfandung): fix flow 

ingriddorioschulze:
This PR fixes the back flow for sozialleistungen-einmalzahlung and removes incorrect Boolean and redundant condition


chohner:
could we test this please? :)

chohner (approved):
nice, I tried and the test indeed fails without your fix :)

PR #2268, opened 2025-06-30
refactor(AutoSuggest): Prepare for Gerichtsfinder adoption

Spencer6497:
This PR makes some necessary refactorings to the AutoSuggestComponent, namely to decouple its logic before it gets adopted by the Gerichtsfinder.


app/components/inputs/autoSuggestInput/useDataListOptions.tsx
```diff
@@ -1,15 +1,37 @@
...
+        return response.ok ? await response.json() : [];
+      };
     case "airlines":
-      return `${API_PATH}/airlines/list`;
+      return async () => {
+        const response = await fetch(`${API_PATH}/airlines/list`);
+        return response.ok ? await response.json() : [];
+      };
+    case "streetNames":
+      return fetchStreetnamesForZipcode;
```
aaschlote, line 34: I see the function `fetchStreetnamesForZipcode` uses `lodash`. This library is super huge and brings problems for the bundle if it's in the client-side. Let's create a new router following the other APIs structure, in this case it would be `API_PATH/street/list` and pass the zip code as query parameter. In this new routing you can call all the logic of `fetchStreetnamesForZipcode` there.

app/components/inputs/autoSuggestInput/useDataListOptions.tsx
```diff
@@ -18,25 +40,31 @@ function getResourcePath(type: DataListType): string {
...
-          const json = await response.json();
-          setDataListOptions(json);
-        }
-      } catch {
+        const dataFetcher = getDataFetcher(dataListType, postleitzahl);
+        const options = await dataFetcher(postleitzahl);
+        setDataListOptions(options);
+      } catch (err: unknown) {
+        // eslint-disable-next-line no-console
+        console.error(err);
```
aaschlote, line 60: any special reason to have this `console.error` here?

app/components/inputs/autoSuggestInput/useDataListOptions.tsx
```diff
@@ -1,15 +1,37 @@
...
-      return `${API_PATH}/airports/list`;
+      return async () => {
+        const response = await fetch(`${API_PATH}/airports/list`);
+        return response.ok ? await response.json() : [];
+      };
     case "airlines":
-      return `${API_PATH}/airlines/list`;
+      return async () => {
+        const response = await fetch(`${API_PATH}/airlines/list`);
+        return response.ok ? await response.json() : [];
```
aaschlote, line 31: once you added the new routing, bring back the old logic of the `response.ok` inside the `hook` function, so we don't need to test this option for every data list option type.

aaschlote (commented):
you have a warning lint, please take a look.

app/components/inputs/autoSuggestInput/__test__/useDataListOptions.test.tsx
```diff
@@ -10,6 +10,8 @@ afterEach(() => {
   vi.restoreAllMocks(); // This clears all mocks after each test
 });
 
+vi.mock("react-router");
```
chohner, line 13: ?

app/components/inputs/autoSuggestInput/useDataListOptions.tsx
chohner: (however, I agree that aligning the API for fetching data would be preferable)

pgurusinga: ☝️  agree to what Alex said and just extra sauce for the route: path name instead of query param `API_PATH/street/list/{zipCode}`

app/components/inputs/autoSuggestInput/useDataListOptions.tsx
Spencer6497: felt kinda weird to send a likely super technical error message to the user, but didn't wanna swallow the error entirely

PR #2284, opened 2025-07-02
feat(PKH): add zusammenfassung page to the flow

ingriddorioschulze:
This PR adds zusammenfassung page to the PKH flow after /weitere-angaben and before /ende when feature flag is true


app/services/isFeatureFlagEnabled.server.ts
```diff
@@ -6,6 +6,7 @@ const localFeatureFlags = {
   showFileUpload: false,
   showNachbefragung: false,
   showKontopfaendungWegweiserFlow: false,
+  showZusammenfassung: false,
```
chohner, line 9: could we rename this flag to something more explicit? `showPKHZusammenfassung` or similar?

app/domains/prozesskostenhilfe/formular/index.ts
```diff
@@ -201,17 +204,34 @@ export const prozesskostenhilfeFormular = {
...
+                target: showPKHZusammenfassung ? "zusammenfassung" : "ende",
               },
             ],
           },
-          dokumente: { on: { BACK: "#weitere-angaben", SUBMIT: "ende" } },
+          ...(showPKHZusammenfassung && {
+            zusammenfassung: {
+              on: {
+                BACK: "#weitere-angaben",
+                SUBMIT: "dokumente",
```
Spencer6497, line 215: should we be checking for `fileUploadRelevant` and `showFileUpload` here too?

app/domains/prozesskostenhilfe/formular/index.ts
```diff
@@ -201,17 +204,34 @@ export const prozesskostenhilfeFormular = {
...
+          },
           ende: {
             on: {
               BACK: [
                 {
                   guard: ({ context }) =>
-                    Boolean(showFileUpload) && fileUploadRelevant({ context }),
+                    (Boolean(showFileUpload) &&
+                      fileUploadRelevant({ context })) ||
+                    Boolean(showPKHZusammenfassung),
```
Spencer6497, line 234: bit confused about the addition of the `showPKHZusammenfassung` guard here: it looks like this will route back to `dokumente` if we're showing the zusammenfassung, _regardless_ of whether file upload is enabled or relevant?

Spencer6497 (commented):
Questions about logical flow, but good work ✨

PR #2285, opened 2025-07-02
[Feat] Add Accordion to Infobox component

melgaafary:
Content managers want to add Accordion to PageContent. The easiest way design wise was to put it inside InfoBox and they can use that in pages

image


app/components/InfoBoxItem.tsx
```diff
@@ -31,7 +33,12 @@ const InfoBoxItem = ({
   inlineNotices,
   buttons,
   separator,
+  accordion,
 }: InfoBoxItemProps) => {
+  console.log("InfoBoxItem props:", {
```
chohner, line 38: this is probably still from debugging?

app/components/InfoBoxItem.tsx
```diff
@@ -70,6 +77,11 @@ const InfoBoxItem = ({
             ))}
           </ButtonContainer>
         )}
+        {accordion && (
+          <div className="max-w-[630px]">
```
chohner, line 81: I'm surprised we need the max-w here, shouldn't the parent take care of that?

app/services/cms/models/StrapiInfoBoxItem.ts
```diff
@@ -7,6 +7,8 @@ import { OptionalStrapiLinkIdentifierSchema } from "./HasStrapiLinkIdentifier";
 import { StrapiButtonSchema } from "./StrapiButton";
 import { StrapiDetailsSchema } from "./StrapiDetails";
 import { StrapiImageOptionalSchema } from "./StrapiImage";
+import { StrapiAccordionSchema } from "./StrapiAccordion";
```
chohner, line 10: nits: some import order fixes :)

chohner (approved):
lgtm except the lint issues

app/components/InfoBoxItem.tsx
melgaafary: ops yea good catch

app/components/InfoBoxItem.tsx
melgaafary: thats the parent

PR #2293, opened 2025-07-04
Preload fonts

aaschlote:
### Background
After preload the styles, the website https://speedvitals.com/ was showing that the we could also preload the fonts.

image

### Changes

- Preload the fonts to we can reduce the LCP
- Remove `font-family` duplication


app/root.tsx
```diff
@@ -62,6 +64,19 @@ export const links: LinksFunction = () => [
...
+    href: BundesSansWebBold,
+    as: "font",
+    crossOrigin: "anonymous",
+  },
+  {
+    rel: "preload",
+    href: "/fonts/Fira-Sans.woff2",
+    as: "font",
+    crossOrigin: "anonymous",
+  },
```
pgurusinga, line 79: If both CSS and fonts are prioritised, should CSS come first? I think HTML, CSS, and JS should take precedence, with fonts and images following. Does this imply that fonts are as important as CSS?

chohner: just checked it - it does indeed avoid a font-flash on first load.

Proposal: I will remove Fira-sans from preloading (since thats only used for kopfzeile) and merge to wrap up the performance work

PR #2297, opened 2025-07-07
bug(Report Problem): report problem button alignment on Amtsgerichtsfinder

ingriddorioschulze:
This PR fixes the positioning of the Report Problem button in the Amtsgerichtsfinder page

Before

Screenshot 2025-07-07 at 09 54 10

After

Screenshot 2025-07-07 at 09 53 37


app/routes/beratungshilfe.zustaendiges-gericht.auswahl.$PLZ.tsx
```diff
@@ -149,7 +149,9 @@ export default function Index() {
           </ButtonContainer>
         </ValidatedForm>
       </Container>
-      <ReportProblem />
+      <div className="self-end mr-32">
```
chohner, line 152: I think both on auswahl & suche the button should be inside the container (for aligned side padding with content) and `w-full`, similar to the flow pages.

right now the dialog doesnt show nicely

chohner: how does it look with the dialog open?

ingriddorioschulze: @melgaafary and I added some margin and it looks fine. I propose we go with this fix for now and then make it better when we do the grid refactor. WDYT @chohner?

melgaafary (approved):

PR #2323, opened 2025-07-11
Improve Strapi schemas components

aaschlote:
### Background
To render our list of components on the page, we need to set the key props for the React components (it's mandatory). We use the function `keyFromElement` to get the key, using the `__component` and `id`. The `id` can be undefined for the current schemas, however Strapi always returns an `id`, making the `id` for the schemas redundant.

### Changes

- For page components, use the schema `HasStrapiIdSchema` to force the component have an `id`,
- Remove __component from the `transform` and add to object schema of the components
- Add `id` as props for ListItems and InfoBoxItems
- Get rid of `id` for the `StrapiImageSchema`.

### Following changes
I will do the same changes for components inputs. I haven't done it yet, because it might have too many changes

### DIFF
--- a/app/components/InfoBoxItem.tsx
+++ b/app/components/InfoBoxItem.tsx
@@ -11,6 +11,7 @@ import Image, { type ImageProps } from "./Image";
 import RichText from "./RichText";
 
 export type InfoBoxItemProps = {
+  id: number; // Strapi id
   identifier?: string;
   label?: HeadingProps;
   headline?: HeadingProps;
--- a/app/components/InfoBox.tsx
+++ b/app/components/InfoBox.tsx
@@ -27,17 +27,7 @@ const InfoBox = ({
           data-testid="info-box-item-container"
         >
           {items.map((item) => (
-            <InfoBoxItem
-              separator={separator}
-              {...item}
-              key={
-                item.identifier ??
-                item.headline?.text ??
-                item.content ??
-                item.label?.text ??
-                ""
-              }
-            />
+            <InfoBoxItem separator={separator} {...item} key={item.id} />
           ))}
         </div>
       )}
--- a/app/components/__test__/InfoBox.test.tsx
+++ b/app/components/__test__/InfoBox.test.tsx
@@ -3,6 +3,7 @@ import InfoBox from "~/components/InfoBox";
 
 const mockInfoBoxItems = [
   {
+    id: 10,
     label: undefined,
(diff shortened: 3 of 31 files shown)


chohner:
nice idea with the `id` 👍🏼 

I stumbled upon the same thing with the mising `__component`. its only in the schema if the component is used in a freeform field (and not if they are just part of a page), so its safer to just always append it in the transform. I also tried `z.literal('x').default('x')` to avoid some useless transforms but that also doesn't work :/

aaschlote:
> nice idea with the `id`, although specifying them for testing is kinda annoying
> 
> I stumbled upon the same thing with the mising `__component`. its only in the schema if the component is used in a freeform field (and not if they are just part of a page), so its safer to just always append it in the transform. I also tried `z.literal('x').default('x')` to avoid some useless transforms but that also doesn't work :/

It seems for the nested component, Strapi decided not to added the `__component`, so we have live without them for nested component. More info [here](https://github.com/strapi/strapi/issues/8453#issuecomment-1511018764)

Spencer6497 (approved):

app/components/InfoBoxItem.tsx
```diff
@@ -11,6 +11,7 @@ import Image, { type ImageProps } from "./Image";
 import RichText from "./RichText";
 
 export type InfoBoxItemProps = {
+  id: number; // Strapi id
```
chohner, line 14: why do we pass the strapi id to out info box props? Using it for React's `key` doesn't require adding it to the element props

aaschlote: do you mean do something like this? 
image

aaschlote: I find this approach a bit odd for a couple of reasons:

- Relying on multiple optional fields to define a key makes the code harder to read and maintain.
- There's a non-zero risk of duplicate values among the optional keys, which could cause rendering issues. While unlikely, it's still a possibility we should consider.

Regarding the tests: our focus should be on testing behavior, not on how difficult something is to test. If we start designing components or logic primarily around testability rather than clarity and correctness, we're prioritizing the wrong concern. Clean, well-structured code is easier to test naturally—tests should adapt to the code, not dictate its structure.

Since Strapi provides a unique identifier, we can use that as the key. It keeps the code simpler and more reliable.

For these reasons, I’ll keep the changes as they are.

PR #2335, opened 2025-07-15
refactor(fileUpload): remove deprecated withZod

chohner:
This has grown bigger than expected (and there's still some tests missing) but I hope the re-arranged logic makes sense for everyone:
- removes the deprecated `withZod`, now directly returns validation errors from `uploadUserFile()`
- splits upload & deletion into separate functions again (removing the `processUserFile()` middle layer)
- only parses formData once in the action
- drops true-myth's `Result` wrapping


app/services/upload/fileUploadHelpers.server.ts
```diff
@@ -12,101 +11,70 @@ import {
...
-  await parseFormData(
-    request,
-    { maxFileSize: FIFTEEN_MB_IN_BYTES },
-    (fileUpload: FileUpload) => {
-      if (fileUpload.fieldName === fieldName && fileUpload.name) {
-        matchedFile = fileUpload;
-      }
-    },
-  );
-  return matchedFile;
```
Spencer6497, line 111: it's not coming to me at the moment but I think we actually need this matching logic for an edge case.... maybe attempting to submit an empty file in No-JS mode? or does this [get](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/2335/files#diff-6effc52a609a21d188d1a08b4a5f512ca0d3bc4653cae50f3c0c89f644a274e7R32) handle this

app/services/upload/fileUploadHelpers.server.ts
```diff
@@ -12,101 +11,70 @@ import {
...
-  userData: UserData,
   flowId: FlowId,
-): Promise<{
-  error?: ValidationErrorResponseData;
-  result?: { data: UserData };
-}> {
+): Promise<{ userData: UserData } | ValidationErrorResponseData> {
   const { fieldName, inputIndex } = splitFieldName(inputName);
-  const file = await parseFileFromFormData(request, inputName);
+  const arraySchema = getContext(flowId)[
```
Spencer6497, line 24: nit: this confused me for a second, could we rename it something like `fileUploadSchema` or `fileUploadArraySchema` instead?

Spencer6497 (commented):
I think these changes have regressed a bug in both JS- and No-JS mode, see screencaps below:

JS mode: uploading a file that fails validation overwrites an existing file visually
https://github.com/user-attachments/assets/00b46b65-270c-4a20-b3f3-09f22c210bd2

No-JS mode: 
https://github.com/user-attachments/assets/2b29eb96-de15-4e8e-8cc9-d95e1cee42dc

Spencer6497 (approved):

PR #2341, opened 2025-07-16
[Feat] Sentry performance metrics

melgaafary:
adding sentry instrumentation to enable Frontend and Backend Performance metrics in Sentry.

Still pending some discussions with the other teams because of their performance metrics resources usage


chohner (approved):
lgtm. maybe we could unify the settings (except the `tracesSampleRate`)? but probably not worth the effort...

PR #2342, opened 2025-07-16
feat(Report Problem): change label name and adjust spacing 

ingriddorioschulze:
This PR changes the label name from `Report Problem` to `Problem auf dieser Seite melden` and adds 80px margin top to the button.

Before

Screenshot 2025-07-16 at 13 09 29

Screenshot 2025-07-16 at 12 10 56

After

Screenshot 2025-07-16 at 13 02 09

Screenshot 2025-07-16 at 13 01 28

### DIFF
--- a/app/services/translations/translations.ts
+++ b/app/services/translations/translations.ts
@@ -87,7 +87,7 @@ export const translations = {
       de: "Ihr Feedback hilft uns, diese Seite für alle Nutzenden zu verbessern.",
     },
     "report-problem": {
-      de: "Problem melden",
+      de: "Problem auf dieser Seite melden",
     },
     cancel: {
       de: "Abbrechen",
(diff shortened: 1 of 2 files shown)


app/services/translations/translations.ts
```diff
@@ -87,7 +87,7 @@ export const translations = {
       de: "Ihr Feedback hilft uns, diese Seite für alle Nutzenden zu verbessern.",
     },
     "report-problem": {
-      de: "Problem melden",
+      de: "Problem auf dieser Seite melden",
```
pgurusinga, line 90: Since we’re using this feature, I’ve already spoken with our PM about this change. Would you mind waiting a bit?

ingriddorioschulze: Not a problem. I will let ours know as well!

chohner (approved):
lgtm

PR #2346, opened 2025-07-17
feat(Report Problem): adjust spacing 

ingriddorioschulze:
This PR adds 80px margin between cta and report problem button

Before

Screenshot 2025-07-17 at 15 44 16

After

Screenshot 2025-07-17 at 15 43 54


chohner (commented):
We tried to move positioning related things out of this component and into each page that uses it to avoid changing layouts across pages. do we want this margin everywhere, including for example the amtsgericht suche page?

PR #2356, opened 2025-07-21
feat(BerH): Calculate Financial Eligibility

Spencer6497:
This PR adds calculations for whether or not the user is eligible for Beratungshilfe based on a number of financial factors. When the user reaches the "Abgabe" site, it submits a custom PostHog event with the users eligibility.

- **groundwork for BerH financial eligibility**
- **add manual summation of eigentum**
- **finish up freibetrag + financial eligibility functions + tests**
- **add asyncFlowAction that submits eligibility**


app/services/flow/userFlowAction/postValidationFlowAction.ts
```diff
@@ -1,14 +1,23 @@
...
+    cookieHeader,
+  );
   const { stepId, currentFlow } = getPageAndFlowDataFromPathname(pathname);
   const { asyncFlowActions } = currentFlow as Flow;
   if (asyncFlowActions && stepId in asyncFlowActions)
-    await asyncFlowActions[stepId](request, userData);
+    await asyncFlowActions[stepId](request, {
+      ...userDataWithPageData,
+      ...newUserData,
+    });
```
Spencer6497, line 22: @chohner do you see any problem with injecting the entire userData here? I tried injecting it down the call stack inside of BerH's `asyncFlowActions` but vite's SSR build kept failing and I couldn't figure out why.

app/domains/beratungshilfe/formular/abgabe/isFinanciallyEligibleForBerH.ts
```diff
@@ -0,0 +1,110 @@
...
+import { calculateFreibetragBerHFormular } from "~/domains/beratungshilfe/vorabcheck/freibetrag";
+import { type GenericGuard } from "~/domains/guards.server";
+import { staatlicheLeistungenIsBuergergeld } from "~/domains/shared/formular/finanzielleAngaben/guards";
+import moneyToCents from "~/services/validation/money/moneyToCents";
+
+const eigentumWorthEligibilityThreshhold = 10000;
+
+export const isFinanciallyEligibleForBerH: GenericGuard<
+  BeratungshilfeFinanzielleAngabenUserData
+> = ({ context }) => {
```
chohner, line 11: could this just be `userData: BeratungshilfeFinanzielleAngabenUserData`?

app/services/flow/userFlowAction/postValidationFlowAction.ts
chohner: lets pass the updated userData into the function, instead of fetching here.

We should then just update the existing `sendCustomAnalyticsEvent` to keep sending the same data as before

chohner (approved):
overall looks good, small comments

PR #2358, opened 2025-07-21
[RAST-24] Upgrade zod4

chohner:
- upgrades to zod v4
- makes `MultiFieldValidations` generic for better type safety
- upgrades all `.superRefine()` to `.check()`
- migrates from `error?.errors` to `error?.issues`
- migrations from `.Enum` / `.Values` to `.enum`
- migrates from `.errorMap` to `{error: "" }`

TODO:
- [x] `schemaOrEmptyString`
- [x] `tests/integration/util.ts`
- [x] deprecations:
	- [x] `.merge()`
	- [x] `ZodTypeAny`
	- [x] `error?.format()`


app/services/cms/components/StrapiTextarea.ts
```diff
@@ -12,13 +12,13 @@ export const StrapiTextareaComponentSchema = z
     __component: z.literal("form-elements.textarea"),
     name: z.string(),
     description: StrapiRichTextOptionalSchema(),
-    details: StrapiDetailsSchema.nullable().transform(omitNull),
+    details: StrapiDetailsSchema.nullable().transform(omitNull).optional(),
```
aaschlote, line 15: why this is optional?

app/services/validation/money/buildMoneyValidationSchema.ts
```diff
@@ -17,7 +17,7 @@ export const buildMoneyValidationSchema = (
     .trim()
     .min(1, "required")
     .transform((v) => preprocessMoney(v))
-    .refine((v) => validateMoney(v), { message: "wrong_format" })
+    .refine((v) => validateMoney(v), { error: "wrong_format", abort: true })
```
aaschlote, line 20: do we need to set the `abort` as true here?

aaschlote (approved):
overall everything sounds good, however I guess next time we can use the option to install multiple dependency versions, so we can make small changes and merging to main. Check out [here](https://stackoverflow.com/questions/70270044/how-to-manage-multiple-versions-of-npm-modules-in-a-project-for-different-packag#answers-header).

LGTM!

app/services/validation/money/buildMoneyValidationSchema.ts
chohner: iirc we want to abort early because later transforms don't make sense if this one fails

app/services/cms/components/StrapiTextarea.ts
chohner: Its the old topic of `field?: schema` vs `field: schema | undefined`, that zod seems to have gotten strict about again. I reverted some but left the ones so I don't have to touch out test factories

app/domains/types.ts
```diff
@@ -1,14 +1,17 @@
 import type { z } from "zod";
+import type { AllowedUserTypes } from "./userData";
```
pgurusinga, line 2: sonarcloud reports a cyclic dep here

PR #2360, opened 2025-07-21
feat(cookieBanner): make desktop similar to mobile (scrollable)

pgurusinga:
cookie


chohner (approved):
lgtm but I'm confused - why do we want sticky on desktop and not on mobile? I thought we wanted the same behavior?

edit: reading the ticket, it should not be sticky but scroll with the content, same as on mobile?

PR #2361, opened 2025-07-21
feat(MultiFieldsValidationBaseSchema): make generic

chohner:
multi-field validations can now specify their needed fields and are typesafe internally


pgurusinga (approved):

app/domains/fluggastrechte/formular/flugdaten/userData.ts
```diff
@@ -52,9 +52,9 @@ export const fluggastrechteFlugdatenInputSchema = {
     customRequiredErrorMessage,
   ),
   anschlussFlugVerpasst: YesNoAnswer,
-  ersterZwischenstopp: airportSchema,
-  zweiterZwischenstopp: airportSchema,
-  dritterZwischenstopp: airportSchema,
+  ersterZwischenstopp: airportSchema.optional(),
```
aaschlote, line 55: why is optional? the fields should be required

app/domains/fluggastrechte/formular/services/validation/__test__/validateAnotherFlightPage.test.ts
```diff
@@ -1,14 +1,18 @@
+import pick from "lodash/pick";
```
aaschlote, line 1: I guess if this change, we are exposing the `lodash` to the front-end. Can we create our own `pick` function and move to use the `lodash`?

image

chohner: Imports inside the tests really shouldn't end up in the front-end bundle. Could you check the production bundle?

Also, deep importing (import from `import pick from "lodash/pick"` instead of `import {pick} from 'lodash'`) are tree-shakable, as far as I know

aaschlote: ohh, true, I haven't noticed this file it's a test one, my bad!

app/domains/fluggastrechte/formular/flugdaten/userData.ts
aaschlote: yeah, this is true, however when the user is on the page with this field, it must enter an information, for this reason it shouldn't be `optional`. As far I haven seen in the project, we only make the element of the userData `optional` when it's not required to enter the data on the page, like `phoneNumber`, `iban` and etc. By default, the whole `userData` is already partial.

image

chohner: True, we have (at least) two different types of "optionality":
- no input required (like `iban`)
- not always relevant (because in sub-flow etc, like `ersterZwischenstopp`)

I marked the zwischenstopp schemas as optional to mark them as potentially `undefined` the inside `validateStopoverDuplicates`. maybe instead it should just mark all inputs as `partial()` 🤔

PR #2373, opened 2025-07-24
RAST-173 fix zahlungArbeitgeber schema

ingriddorioschulze:
This PR fixes the schema for the zahlungArbeitgeber which is a select schema and it was added as YesNo schema


chohner (approved):
lgtm

PR #2392, opened 2025-07-29
Fix requests strapi pLevel

aaschlote:
### Background

Since we implemented the FieldSet component, we were required to increase the pLevel to request the data from Strapi from 5 to 6. This change caused performance issues, taking more than 4 seconds to load the page.

### Changes

- Remove unneeded `fetechErrors` api (The errors are stored in the code)
- Create parameter pLevel in the function `getStrapiEntryFromApi` to set the pLevel parameter to Strapi
- Set common requests data as translations, cookie banner and page header as pLevel 2
- Set request data for footer as pLevel 3
- Set request data for pages as pLevel 5 (default value)
- Set request data for form flow pages as pLevel 6

Before
image

After
image


app/services/cms/__test__/index.test.ts
```diff
@@ -30,27 +28,6 @@ describe("services/cms", () => {
     });
   });
 
-  describe("fetchEntries", () => {
```
chohner, line 33: not for this PR but maybe we could fully remove `fetchEntries()`? its a very thin abstraction used only by `fetchMultipleTranslations()`

chohner (approved):
nice work, lgtm

pgurusinga (approved):

PR #2395, opened 2025-07-29
Create the abgabe page as downloadable pdf

aaschlote:
### Changes

- Add `print:hidden` to hidden the components `FlowNavigation`, `UserFeedback` and part of the footer when prints the page
- Add to the top of the page the message `Diese Seite wurde heruntergeladen am {currentDate} um {currentTime} Uhr` when prints the page
- Modify the title of the page when prints the page on `/fluggastrechte/formular` 
- Move function `sanitizeHtml` to own file
- Fix: Add the default renderer markdown when set a custom render for the function `parseAndSanitizeMarkdown`

Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-275


app/util/date.ts
```diff
@@ -30,6 +36,14 @@ export const toGermanDateFormat = (date: Date) => {
   });
 };
 
+export function toHourAndMinuteTime(date: Date) {
+  const hours = date.getHours();
+  const minutes = date.getMinutes();
+  return `${hours.toString().padStart(2, "0")}:${minutes
+    .toString()
+    .padStart(2, "0")}`;
```
pgurusinga, line 44: can we use international date API? e.g.

```ts
const hhmmFormatter = new Intl.DateTimeFormat(undefined, {
  hour:   "2-digit",
  minute: "2-digit",
});
function toHourAndMinuteTime(date: Date): string {
  return hhmmFormatter.format(date);
}

console.log(toHourAndMinuteTime(new Date()))
```
this will produce e.g. `17:00`

app/root.tsx
```diff
@@ -180,7 +183,7 @@ function App() {
   return (
     <html lang="de">
       <head>
-        <title>{title}</title>
+        <title>{getTitlePage(title, pathname, shouldPrint)}</title>
```
chohner, line 186: I personally would prefer to keep the context of pathnames & booleans in the caller context instead of pushing into the sub-function. ie `shouldPrint ? title : generatePrintTitle(pathname) ?? title`.

app/styles.css
```diff
@@ -302,3 +302,9 @@ details:not(:has(.summary-content:focus-visible)) {
 h1:focus {
   outline: none;
 }
+
+@media print {
```
chohner, line 306: I like it but the obvious question - was this aligned across streams? would be nice to have a coherent print style..

chohner: also, I'm unsure what the pro's and con's are of using the `useShouldPrint()` hook vs css for styling 🤔

chohner (commented):
lots of comments but no blocker

app/styles.css
```diff
@@ -302,3 +302,9 @@ details:not(:has(.summary-content:focus-visible)) {
...
+  .hidden-print {
+    display: none;
+  }
+}
```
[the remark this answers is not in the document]
aaschlote: ohh nice, I haven't seen it before. Let's use it!

app/root.tsx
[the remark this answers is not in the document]
aaschlote: I have tested, it's fine. Thanks for pointing out!

image

app/util/date.ts
aaschlote: it sounds better, changing it!

app/styles.css
aaschlote: I can mention to Kristin to align with you folks, however I see removing a few parts makes the process of printing the page much better, for example removing the side navigation, user feedback and etc.

About the hook and css styling, I would prefer using the `css` style, because it's bring less logic to the source code and it works perfect.

Btw: I'm using the https://tailwindcss.com/docs/hover-focus-and-other-states#print recommended by @pgurusinga

pgurusinga (approved):

tests/e2e/critical/homepage.spec.ts
```diff
@@ -60,4 +60,13 @@ test.describe("homepage", () => {
 
     footerLinks.forEach((url) => expect(footerHrefs).toContain(url));
   });
+
+  test("Printing the page hidden footer navigation", async ({ page }) => {
```
pgurusinga, line 64:
```suggestion
  test("should hide the footer navigation when printing the page", async ({ page }) => {
```

PR #2402, opened 2025-07-30
Adjust cookie dialog role label

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-406

### DIFF
--- a/app/components/cookieBanner/CookieBanner.tsx
+++ b/app/components/cookieBanner/CookieBanner.tsx
@@ -3,16 +3,20 @@ import { useFetcher } from "react-router";
 import Button from "~/components/Button";
 import Container from "~/components/Container";
 import Heading, { type HeadingProps } from "~/components/Heading";
-import RichText, { type RichTextProps } from "~/components/RichText";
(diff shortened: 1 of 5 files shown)


app/components/cookieBanner/CookieBanner.tsx
```diff
@@ -47,10 +51,16 @@ export function CookieBanner({
...
 
+  const paragraphAriaDescribedBy = arrayIsNonEmpty(content.paragraphs)
+    ? `paragraph-${content.paragraphs[0].id}`
+    : undefined;
+
   return (
-    <section
-      className="right-16 left-16 border-b-2 border-blue-800 z-50 bg-blue-300"
-      aria-label="Cookie banner"
+    <dialog
```
chohner, line 59: Also commented on the jira ticket but as far as I understand, the cookie banner's new styling doesn't make it a `dialog` anymore (it's not placed on top of content)

aaschlote: this is a good point, however Marion is on holidays. I'll keep the PR and hold the ticket.

PR #2416, opened 2025-08-04
Improve feedback component for sr users

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-542


app/components/userFeedback/FeedbackFormBox.tsx
```diff
@@ -39,6 +39,8 @@ const getFeedbackFromUrl = (url: string) => {
   return new URLSearchParams(url).get("wasHelpful");
 };
 
+const HEADING_PERSONAL_DATA_FEEDBACK_ID = "heading-personal-data-feedback-id";
```
chohner, line 42: I think the React pattern would be to use `const headingId = useId()`?

app/components/userFeedback/FeedbackSuccessMessage.tsx
```diff
@@ -1,21 +1,20 @@
...
   subtitle: string;
   innerRef?: React.Ref<HTMLParagraphElement>;
 }) => {
   return (
-    <div className="flex max-sm:flex-col sm:items-center text-base mb-[1em]">
+    <output className="flex max-sm:flex-col sm:items-center text-base mb-[1em]">
       <CheckCircleIcon className="w-[2em] h-[1.5em] mr-[0.25em] text-green-600 " />
       <p className="font-bold mr-4" ref={innerRef} tabIndex={-1}>
-        {title}
+        {translations.feedback["success-message"].de}
```
chohner, line 15: I personally think its nice to keep the nested components with props instead of hard-wiring translations into them.

But either way, lets use the same mechanism for both strings here

app/components/userFeedback/__test__/FeedbackSuccessMessage.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+    expect(title).toBeInTheDocument();
+    expect(getByText(subtitleText)).toBeInTheDocument();
+  });
+
+  it("should render an icon", () => {
+    const { getByTestId } = render(<FeedbackSuccessMessage subtitle={""} />);
+    const icon = getByTestId("CheckCircleIcon");
+    expect(icon).toBeInTheDocument();
+    expect(icon).toHaveClass("text-green-600");
+  });
```
chohner, line 23: isn't this icon static?

chohner (approved):
lgtm, just a few comments

app/components/userFeedback/FeedbackFormBox.tsx
aaschlote: I didn't know about this hook, I'll use it!

app/components/userFeedback/__test__/FeedbackSuccessMessage.test.tsx
aaschlote: yes, actually this test exist already, I just added it into a new test

PR #2423, opened 2025-08-05
Update typescript 5.9.2

aaschlote:
Update typescript to the version 5.9.2.

[Release changes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/)


app/routes/shared/pdfDownloadLoader.ts
```diff
@@ -83,7 +83,7 @@ export async function loader({ request }: LoaderFunctionArgs) {
   const fileSize = fileContent.length;
   const filename = `${name}_${pdfDateFormat(today())}.pdf`;
 
-  return new Response(fileContent, {
+  return new Response(Buffer.from(fileContent), {
```
pgurusinga, line 86: 🤔 What error needs changing?

aaschlote: I was getting this:

```
app/routes/shared/pdfDownloadLoader.ts:86:23 - error TS2345: Argument of type 'Buffer<ArrayBufferLike> | Uint8Array<ArrayBufferLike>' is not assignable to parameter of type 'BodyInit | null | undefined'.
  Type 'Buffer<ArrayBufferLike>' is not assignable to type 'BodyInit | null | undefined'.
    Type 'Buffer<ArrayBufferLike>' is missing the following properties from type 'URLSearchParams': size, append, delete, get, and 2 more.

86   return new Response(fileContent, {
```

PR #2427, opened 2025-08-06
Move arrays translations to component array sumarry

aaschlote:
### Changes

- Use content of component `page.array-summary` to render `title`, `button label`, `description` (under the title) and and items
- Remove `arrayDeleteButtonLabel`, `arrayEditButtonLabel`, `arrayAddButtonLabel` from Strapi translations and bring to the source code translations
- Remove `getArraySummaryPageTranslations` as they have been moved to Strapi as content
- Add parameter in the function `applyStringReplacement` called `skipIndexArrayReplacement`, to skip the array replacement in the back-end. We need to the replacement in the front-end. It's current doing it, but the content it was in the translations where it has been not interpolated.

Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-358


app/components/arraySummary/ArraySummary.tsx
```diff
@@ -16,58 +13,42 @@ type ArraySummaryProps = {
     data: ArrayData;
     configuration: ArrayConfigClient;
   };
-  readonly translations?: Translations;
+  readonly content: {
```
chohner, line 16: maybe even easier to have them as top-level props?

app/components/arraySummary/ArraySummary.tsx
```diff
@@ -16,58 +13,42 @@ type ArraySummaryProps = {
...
+    readonly title?: HeadingProps;
+    readonly description?: string;
+    readonly buttonLabel: string;
+    readonly subtitle?: HeadingProps;
```
chohner, line 20: maybe simpler: you can wrap the whole type in Readonly<> instead of repeating the readonly attribute

app/components/arraySummary/ArraySummary.tsx
```diff
@@ -16,58 +13,42 @@ type ArraySummaryProps = {
...
+    readonly items: Array<{ item: string; value: string }>;
```
chohner, line 21: I'm not sure what `items` means here. maybe `itemLabels`? (assuming this is to translate for example enum values). if so, an object might be a better data format

app/components/arraySummary/ArraySummaryDataItems.tsx
```diff
@@ -1,22 +1,35 @@
...
+const getItemValue = (
+  itemsContent: Array<{ item: string; value: string }>,
+  itemKey: string,
+  itemValue: BasicTypes,
+) => {
+  return (
+    itemsContent.find((item) => item.item === `${itemKey}.${itemValue}`)
+      ?.value ?? itemValue
+  );
 };
```
chohner, line 33: regarding `items`: I think these functions sound that an array isn't the right data format, but instead something like `const itemLabels = {key: string}`

chohner (approved):
much nicer interface! left a few comments

ingriddorioschulze (approved):

app/components/arraySummary/ArraySummaryDataItems.tsx
aaschlote: I can check if I can transform it from `array` to an `object`

app/services/cms/models/StrapiArraySummary.ts
```diff
@@ -1,9 +1,36 @@
...
+    ...HasStrapiIdSchema.shape,
+  })
+  .transform(({ itemLabels, ...cmsData }) => {
+    const items = itemLabels.reduce(
+      (acc, { item, value }) => {
+        acc[item] = value;
+        return acc;
+      },
+      {} as Record<string, string>,
+    );
```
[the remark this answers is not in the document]
pgurusinga: +1 on map bitte 🙏

pgurusinga (approved):

PR #2428, opened 2025-08-06
Bugfix: Use lodash.escapeRegExp for safer regex escaping

pgurusinga:
## Problem
We only escape parentheses, which break whenever someone types `?`, `+`, `.` or any other regex special character. 

## Proposed changes
- Added `lodash.escapeRegExp` to escape all regex-meta chars.
- Updated splitHighlightWord to use it instead of our old helper.
- Expanded test suite with other special characters


Spencer6497 (approved):
easy 👾

aaschlote (approved):
LGTM!

PR #2435, opened 2025-08-07
RAST-194 move persoenlichen daten to pageSchemas

ingriddorioschulze:
This PR migrates BerH persoenlichen daten to pageSchemas

### DIFF
--- a/app/domains/shared/formular/persoenlicheDaten/pages.ts
+++ b/app/domains/shared/formular/persoenlicheDaten/pages.ts
@@ -1,0 +1,46 @@
+import { type PagesConfig } from "~/domains/pageSchemas";
+import { createDateSchema } from "~/services/validation/date";
+import { germanHouseNumberSchema } from "~/services/validation/germanHouseNumber";
+import { phoneNumberSchema } from "~/services/validation/phoneNumber";
+import { postcodeSchema } from "~/services/validation/postcode";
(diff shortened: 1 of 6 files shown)


app/domains/shared/formular/persoenlicheDaten/pages.ts
```diff
@@ -0,0 +1,46 @@
...
+    pageSchema: {
+      street: stringRequiredSchema,
+      houseNumber: germanHouseNumberSchema,
+      ort: stringRequiredSchema,
+    },
+  },
+  telefonnummer: {
+    stepId: "persoenliche-daten/telefonnummer",
+    pageSchema: {
+      telefonnummer: phoneNumberSchema.optional(),
```
ingriddorioschulze, line 43: @chohner Maybe you know more about that...
Do you think it is ok if I use the `phoneNumberSchema.optional()` instead of `schemaOrEmptyString(phoneNumberSchema)`?
Using the schemaOrEmptyString does not show the input. Spencer thinks it is related to ZodUnion.

chohner: If nothing is entered in the field, the form submits an empty string `''`, not `undefined`. So for a field be optional for the user, we need the schema to validate successfully against `''`, which is different from `z.optional()`.

for handling the union we indeed might need to think about how to unpack it

PR #2451, opened 2025-08-13
[Rast-176] Antragstellende  pageSchema refactor

melgaafary:
This covers all Antragstellende Person pages and subpages/arrays. 

pageSchemas now handles flat arrays and also nested arrays in prep for finanzielleAngaben which will come next


app/domains/pageSchemas.ts
```diff
@@ -17,11 +17,91 @@ const pages: Partial<Record<FlowId, PagesConfig>> = {
...
+
+  // For non-array pages, return the page schema as is
+  return pageConfig.pageSchema;
+}
+
+function findArrayPageSchemaInPage(
+  pageConfig: PageConfig,
+  stepPathParts: string[],
+): SchemaObject | undefined {
+  if (!pageConfig.arrayPages || stepPathParts.length === 0) return undefined;
```
Spencer6497, line 77: isn't `pageConfig` guaranteed to have `arrayPages` at this point, since we checked it earlier?

app/domains/pageSchemas.ts
```diff
@@ -17,11 +17,91 @@ const pages: Partial<Record<FlowId, PagesConfig>> = {
...
+      // Check if this page's stepId matches the beginning of our path
+      const pageStepId = page.stepId;
+      const pathWithoutLeadingSlash = stepPathParts.join("/");
+
+      // The page should be the one that contains the array we're looking for
+      // We need to check if the path starts with this page's stepId
+      return pathWithoutLeadingSlash.startsWith(pageStepId);
+    });
+
+    if (currentPageConfig?.arrayPages) {
```
Spencer6497, line 55: i think this check is redundant, because the `Object.values().find()` already filtered pages without `arrayPages`?

app/domains/__test__/pageSchemas.test.ts
```diff
@@ -12,6 +12,99 @@ describe("getPageSchema", () => {
   it("should return undefined if no page schema is found", () => {
     expect(getPageSchema("/step-does-not-exist")).toBeUndefined();
   });
+
+  describe("array pages", () => {
```
Spencer6497, line 16: maybe i'm confused, but is there actually a test where we're testing nested arrays? I only see a test for _invalid_ nested arrays but not _valid_ ones

app/domains/pageSchemas.ts
```diff
@@ -17,11 +17,91 @@ const pages: Partial<Record<FlowId, PagesConfig>> = {
...
+      if (!page.arrayPages) return false;
+
+      // Check if this page's stepId matches the beginning of our path
+      const pageStepId = page.stepId;
+      const pathWithoutLeadingSlash = stepPathParts.join("/");
+
+      // The page should be the one that contains the array we're looking for
+      // We need to check if the path starts with this page's stepId
+      return pathWithoutLeadingSlash.startsWith(pageStepId);
+    });
```
Spencer6497, line 53: don't we already have `pageConfig` defined for the current page, in [this line](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/692d3647e1f2b4bf89a2e43bc86859a0b77e2fe6/app/domains/pageSchemas.ts#L25)?

app/domains/pageSchemas.ts
```diff
@@ -17,11 +17,91 @@ const pages: Partial<Record<FlowId, PagesConfig>> = {
...
+
+    if (currentPageConfig?.arrayPages) {
+      // Try to find the array page schema by navigating through the path
+      const arraySchema = findArrayPageSchemaInPage(
+        currentPageConfig,
+        stepPathParts,
+      );
+      if (arraySchema) {
```
Spencer6497, line 61: it looks like `findArrayPageSchemaInPage` can return undefined, shouldn't we return `arraySchema` regardless of whether it's defined or not? or is that what the ultimate fallback to `return pageConfig.pageSchema` is doing

melgaafary: yup its the ultimate fallback, not all pages have arraySchema

app/domains/pageSchemas.ts
melgaafary: true, just having it as a failsafe just in case someone made an array page without putting in `arrayPages`

chohner (approved):

app/domains/pageSchemas.ts
```diff
@@ -17,11 +17,91 @@ const pages: Partial<Record<FlowId, PagesConfig>> = {
...
+    }
+  }
+
+  if (!pageConfig?.pageSchema) return undefined;
+
+  // For non-array pages, return the page schema as is
+  return pageConfig.pageSchema;
+}
+
+function findArrayPageSchemaInPage(
```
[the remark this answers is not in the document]
melgaafary: yea makes sense

PR #2453, opened 2025-08-13
Restructure component — part 2: Content

pgurusinga:


chohner:
I understand we talked about it but looking at the change, I feel like `content` isn't a strong enough group to collect our basic components in (compare to `form`, which are quite specialized). not a blocker, just a thought.

In fact, I feel like the larger layout components (pagehheader, breadcrumbs, footer, hero, ...) might be a better sub-group?

```
/components
	/Heading.tsx
	/Paragraph.tsx
	/...
	/form
		/Input.tsx
		/...
	/layout
		/Breadcrumbs.tsx
		/Hero.tsx
		/...
```

what do you think?

chohner:
I like it! I think ideally we should strive to minimize CMS coupling in our front-end components but making it visible like this is a great idea

PR #2458, opened 2025-08-14
feat(components): remove alignToMainContainer

chohner:
on flow pages:
	- moves sidebar navigation into the main content
	- removes option to left-align header, breadcrumbs, kopfzeile

image

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -123,14 +123,11 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
     pathname,
     trackingConsent,
   );
-  const flowIdMaybe = flowIdFromPathname(pathname);
+  const isAnyFlowPage = Boolean(flowIdFromPathname(pathname));
   return data(
     {
       breadcrumbs,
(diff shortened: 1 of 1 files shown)


app/components/Breadcrumbs.tsx
```diff
@@ -12,27 +11,24 @@ type BreadcrumbsProps = {
...
-  const navClassNames = classNames("py-8 px-16 flex flex-wrap items-center", {
-    [`${alignToContainer} !py-8`]: alignToMainContainer,
-  });
-
   return (
     validBreadcrumbs.length > 0 && (
       <div className="bg-blue-100">
-        <nav className={navClassNames} aria-label={ariaLabel}>
+        <nav
+          className={`!py-8 px-16 flex flex-wrap items-center ${alignToContainer}`}
```
aaschlote, line 29: do we need this `!important` (!py-8)

chohner: yes, because `container` brings its own vertical padding (which imo we should change)

PR #2470, opened 2025-08-18
feat(getNestedSchema): unpack unions to first schema

chohner:


app/components/formElements/schemaToForm/getNestedSchema.ts
```diff
@@ -1,9 +1,17 @@
...
-  schema: T | z.ZodNullable<T> | z.ZodOptional<T>,
+  schema:
+    | T
+    | z.ZodNullable<T>
+    | z.ZodOptional<T>
+    | z.ZodUnion<readonly [T, ...z.ZodType[]]>,
 ): T => {
   if ("in" in schema) return getNestedSchema(schema.in as T);
   if ("unwrap" in schema) return getNestedSchema(schema.unwrap());
+  if (isZodUnion(schema)) return getNestedSchema(schema.options[0]); // Union: pick first schema
```
Spencer6497, line 15: is the first union option guaranteed to be the one we're looking for? I think I ran into this with `persoenliche-daten.telefonnummer` where the first option was the string literal, and the second was the optional string

Spencer6497 (approved):

PR #2471, opened 2025-08-19
feat(zod): custom enum required message

chohner:
Proposal: Replace `customRequiredErrorMessage` with custom error in `z.config()`

Slightly more config magic, but less chance of forgetting und much less configuration on each enum


Spencer6497 (approved):
👍

app/services/validation/__test__/configureZod.test.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "zod/mini";
+import { configureZod, restoreZodDefaults } from "../configureZod";
+
+describe("configureZod", () => {
+  afterEach(() => {
+    restoreZodDefaults();
+  });
+
+  it("sets customErrors", () => {
```
aaschlote, line 9: this tests is not clear what it does, I would recommend to make two tests:

1. Test when it does not have the configuration, so you can implement the function `restoreZodDefaults` in this test file
2. Test when it has the configuration of zod, in this case `configureZod`

Most of the tests should check one behaviour and not multiple. Summarise tests is not a good practise.

chohner: Imo this is not two tests - this is verifying that setting a public config changes behaviour by comparing before and after. adding two tests does not verify this behaviour

app/services/validation/configureZod.ts
```diff
@@ -0,0 +1,13 @@
...
+  z.config({
+    jitless: true, // Should disable CSP warning, see https://github.com/colinhacks/zod/issues/4461
+    customError: (iss) => {
+      // Empty strings on enums trigger an 'invalid_value' error. However, we want a generic error message, which can be replace with a user-friendly message.
+      if (iss.code === "invalid_value" && iss.input == "")
+        return { message: "required" };
+    },
+  });
+
+export const restoreZodDefaults = () => z.config({ customError: undefined });
```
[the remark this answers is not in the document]
chohner: I'm torn on this, since this is encapsulating the interaction with `z.config()` and the tests shouldn't need to now the implementation details to do the cleanup. It is true however that we generally don't need to do this cleanup in the app

app/components/formElements/__test__/ValidatedFlowForm.test.tsx
```diff
@@ -41,6 +42,14 @@ const fieldNameValidatorSpy = vi.spyOn(
 );
 
 describe("ValidatedFlowForm", () => {
+  beforeAll(() => {
+    configureZod(); // Need to auto-configure enum errors
+  });
+
+  afterAll(() => {
+    restoreZodDefaults();
```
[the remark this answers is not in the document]
chohner: To undo the global config thats happening in `beforeAll()`? I would argue thats pretty self-explanatory but I could add a comment

app/services/validation/__test__/configureZod.test.ts
aaschlote: It looks like we’re dealing with two different behaviours: one without `configureZod` and another with it. From the test description and the way it’s written, I don’t see where the comparison you mentioned in your previous comment is actually being made. Am I missing something?

app/components/formElements/__test__/ValidatedFlowForm.test.tsx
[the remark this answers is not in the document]
aaschlote: I was mentioned the `restoreZodDefaults`.

As we are using vitest config to run as `threads`, every test runs in a separate pool, so either using `restoring zod config` or not it won't break the others tests, making the `afterAll` redundant. 

Source: https://vitest.dev/guide/improving-performance

app/services/validation/__test__/configureZod.test.ts
aaschlote: "Good question.

‘How do you know that the change in behavior comes from configureZod()?’
→ Because the current test explicitly calls `configureZod`, similar as your current test is doing.

‘And the second test simply tests the library default errors.’
→ That’s true, but note that your test calls `restoreZodDefaults()` in the afterEach. That’s not fundamentally different from my approach—the main difference is that I’m calling it within a specific test unit (it) instead of initializing the test without restoring the Zod config."

aaschlote (approved):
LGTM!

PR #2478, opened 2025-08-20
refactor(start.sh): Remove startup env variable script

Spencer6497:
This PR removes the clunky start.sh script in favor of a much more declarative approach reading secrets from the filesystem first, and, if not available, from the node process

- **migrate file reads away from start.sh**
- **fully remove start.sh**


aaschlote (approved):
Nice implementation! I’ve left some comments to help improve the code, as well as one pointing out a missing line related to the node environment.

Since changes are deployed directly to production, we can’t be completely sure they won’t cause issues. I’d recommend testing the application by running it locally on Kubernetes with these changes. I know this can be a bit challenging, but it would be great to try. If it’s not possible, no worries.

Note: If you are able to run the application locally with Kubernetes, please consider writing a short document explaining how to set it up.

start.sh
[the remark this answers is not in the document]
Spencer6497: afaict, this line was actually redundant to begin with, as we're calling `NODE_ENV=production node ./server.js` directly in `package.json`

app/services/env/env.server.ts
```diff
@@ -27,16 +29,32 @@ type Config = {
...
+      "",
     CMS: process.env.CMS ?? "FILE",
+    GERICHTSFINDER_ENCRYPTION_KEY:
+      readSecretVolume("/etc/courtdata-secrets/password") ??
+      process.env.GERICHTSFINDER_ENCRYPTION_KEY ??
+      "",
     REDIS_ENDPOINT: process.env.REDIS_ENDPOINT ?? "localhost:6380",
-    REDIS_PASSWORD: process.env.REDIS_PASSWORD ?? "",
-    COOKIE_SESSION_SECRET: process.env.COOKIE_SESSION_SECRET ?? "s3cr3t",
+    REDIS_PASSWORD:
```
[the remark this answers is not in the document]
Spencer6497: I'm not quite sure what you mean?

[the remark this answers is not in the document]
aaschlote: For all configuration variables that need to resolve a value (from a secret, environment variable, or a default), we could create an internal helper function. For example:

```js
const getRedisPassword = () => {
return readSecretVolume("/etc/redis-password-secret/password") ??
      process.env.REDIS_PASSWORD ??
      "",
}
```

The idea is to simplify the config() function, which is already around 70 lines long and could grow even more as we add additional configuration variables for new features or platform settings.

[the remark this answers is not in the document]
Spencer6497: are we sure this is the best pattern? I see what you mean about reducing the complexity of `config()` itself, but we then add +30-40 more lines to the file with identically-seeming getter functions, which I might imagine could be confusing in cases like 
`getS3DataStorageAccessKey`
and 
`getS3DataStorageSecretKey`


https://github.com/user-attachments/assets/5e3cf1d8-5319-4458-8035-e1e634df819f

app/services/env/__test__/environmentVariableHelpers.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    "S3_DATA_STORAGE_SECRET_KEY",
+    "/etc/s3-storage-credentials-secret-key/password",
+    "FAKE_S3_DATA_STORAGE_SECRET_KEY",
+  ],
+];
+
+describe("environment variables", () => {
+  describe.each(envMap)("%s", (key, path, value) => {
+    beforeEach(() => {
+      vi.mocked(fs.readFileSync).mockImplementation(mockedReadFileSync);
```
aaschlote, line 45: you don't need these two lines in every beforeEach, just in the test `can be read from a local file`, so I would recommend to change:

```js
beforeEach(() => {
      mockedReadFileSync.mockReturnValue(undefined);
    });
    it("can be read from a local file", () => {
      mockedReadFileSync.mockReturnValue(value);
      expect(config()[key]).toBe(value);
      expect(mockedReadFileSync).toHaveBeenCalledWith(path, "utf8");
    });
```

You can remove the `mockedReadFileSync.mockReturnValue(undefined);` from the other two tests and move `      vi.mocked(fs.readFileSync).mockImplementation(mockedReadFileSync);` to outside the describe, to mock the function only once

app/services/env/__test__/environmentVariableHelpers.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      expect(mockedReadFileSync).toHaveBeenCalledWith(path, "utf8");
+    });
+
+    it("can be read from process.env when local file is missing", () => {
+      mockedReadFileSync.mockReturnValue(undefined);
+      vi.stubEnv(key, value);
+      expect(config()[key]).toBe(value);
+    });
+
+    it("return a default empty string", () => {
```
aaschlote, line 59: you create a generic tests using `describe.each` and this is ok. However if once of the config variable the default value is not string, how would you make sure this test doesn't fail in this generic test?

app/services/env/environmentVariableHelpers.ts
```diff
@@ -0,0 +1,67 @@
...
+    ""
+  );
+}
+export function getStrapiAccessKey() {
+  return (
+    readSecretVolume("/etc/strapi-access-key-secret/password") ??
+    process.env.STRAPI_ACCESS_KEY ??
+    ""
+  );
+}
```
aaschlote, line 16: let's add an empty space between the functions 😃

aaschlote (approved):
LGTM again, let's remove the test `Expect the server config to contain a valid url to strapi` in the file `app/services/env/__test__/index.test.ts` as we don't need anymore :)

docker-compose.yml
Spencer6497, line 1: @aaschlote is this kind of what you had in mind re: containerization? works for me locally

docker-compose.yml
```diff
@@ -18,6 +18,19 @@ services:
...
+    container_name: frontend
+    depends_on:
+      - redis
+    ports:
+      - "3001:3000"
+    command: npm run start
+    environment:
+      REDIS_ENDPOINT: 172.17.0.1:6380 # Expose running redis container port
+    env_file: .env
+
```
Spencer6497: @aaschlote is something like this what you had in mind?

PR #2479, opened 2025-08-20
Optimise error announcement

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-540


pgurusinga (approved):
please double check with VoiceOver and NVDA

PR #2481, opened 2025-08-20
feat: clean up current implementation TGA

pgurusinga:


aaschlote:
It's missing the remove the prefix of the formular of the `geld-einklagen` in the file `app/routes.ts`

PR #2490, opened 2025-08-22
refactor: move berh persoenliche daten to page schemas

ingriddorioschulze:
This PR moves BerH Persönliche Daten to pageSchemas

### DIFF
--- a/app/domains/beratungshilfe/formular/persoenlicheDaten/__test__/doneFunctions.test.ts
+++ b/app/domains/beratungshilfe/formular/persoenlicheDaten/__test__/doneFunctions.test.ts
@@ -11,6 +11,7 @@ describe("beratungshilfePersoenlicheDatenDone", () => {
     houseNumber: "123",
     plz: "12345",
     ort: "ABC",
+    telefonnummer: "12233",
   };
 
   it("should return true when all required fields are present", () => {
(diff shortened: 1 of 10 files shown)


app/domains/beratungshilfe/formular/persoenlicheDaten/__test__/doneFunctions.test.ts
```diff
@@ -11,6 +11,7 @@ describe("beratungshilfePersoenlicheDatenDone", () => {
     houseNumber: "123",
     plz: "12345",
     ort: "ABC",
+    telefonnummer: "12233",
```
Spencer6497, line 14: we shouldn't need the user to actually enter a phone number for the `doneFunction` to return true, is this a leftover from a previous iteration?

Spencer6497 (approved):

app/domains/beratungshilfe/formular/persoenlicheDaten/__test__/doneFunctions.test.ts
Spencer6497: when I remove the `telefonnummer` entry from `validContext`, the test fails? It should pass when the user doesn't enter a phone number

PR #2504, opened 2025-08-26
feat(courtData): remove unused partnerCourt logic

chohner:
this wasn't used anymore but I'm not sure whether the plan was to use it again in the future?


melgaafary (approved):
We can put it back again if it's needed in the future?

m0dh4x:
same like @melgaafary says..if we needed in the future we can put it back.

aaschlote (approved):
Never have seen this logic before and as far I got, the logic of `partnerCourt` is based more in the flows right now.

LGTM!

PR #2526, opened 2025-08-29
Keep alt text empty in case it's not provided in CMS

aaschlote:
### Changes

- Keep element `alt-text` for image as empty string in case alternative text is not provided in the CMS
- Add `aria-hidden` true in case alternative text is not provided in the CMS for `svg` images

### Disclaimer

For images from `RichText` (or markdown) they are rendering current correct, without the element `alt-text` when it's not provided the alternative text in the CMS.

image

Ticket: https://digitalservicebund.atlassian.net/browse/ZOV-498


app/components/common/Image.tsx
```diff
@@ -12,13 +12,17 @@ export type ImageProps = Readonly<{
...
 
 function Image({ url, ariaHidden, alternativeText, ...props }: ImageProps) {
   const jsAvailable = useJsAvailable();
 
   const isSvg = url.endsWith(".svg");
   const altText =
-    !alternativeText || alternativeText === "" ? "image" : alternativeText;
+    !alternativeText || alternativeText === ""
+      ? EMPTY_ALTERNATIVE_TEXT
+      : alternativeText;
```
chohner, line 25: isn't this just `alternativeText ?? ""` ?

app/components/common/__test__/Image.test.tsx
```diff
@@ -52,12 +52,26 @@ describe("Image", () => {
...
     expect(image).toBeInTheDocument();
     expect(image).not.toHaveAttribute("aria-hidden");
   });
+
+  it("should render an image with empty alt text as decorative if alternativeText empty", () => {
+    const { container } = render(<Image url="photo.jpg" alternativeText="" />);
+    const image = container.querySelector("img");
+    expect(image).toBeInTheDocument();
+    expect(image).toHaveAttribute("alt", "");
+  });
```
chohner, line 69: I dont think this test is strictly necessary...

chohner (approved):

app/components/common/Image.tsx
```diff
@@ -12,13 +12,14 @@ export type ImageProps = Readonly<{
...
 function Image({ url, ariaHidden, alternativeText, ...props }: ImageProps) {
   const jsAvailable = useJsAvailable();
 
   const isSvg = url.endsWith(".svg");
-  const altText =
-    !alternativeText || alternativeText === "" ? "image" : alternativeText;
+  const altText = alternativeText ?? EMPTY_ALTERNATIVE_TEXT;
 
   const ImageComponent = (
     <img
```
pgurusinga, line 25: what about the aria-hidden in here? should it have same logic as the SVG? 🤔

aaschlote: I thought about that, but following this info [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA#before_using_aria), it's mentioned to always prefer to use semantic html for a11y usage. But, the `svg` element doesn't have any `alt-text` and to make decorative is needed to use the `aria-hidden`.

chohner: not exactly sure of the context but afaik svg's `title` tag is used as its alt-text. but sounds like a different PR either way

app/components/common/Image.tsx
```diff
@@ -50,7 +51,8 @@ function Image({ url, ariaHidden, alternativeText, ...props }: ImageProps) {
       src={url}
       title={altText}
       role={SVG_ROLE}
-      aria-hidden={ariaHidden}
+      // If the alt text is empty, the image is decorative, so we set aria-hidden to true
+      aria-hidden={altText === EMPTY_ALTERNATIVE_TEXT ? true : ariaHidden}
```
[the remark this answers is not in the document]
aaschlote: yes, but it depends of the logic when the `alt-text` is empty, but it also can receive an empty space like this " " (it's also considerate as decorative image for a11y tool), so I added the logic to check the value of the variable `EMPTY_ALTERNATIVE_TEXT` in case it's needed to change in the future.

pgurusinga (approved):

PR #2528, opened 2025-08-29
Post-refactor cleanup

chohner:
probably only helpful to review commit-by-commit.

mostly cleaning unused guards, plus fixing inconsistencies when re-using shared stuff


Spencer6497 (approved):
excellent work, thanks for this 😇

PR #2549, opened 2025-09-04
replace tsc with vite-node for unusedStrapiEntries script

rbrtrfl:


aaschlote:
> @aaschlote why would you bundle it in the dependabot group under `vitest`? it's not `vitest` related, but `vite` related, no?

The dependency is bundled in the `vitest` environment, check it out [here](https://github.com/vitest-dev/vitest/tree/main/packages/vite-node).

chohner:
Seems like the `vite-node` dependency got removed again from the `devDependencies` but included it in the dependabot group. I would strongly vote for keeping it in the `package.json`, because we rely on it and if it ever gets unbundled we break without realizing it.

chohner (approved):
lgtm

PR #2618, opened 2025-09-18
Draft version of Stepper component

aaschlote:
First draft version of the Stepper component. The UX mockup is not done yet, so probably we will modify the component soon.

image


chohner:
could we get a screenshot? 🙏🏼

PR #2630, opened 2025-09-22
feat(knip): use knip to find unused module

pgurusinga:
#### Adds a knip.ts configuration that
- registers a simple CSS compiler to convert @import/@plugin/@config CSS at-rules into fake JS imports so Knip can analyse them
- configures various ignore rules (e.g. @vitest/coverage-v8 and generated +types/root/router types).
- Set **warning** for now for any violation

#### Next steps:
- Fix or ignore errors as identified in the exploration phase of the referenced PR.


aaschlote (approved):
LGTM!

PR #2639, opened 2025-09-23
Implement stepper config

aaschlote:
### Changes

- Add `useStepper` parameter
- Generate navigation props with the current subStates in case `useStepper` is true
- Add parameter to show `unreachableSubSteps` in case `useStepper` is enabled. This parameter is needed, due the in ZOV a few subStates are not reachable but they need to be shown in the Flow Navigation

### DIFF
--- a/app/domains/beratungshilfe/formular/__test__/testcases.ts
+++ b/app/domains/beratungshilfe/formular/__test__/testcases.ts
@@ -77,6 +77,7 @@ export const testCasesBeratungshilfeFormularDefault = [
       "/finanzielle-angaben/eigentum/wertgegenstaende/wertgegenstaende-frage",
       "/finanzielle-angaben/eigentum/grundeigentum/grundeigentum-frage",
       "/finanzielle-angaben/ausgaben/ausgaben-frage",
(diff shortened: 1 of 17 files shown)


app/services/flow/userDataAndFlow/getUserDataAndFlow.ts
```diff
@@ -11,8 +11,9 @@ import { validateStepIdFlow } from "./validateStepIdFlow";
 import { getPageAndFlowDataFromPathname } from "../getPageAndFlowDataFromPathname";
 import { getPrunedUserDataFromPathname } from "../getPrunedUserDataFromPathname";
 import { type UserDataWithPageData } from "../pageData";
-import { FeatureFlag } from "~/services/isFeatureFlagEnabled.server";
+import { type FeatureFlag } from "~/services/isFeatureFlagEnabled.server";
 import { throw404IfFeatureFlagDisabled } from "~/services/errorPages/throw404";
+import { Flow } from "~/domains/flows.server";
```
pgurusinga, line 16:
```suggestion
import { type Flow } from "~/domains/flows.server";
```

PR #2668, opened 2025-09-30
refactor(useStepper): move config into xstate Meta

chohner:
keeps the config in the xstate abstraction to pass around less variables


app/domains/geldEinklagen/formular/index.ts
```diff
@@ -7,10 +7,10 @@ export const geldEinklagenFormular = {
   config: {
     id: "/geld-einklagen/formular",
     initial: "gericht-pruefen",
+    meta: { useStepper: true },
```
aaschlote, line 10: As mentioned in the [thread](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/2639#discussion_r2372771231), the stepper should be treated as a configuration for the flow as a whole. By adding it as Meta, you also make it possible to configure it at lower levels, where it doesn’t really make sense.

I understand that you only read the useStepper value from the root of the flow, but since it can still be defined in lower levels, this introduces unnecessary complexity and makes the setup harder to maintain.

I’d strongly recommend keeping the configuration as it was previously.

Spencer6497 (approved):
Nice! fwiw I would also loosely prefer this approach, just for it's simplicity and since possibly erroneously defining `useStepper` causes no harm, but if there are huge downsides I'm not seeing then I would hold off

PR #2684, opened 2025-10-02
chore(deps): Upgrade to Node v24

Spencer6497:
This PR upgrades our app to Node v24, available as LTS from 28.10. https://github.com/nodejs/Release?tab=readme-ov-file#release-schedule
Currently, the only thing that seemingly breaks is two tests that rely on jsdom's polyfill implementation of AbortController, something that has been [fixed](https://github.com/vitest-dev/vitest/pull/8390) as a part of Vitest 4, but it's unclear when this will release.

See [this thread](https://digitalservicebund.slack.com/archives/C04MC0KTHJA/p1759393979714069?thread_ts=1759307552.838399&cid=C04MC0KTHJA) for a more in-depth discussion.


chohner:
I would ask to wait at least until LTS.

The security issues are not with node but with ssl / crypto libs inside the container and are also present on the node 24 image. once a new alpine image will be cut these will be fixed for both 22 and 24

.nvmrc
```diff
@@ -1 +1 @@
-v22.16.0
+v24.9.0
```
chohner: 24.10

PR #2738, opened 2025-10-15
refactor(PKH): improve abzuege array schema

ingriddorio:
This PR refactors the array schema for Abzüge in PKH

### DIFF
--- a/.github/workflows/npm-commands.yml
+++ b/.github/workflows/npm-commands.yml
@@ -19,9 +19,6 @@ jobs:
         with:
           fetch-depth: 0
       - uses: ./.github/actions/cached-checkout-install
-      - name: Remove storybook file
-        if: ${{ matrix.npmCMD == 'build' }}
-        run: rm -r -f build/client/storybook ## Remove storybook path to avoid CI build problems
       - run: npm run ${{ matrix.npmCMD }}
         env:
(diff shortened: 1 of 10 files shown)


melgaafary (approved):

app/domains/prozesskostenhilfe/formular/finanzielleAngaben/abzuege/pages.ts
```diff
@@ -1,12 +1,45 @@
...
+    z.object({
+      ...sharedAbzuegeFields,
+      zahlungsfrequenz: z.literal("yearly"),
+    }),
+    z.object({
+      ...sharedAbzuegeFields,
+      zahlungsfrequenz: z.literal("one-time"),
+    }),
+  ])
+  .array();
```
rbrtrfl: There is still the `financialEntryInputSchema` in `app/domains/shared/formular/finanzielleAngaben/userData.ts` which is used in 5 more locations of PKH.
Ultimately, all those 6 `pageSchema`'s will make use of one shared and refactored `financialEntryInputSchema`.
Even if you want to keep this PR small and only refactor for PKH / finanzielleAngaben / abzuege, I would at least move your refactored schema to `userData.ts` and stick with a generic name like `financialEntryArraySchema`.

rbrtrfl (changes requested):
Left one comment regarding location an naming, but the schema itself lgtm!

PR #2739, opened 2025-10-15
refactor(Bucket Dependencies): Embed bucket images into content.json

Spencer6497:
This PR modifies the dumpCmsToFile.ts script to fetch and embed the bucket images into the content.json file, encoded in base64. This ensures that deleted Strapi images only appear deleted to end users after a successful content deploy, removing the production dependency on our images bucket.

This is a proof-of-concept. The script works and is performant, but the average file size of content.json jumped from ~3MB to ~20MB.

### DIFF
--- a/.github/workflows/build-content.yml
+++ b/.github/workflows/build-content.yml
@@ -1,7 +1,6 @@
 name: buildContentImage
 on:
   workflow_dispatch:
-  pull_request:
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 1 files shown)


aaschlote (commented):
Although these changes bring significant improvements to the application, they also substantially increase the HTML size for pages containing many non-SVG images. This happens because the HTML is delivered to the client with images embedded as Base64 URLs.

Example:
Page: `/kitchensink`

Current HTML size: 20 KB
After PR changes: 144 KB

This represents a 7× increase in page size, which could negatively impact performance on mobile devices. However, we believe this won’t be a major issue in practice, as pages with many non-SVG images are rarely created by content creators.

Next steps / Improvements to include in this PR:

1. Implement monitoring of the content file size to avoid unexpected growth (e.g., up to 1 GB).
2. Add tests for `StrapiImage.ts`.

PR #2744, opened 2025-10-16
refactor(PKH Array Schema): improve eigentum array schema

ingriddorio:
This PR refactors the array schema for Eigentum in PKH


rbrtrfl (commented):
I've left a few comments about missing permutations in the schemas.
Since there are quite a lot permutations it might be a good idea generating them programmatically instead of hardcoding every one of them. But I'd be interested what the others say @melgaafary @Spencer6497?

For my taste, 500 lines in `app/domains/prozesskostenhilfe/formular/finanzielleAngaben/eigentum/pages.ts` are already far too much for one file. Plus the missing permutations will let the schema definitions grow even bigger. I would suggest placing them in separate files, so we might have to come up with a convention for folder name and location.

app/domains/prozesskostenhilfe/formular/finanzielleAngaben/eigentum/pages.ts
```diff
@@ -52,6 +45,264 @@ const befristetArtSchema = z.enum([
...
+  hasArbeitsweg: YesNoAnswer,
+  wert: kraftfahrzeugWertSchema,
+};
+
+const kraftfahrzeugeArraySchema = z
+  .union([
+    z
+      .object({
+        ...sharedKraftfahrzeugeFields,
+        kontoEigentuemer: z.literal("myself"),
```
rbrtrfl: `kontoEigentuemer` should be `eigentuemer`

aaschlote:
sorry, update this PR by mistake

PR #2754, opened 2025-10-17
refactor(PKH Array Schema): improve andere unterhaltszahlung array schema

ingriddorio:
This PR refactors the array schema for Andere Unterhaltszahlung in PKH


Spencer6497 (approved):
perhaps a more meta question: does this array even need to be improved, as the shared fields are actually common to every family relationship? E.g. There's no technical difference between `father` and `ex-spouse`? The new shared schema reads a bit hard

rbrtrfl (commented):
Left one suggestion for code change. After discussion with @Spencer6497 there's no need hard coding all the possible options, since they all form the same object aka have no impact on the other properties in the object.

PR #2777, opened 2025-10-22
[RAST-317] Auto generated summary - BerH. only

melgaafary:
Designs : https://www.figma.com/design/gKylnLUREaPll4hBqWCrTl/%F0%9F%9F%A3-RAST-%7C-Ready-To-Send?node-id=28600-6972&t=xAmD4cflahMEI7vg-0

Ticket: https://digitalservicebund.atlassian.net/jira/software/c/projects/RAST/boards/688?selectedIssue=RAST-317

How this works:
1. Creates the sections from the navigation menu
2. For each user data it gets the page used by this user data and gets the question and the options available if any
3. Creates sections and boxes. Each box needs the translated selected option, in userData we have the raw value, so using the  options we go we get the chosen option and display that

There is a new component for the summary, because the current one uses Strapi, this one doesnt.

This can be used in ZOV too, but for now its only enabled in BerH

Remaining todos:

- [x]  Create boxes for Array pages
- [x]  Enable Feature Toggle
- [x] Add tests
- [x] fix SonarQube issues

Note: Sorry for the large PR, I tried my best to make it small but failed miserably.. it's just one complex beast of a feature 
Note 2: I will create a documentation file for how this works - in Confluence. And I will do a share back session as well

### DIFF
--- a/app/services/summary/fieldEntryCreation.ts
+++ b/app/services/summary/fieldEntryCreation.ts
@@ -1,7 +1,6 @@
 import type { AllowedUserTypes, UserData } from "~/domains/userData";
 import type { FieldItem } from "./types";
 import { formatFieldValue } from "./formatFieldValue";
-// import { isUserDataFieldEmpty } from "./fieldValidation";
 import { getUserDataFieldLabel } from "./templateReplacement";
 import { createArrayEditUrl } from "./arrayFieldProcessing";
 import { parseArrayField } from "./fieldParsingUtils";


app/components/content/summaryOverview/AutoGeneratedSummary.tsx
```diff
@@ -0,0 +1,224 @@
...
+          const itemId = item.id ?? index;
+          const startOpened = shouldPrint || openSections.has(itemId);
+
+          return (
+            <SummarySection
+              key={itemId}
+              item={item}
+              itemId={itemId}
+              startOpened={startOpened}
+              onGlobalToggle={handleSectionToggle}
```
Spencer6497: I think the naming here is a bit confusing, why is this called global toggle if we're only toggling one section?

app/services/summary/__test__/autoGenerateSummary.test.ts
```diff
@@ -0,0 +1,464 @@
...
+
+      expect(result.length).toBeGreaterThan(0);
+
+      // Should have both persoenliche-daten and finanzielle-angaben sections
+      const sectionIds = result.map((section) => section.id);
+      expect(sectionIds).toContain("persoenliche-daten");
+      expect(sectionIds).toContain("finanzielle-angaben");
+    });
+  });
+
```
rbrtrfl: These three tests use the exact same assert- and act-data and therefore have a lot of overlap in the assert section. I think this could either be one test or the tests should be more distinct about the different parameters that are being tested.

app/services/summary/autoGenerateSummary.ts
```diff
@@ -0,0 +1,195 @@
...
+    fields: nonArrayFields,
+    arrayGroups: arrayGroups.length > 0 ? arrayGroups : undefined,
+  };
+}
+
+export async function generateSummaryFromUserData(
+  userData: UserData,
+  flowId: FlowId,
+  flowController?: FlowController,
+  translations?: Translations,
```
rbrtrfl: When would those two properties be optional?

Spencer6497 (approved):
_WHEW_ we made it 🙏 excellent work, excellent PR, left a few notes but super stoked to see this come alive! ⭐

app/services/summary/templateReplacement.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (Array.isArray(arrayData) && arrayData[arrayIndex]) {
+      const arrayItem = arrayData[arrayIndex] as Record<string, unknown>;
+
+      // Replace template variables like {{kind#vorname}}, {{kind#nachname}}
+      return questionText.replaceAll(
+        /\{\{(\w+)#(\w+)\}\}/g,
+        (match: string, arrayType: string, fieldKey: string) => {
+          if (
+            arrayType === baseFieldName ||
+            (arrayType === "kind" && baseFieldName === "kinder")
```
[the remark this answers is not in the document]
melgaafary: its an edge case yea, the baseFieldName for kinder is "kinder" but the variable in Strapi is "kind#vorname"

app/services/summary/templateReplacement.ts
```diff
@@ -0,0 +1,58 @@
+import type { UserData } from "~/domains/userData";
+import { parseArrayField } from "./fieldParsingUtils";
+
+function replaceTemplateVariables(
```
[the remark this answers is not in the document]
melgaafary: this is happening in the server. AFAIK mustache only works in the frontend? I could be wrong

rbrtrfl (commented):
Wow! What a hulk!
Because it's such a big PR, I will write review(s) in chunks.

One comments on performance: Locally, when pointing to Strapi staging, the page needs over 30sec to be loaded. With local content, it's fine in the range of ms. Not sure why that is and might be addressed.

app/services/summary/arrayFieldProcessing.ts
```diff
@@ -0,0 +1,98 @@
+import type { UserData } from "~/domains/userData";
+import { parseArrayField } from "./fieldParsingUtils";
+import { fieldIsArray } from "~/services/array";
+
+function isArrayField(fieldName: string, userData: UserData): boolean {
+  const value = userData[fieldName];
+  return Array.isArray(value) && value.length > 0;
+}
```
rbrtrfl: A function with this functionality already exists: `arrayIsNonEmpty`.

app/services/summary/arrayFieldProcessing.ts
```diff
@@ -0,0 +1,98 @@
...
+      // Only add fields that actually exist in the userData
+      if (arrayItem && subField in arrayItem) {
+        const expandedField = `${fieldName}[${index}].${subField}`;
+        expandedFields.push(expandedField);
+      }
+    }
+  }
+
+  return expandedFields;
+}
```
rbrtrfl: I think we already have a function that does this: `resolveArrayCharacter`.

app/services/summary/getFormQuestions.ts
```diff
@@ -0,0 +1,275 @@
...
+  fieldToStepMapping: Record<string, string>,
+  flowId: FlowId,
+): Promise<Record<string, FieldQuestion>> {
+  const fieldQuestions: Record<string, FieldQuestion> = {};
+  const stepPagesCache: Record<string, StrapiFormFlowPage> = {};
+
+  const flow = flows[flowId];
+  if (!flow) {
+    throw new Error(`Unknown flowId: ${flowId}`);
+  }
```
rbrtrfl: This condition should not be necessary, since it's already done in `formular` -> `getUserDataAndFlow` -> `getPageAndFlowDataFromPathname` -> `parsePathname`

app/services/summary/arrayFieldProcessing.ts
melgaafary: very close but its not the same

app/services/summary/autoGenerateSummary.ts
```diff
@@ -0,0 +1,189 @@
...
+    groupFieldsByArrayType(allFields);
+  const arrayGroups = buildArrayGroups(arrayFieldsByBase, translations);
+
+  return {
+    id: sectionName,
+    title: sectionTitles[sectionName] ?? sectionName,
+    fields: nonArrayFields,
+    arrayGroups: arrayGroups.length > 0 ? arrayGroups : undefined,
+  };
+}
```
rbrtrfl: Suggestion: Move those four functions to a separate file(like `fieldGrouping*.ts`)

PR #2799, opened 2025-10-27
feat(MissingDataList): automatic missing data overview

chohner:
We have two extra pages that trigger data validation and show links to parts that still have data missing. this was done manually using string replacements, which is both very cumbersome, error-prone  (see https://digitalservicebund.atlassian.net/browse/RAST-404) and results in really bad markup

this new component simply re-uses the existing menu props:

image

- this is a fix until this might get integrated into an overview page
- we will adapt the strapi page once this change is live (we are fine with having items twice on the page)

follow-up: remove all strapi content and corresponding string replacements


aaschlote (approved):
LGTM!

ingriddorio (approved):

PR #3306, opened 2026-03-03
feat(DateInput): change date display on Summary Page 

ingriddorio:
This PR changes how the date is displayed in the Summary Page


app/services/validation/date.ts
```diff
@@ -11,6 +15,29 @@ const isValidDate = (date: string) =>
...
+  | { day: number; month: number; year: number };
+
+export const toDateStringFromSplitDate = (
+  value: SplitDate | "" | undefined,
+) => {
+  if (value && typeof value === "object") {
+    const { day, month, year } = value as {
+      day: string | number;
+      month: string | number;
+      year: string | number;
```
pgurusinga: don't think we need an `as`?

PR #3373, opened 2026-03-12
feat(nested arrays) arrayField approach PoC

rbrtrfl:
see `doc/erbschein-wegweiser-poc.md`


app/domains/erbschein/wegweiser/pages.ts
```diff
@@ -1,7 +1,24 @@
...
+);
+
+const kinderArraySchema = z.array(
+  z
+    .object({
+      vorname: stringRequiredSchema,
+      istVerstorben: YesNoAnswer,
+      enkelkinder: enkelkinderArraySchema,
+    })
+    .partial(),
```
chohner: also - why partial?

app/domains/erbschein/wegweiser/xStateConfig.ts
```diff
@@ -152,5 +155,86 @@ export const erbscheinWegweiserXstateConfig = {
...
+                        }) => {
+                          // If child is deceased, go to enkelkinder
+                          const indexes = (context as Record<string, unknown>)
+                            .pageData as
+                            | { arrayIndexes?: number[] }
+                            | undefined;
+                          const idx = indexes?.arrayIndexes?.[0] ?? 0;
+                          return context.kinder?.[idx]?.istVerstorben === "yes";
+                        },
+                        target: "#enkelkinder-uebersicht",
```
Spencer6497: i think maybe there's been a misunderstanding... the way this should function:
* user indicates if a single person (either Erblasser themselves or individual children) had children
* When yes, give the number of children
* Display variable number of "children" and give their info (i.e. name/verstorben pairs)
* Of the given, variable number of children, go into the first one that is deceased

(see the [Figma](https://www.figma.com/design/htcZUiIQPcbFMX0AczgsNo/A2J-%7C-RAST-Erbschein?node-id=2371-12439&t=Xy0ybAOmT4s3ROLK-4) for more info). I'm not sure that this solution will work for us in this regard :/

app/services/flow/server/types.ts
```diff
@@ -3,7 +3,11 @@ import type { Guards } from "~/domains/guards.server";
 import { type UserData } from "~/domains/userData";
 import { type ArrayConfigServer } from "~/services/array";
 
-export type NavigationEvent = "SUBMIT" | "BACK" | ArrayConfigServer["event"];
+export type NavigationEvent =
+  | "SUBMIT"
+  | "BACK"
+  | ArrayConfigServer["event"]
+  | `add-${string}`; // Support nested array events (arrayField approach)
```
Spencer6497: would this erase the typing benefits we get by also forming a union with `ArrayConfigServer["event"]`?

app/services/array/resolveArrayField.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Resolves a plain field name to its full lodash-style data path
+ * using the array scope (collected from the tree) and URL indexes.
+ *
+ * Example:
+ *   resolveArrayField("vorname", ["kinder"], [0])
+ *   → "kinder[0].vorname"
+ *
+ *   resolveArrayField("name", ["kinder", "schulen"], [0, 1])
+ *   → "kinder[0].schulen[1].name"
```
Spencer6497: nit: i think it reads nicer when the `arrayFields` come first in the parameter order, then the `fieldName`

Spencer6497 (commented):
Great work on this, very well-thought-out! I'm not sure if it'll be the solution that we need, but it's a good start and I'm a huge fan of doing away with the hashes.

Furthermore -- it seems clear to me that with this solution we traverse "deep" before going "broad" (i.e. Erblasser -> single kind -> single enkelkind -> single urenkelkind instead of Erblasser -> all kinder -> first applicable kind -> all enkelkinder).... but my question would be: how do we go broad at all?

i.e. after we've explored a leaf to the maximum depth and go back up one level to the overview page, how do we know which next child to jump into? how do we indicate that to the user, on an overview page, that a leaf needs to be "explored"? Would we then track which indices, per-level, have been fully explored?

doc/nested-arrays-arrayfield-poc.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Design draft coverage
+
+### Erbschein — two variants
+
+| Variant                                     | Description                                                                          | Supported by this PoC?                                                                                      |
+| ------------------------------------------- | ------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------- |
+| **Variante geschlossen**                    | User states number of children upfront → that many input fields rendered on one page | No — requires dynamic form generation (novel solution), unrelated to `arrayField`                           |
+| **Variante Zusammenfassung** (open/summary) | User adds children one-by-one via overview page → add child → back to overview       | **Yes** — this is exactly what the PoC models with `add-kinder` events and the overview/daten state pattern |
```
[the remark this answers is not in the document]
rbrtrfl: Each child is fully completed (including all their grandchildren) before returning to the parent overview. So it's a 'deep' approach.

rbrtrfl (commented):
I tried answering most of your comments. To enable the 'geschlossen' approch, parts of this PoC will have to be revisited. That makes things much more complicated 😬
We will try.

app/domains/erbschein/wegweiser/pages.ts
Spencer6497: I gotta say... This is promising, but I really don't like the fact that there are sooo many separate schemas defined. @chohner and I came up with a [recursive, universal schema](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/482d71def563a674007f68040011683d6dc486a3/app/domains/erbschein/wegweiser/pages.ts#L7) that could work with some minor adjustments, what do you think? If not, I'd still like to see some unified, generic recursive schema that could be applied across children/grandchildren/siblings/etc

PR #3517, opened 2026-04-07
TGA: add "Euro" word for currency format in overview page

pgurusinga:


app/components/content/summaryOverview/SummaryOverviewBoxItem.tsx
```diff
@@ -51,7 +52,15 @@ const SummaryOverviewBoxItem = ({
...
 }: Props) => {
-  const itemValue = getItemValueBox(translations, userData, inlineItems);
+  const rawItemValue = getItemValueBox(translations, userData, inlineItems);
+  const detectedAsCurrency = isCurrency(rawItemValue, {
+    thousands_separator: ".",
+    decimal_separator: ",",
+    require_decimal: true,
+  });
+
+  const itemValue = detectedAsCurrency ? `${rawItemValue} €` : rawItemValue;
```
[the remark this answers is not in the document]
chohner: funny this is coming up now - we actually have a [refactoring proposal](https://digitalservicebund.atlassian.net/browse/RAST-688) on this :)

tldr: have an (optional) bi-directional front-end <> database conversion as a codec, and an (optional) database -> viewLayer conversion tied to it

this should allow us to keep string-representation conversion directly next to the schema, and unify it everywhere (summary, array, pdfs..)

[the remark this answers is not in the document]
aaschlote: @pgurusinga About your comments:

- Keeping Strapi content-focused: The idea of the suffix, it's not changing the format values or something related, but given the content creator an option to enter an extra description in the summary component. For example:
  item -> anwaltskosten -> suffix: Euros. 
We should keep in mind that, that the suffix is editable in the CMS, so in case they decided to change Euros to $ or anything else, the change can be done via CMS and not source code.

- Consistency across the application: I see you mentioned (`risk of inconsistencies across pages`), it doesn't have risk for more than one page, we only need it for the Summary page, and we have a summary translation specific for that. The translations has been already being used, for the airports, airlines and other condition situations.

- Question around pageSchemas: The idea instead of checking the value to describe on that is `currency` you can already check what we define for field in the `pageSchemas`. Example: The field is `anwaltskosten` and it uses the schema `buildMoneyValidationSchema`. If the field uses this schema, the value is currency, because is define in the `pageSchemas`.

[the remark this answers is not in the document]
pgurusinga: Regarding the Euro suffix in Strapi: I’ve clarified this with the content and design, and they prefer not to move this logic into the translations.
I will check the pageSchema logic

app/domains/geldEinklagen/formular/klage-erstellen/pages.ts
```diff
@@ -111,7 +112,9 @@ export const geldEinklagenKlageErstellenPages = {
   forderungGesamtbetrag: {
     stepId: "klage-erstellen/forderung/gesamtbetrag",
     pageSchema: {
-      forderungGesamtbetrag: buildMoneyValidationSchema({ max: 1000000 }),
+      forderungGesamtbetrag: buildMoneyValidationSchema({
+        max: 1000000,
+      }).meta({ description: formatCurrencyZodDescription }),
```
aaschlote, line 117: maybe we can add the ``meta({ description: formatCurrencyZodDescription })`` directly into the schema, I don't see much problems having it there.

app/components/kern/summaryOverview/SummaryOverviewBoxItem.tsx
```diff
@@ -52,11 +45,17 @@ const SummaryOverviewBoxItem = ({
...
 }: Props) => {
-  const itemValue = getItemValueBox(translations, userData, inlineItems);
-  if (itemValue.trim() === "") return null;
+  const rawItemValue = getItemValueBox(translations, userData, inlineItems);
+  if (rawItemValue.trim() === "") return null;
 
   const fieldItems = extractFieldItemsFromInlineItems(userData, inlineItems);
+  const shouldAppendEuroWord = hasMoneyValidationSchema(pathname, fieldItems);
+  const itemValue = shouldAppendEuroWord
+    ? `${rawItemValue} Euro`
```
aaschlote: can we move the `Euro` to file `app/services/translations/translations.ts`?

app/components/content/summaryOverview/SummaryOverviewBoxItem.tsx
```diff
@@ -1,27 +1,20 @@
...
 
-const hasNonEmptyLongTextField = (fieldItems: FieldItems) =>
-  fieldItems.some(
-    ({ fieldName, fieldValue }) =>
-      isGeldEinklagenLongTextField(fieldName) &&
-      typeof fieldValue === "string" &&
-      fieldValue.trim().length > 0,
-  );
-
 const SummaryValueOverflowContainer = ({
```
aaschlote, line 25: I would discard for now change in the old components and only apply the changes in kern file, so we can reduce big amount of changes in this PR.

app/domains/geldEinklagen/formular/klage-erstellen/pages.ts
pgurusinga: What is the reason? Because meta would also be instantiated for non-ZOV-related schemas, which likely do not need it, I’d prefer to keep it as is so that the main schema remains neutral

aaschlote: It looks like all monetary values should include the description “Euro” in the Summary/Overview component. Since the `buildMoneyValidationSchema` is used for handling money values, it would make sense to define `meta.description` directly within this schema.

Other flows that use the same schema are unlikely to be affected, as this description is currently only utilized in the Summary/Overview component.

Additionally, this `meta.description` could potentially be leveraged in the future when rendering the `KernInputNumber` component, though this would require further evaluation.

aaschlote: but for now I wouldn't mind to keep the `meta.description` for the money value directly in zov schemas

PR #3547, opened 2026-04-14
refactor(Footer): Restructure footer after KERN implementation

Spencer6497:
This PR restructures the footer following the KERN implementation. PR created because I moved the BMJV logo to a new folder, /app/assets, the idiomatic way to statically import files in Remix. Initially I just added it to /public, but the linter and sonarcloud was throwing a fit, so instead of writing a handful of exceptions for this edge case, I decided to do things the idiomatic way. Let me know if you have questions!

**Before**
old_footer

**After**
new_footer


melgaafary:
looks good! 

just one small request, can you add some tests - you can check the old tests for footer and see if they are useful. Just to make our live easier when adding back the tests for all the Kern components

aaschlote:
what were the errors from linter and sonarcloud?

Spencer6497:
> what were the errors from linter and sonarcloud?

@aaschlote basically, it didn't like us importing the file directly from `/public`

aaschlote:
> > what were the errors from linter and sonarcloud?
> 
> @aaschlote basically, it didn't like us importing the file directly from `/public`

Did you try to set the `path` in the tsconfig.json file? You can set as the `data` folder. Alternative we can move the file to `data`, just to avoid to create a new folder for adding assets.

app/components/kern/layout/footer/footerContent.ts
```diff
@@ -25,15 +25,8 @@ export const footerContent: FooterContent[] = [
         text: "Barrierefreiheit",
```
aaschlote: can we have the `text` inside the translation file? The name of the BMJV and DS, we can keep as hardcode

app/components/kern/layout/footer/KernFooter.tsx
```diff
@@ -1,48 +1,56 @@
...
-          {image && (
-            <div className="forced-colors:bg-black">
-              <Image {...image} />
-            </div>
-          )}
-          <KernFooterExternalLinks />
+          <div className="col-start-1 col-span-12 lg:col-span-6 p-kern-space-default! gap-kern-space-small">
+            <h2 className="kern-title">Justiz-Services</h2>
+            <p className="kern-body kern-body--muted text-kern-static-medium">
+              Onlinedienste der Justiz
```
aaschlote: can we move it to translation file?

app/components/kern/layout/footer/KernFooterInternalLinks.tsx
```diff
@@ -8,13 +8,16 @@ export const KernFooterInternalLinks = () => {
...
+      className="col-start-1 lg:col-start-7 col-span-12 lg:col-span-6 flex flex-col md:flex-row gap-kern-space-default! print:hidden"
       aria-label={translations.footer.internalLinks.de}
     >
       {internalLinksSections.map((section) => {
         const ariaLabelledBy = `footer-list-${section.key}`;
         return (
-          <div key={section.key} className="flex flex-col gap-kern-space-small">
+          <div
+            key={section.key}
+            className="kern-body kern-body--bold flex flex-col gap-kern-space-small md:w-1/2 md:max-w-1/2 p-kern-space-default!"
```
aaschlote, line 19: I guess we can't get rid of the important (!), right?

melgaafary: in this case no, because kern-body has default padding. Kern doesn't have tokens to remove that, so important is necessary. Just in case you come across this in the future

PR #3549, opened 2026-04-15
Update and upgrade apk to remove vulnerabilities

aaschlote:
Before
image

After
image


pgurusinga (approved):

chohner:
I'm a bit torn on "blindly" running `apk upgrade` in the build process.

- It will upgrade OS dependencies to potentially patched versions, clearing CVEs and trivy
- It also destroys reproducibility and may introduce incompatibilities

Maybe something to discuss next dev exchange?

aaschlote:
> I'm a bit torn on "blindly" running `apk upgrade` in the build process.
> 
> * It will upgrade OS dependencies to potentially patched versions, clearing CVEs and trivy
> * It also destroys reproducibility and may introduce incompatibilities
> 
> Maybe something to discuss next dev exchange?

Yeah, I can bring this topic to the next dev exchange. However, I found weird that yesterday we had 29 vulnerabilities and now we have 11, it looks like they are fixed randomly and we are losing control of it.

PR #3550, opened 2026-04-15
refactor(CSRF): Sitewide CSRF token handling

Spencer6497:
This PR removes token rotation in favor of a single CSRF token (to rule them all). Additionally, it secures the UserFeedback form and simplifies the surrounding logic related to sessions, etc

- **create CSRF token in root loader if non-existent, secure feedback form**
- **add updateLastVisitedStep function to separate concerns**
- **(working) csrf refactor**
- **update tests**


app/services/session.server/index.ts
```diff
@@ -115,3 +119,26 @@ export const getSessionIdByFlowId = async (
...
+  if (!getCSRFFromSession(mainSession)) {
+    mainSession.set(CSRFKey, randomBytes(24).toString("base64"));
+  }
+  try {
+    const { flowId, stepId } = parsePathname(pathname);
+    mainSession.set(lastStepKey, { [flowId]: stepId });
+    // oxlint-disable-next-line no-unused-vars
+  } catch (err) {} //NOSONAR
+
+  const headers = await getSessionManager("main").commitSession(mainSession);
```
aaschlote, line 142: do you need to `commit` all the time or only when the `mainSession` has a value `set`?

app/root.tsx
```diff
@@ -133,10 +136,16 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
   globalFeatureFlags.showKernUX = Boolean(showKernUX);
   globalFeatureFlags.showFGROnlineVerfahren = Boolean(showFGROnlineVerfahren);
 
+  const { headers, feedback, csrf } = await initializeMainSession(
```
aaschlote: maybe you can call the `mainSessionFromCookieHeader` inside the `initializeMainSession` and have this function in the Promise.all. What do u think?

app/root.tsx
```diff
@@ -133,10 +136,16 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
...
+    mainSession,
+    pathname,
+  );
+
   const shouldAddCacheControl = shouldSetCacheControlHeader(
     pathname,
     trackingConsent,
   );
+  headers.set(cacheControlHeaderKey, String(shouldAddCacheControl));
```
aaschlote: wondering if we should move this line to the function `initializeMainSession` and avoid less logic in the `root.tsx`

app/components/formElements/filesUpload/useFileHandler.tsx
```diff
@@ -8,7 +8,7 @@ const handleSubmissionError = (error: unknown) =>
   logError({ message: "An error occurred during form submission:", error });
 
 export function useFileHandler() {
-  const { csrf } = useLoaderData<typeof loader>();
+  const csrf = useRouteLoaderData<RootLoader>("root")?.csrf ?? "";
```
pgurusinga, line 11: Just wondering why we default to an empty string. That makes the CSRF token useless. I think if there’s no CSRF token, this should fail.

Spencer6497: It will fail! As soon as the request hits the action, there will be a 403.

app/services/session.server/index.ts
Spencer6497: I believe only when a value is set, but the action is idempotent, so I figured this was cleaner than conditionally calling `commit` :)

aaschlote (commented):
LGTM!

Let's create later a component for the input csrf, so we don't need to import three files to have the csrf :)

pgurusinga (approved):

PR #3556, opened 2026-04-16
RAST-856 Remove useed stories and Storybook cleanup

melgaafary:
image

That's how storybook look like now


ingriddorio (approved):
🚀

stories/form/Button.stories.tsx
```diff
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react-vite";
 import KernButton from "~/components/kern/KernButton";
 
 const meta = {
-  title: "kern/KernButton",
+  title: "form/Button",
```
aaschlote: I believe this should be in the component, no? Because I can use buttons outside of a form

stories/form/ButtonContainer.stories.tsx
```diff
@@ -3,7 +3,7 @@ import KernButton from "~/components/kern/KernButton";
 import KernButtonContainer from "~/components/kern/KernButtonContainer";
 
 const meta = {
-  title: "kern/KernButtonContainer",
+  title: "form/ButtonContainer",
```
aaschlote: should we add to components folder?

stories/form/IbanInput.stories.tsx
```diff
@@ -3,7 +3,7 @@ import type { Meta, StoryObj } from "@storybook/react-vite";
 import KernIbanInput from "~/components/kern/formElements/input/KernIbanInput";
 
 const meta = {
-  title: "kern/KernIbanInput",
+  title: "form/IbanInput",
```
aaschlote: should it be in the subfolder `inputs`?

stories/feedback/CookieBanner.stories.tsx
```diff
@@ -6,7 +6,7 @@ import { GridItem } from "~/components/layout/grid/GridItem";
 import { GridSection } from "~/components/layout/grid/GridSection";
 
 const meta: Meta<typeof KernCookieBanner> = {
-  title: "kern/KernCookieBanner",
+  title: "feedback/CookieBanner",
```
aaschlote: why the cookie is in the feedback?

stories/feedback/ReportProblem.stories.tsx
```diff
@@ -9,7 +9,7 @@ import { Grid } from "~/components/layout/grid/Grid";
 import { GridItem } from "~/components/layout/grid/GridItem";
 
 const meta = {
-  title: "kern/KernReportProblem",
+  title: "feedback/ReportProblem",
```
aaschlote: why this component is in the feedback component?

stories/typography/Heading.stories.tsx
```diff
@@ -2,7 +2,7 @@ import type { Meta, StoryObj } from "@storybook/react-vite";
 import KernHeading from "~/components/kern/KernHeading";
 
 const meta = {
-  title: "kern/KernHeading",
+  title: "typography/Heading",
```
aaschlote, line 5: I liked this `typography` sub-folder 👍

aaschlote (commented):
Missing the following components as stories:

FlowStepperComponent
Video
SkipToContentLink
AutoSuggestionInput
Image

stories/feedback/ReportProblem.stories.tsx
melgaafary: Because we collect feedback from users reporting a problem? it makes sense to me, but happy to listen if you have another point of view on this

stories/feedback/CookieBanner.stories.tsx
melgaafary: Because we collect feedback from users on their cookie preference? it makes sense to me, but happy to listen if you have another point of view on this

melgaafary:
> Missing the following components as stories:
> 
> FlowStepperComponent Video SkipToContentLink AutoSuggestionInput Image

I will add those in another PR. Thanks for pointing them out tho!

stories/feedback/CookieBanner.stories.tsx
aaschlote: I see this component primarily as a way for the user to give or deny consent so we can track their journey on the page. The term “cookie” in the name feels a bit misleading, since we’re only using it to store the user’s preference, whether on the server or client side.

Regarding the folder structure, I understand your point that this component collects a form of feedback. However, we already have a UserFeedback component that explicitly gathers user input, whereas this one handles it implicitly.

If we follow the same logic, we’d also need to move all input components into a feedback folder, since they technically collect user data or feedback as well—which might not be the best way to organize things.

stories/feedback/ReportProblem.stories.tsx
aaschlote: Regarding the folder structure, I understand your point that this component collects a form of feedback. However, we already have a UserFeedback component that explicitly gathers user input, whereas this one handles it implicitly.

If we follow the same logic, we’d also need to move all input components into a feedback folder, since they technically collect user data or feedback as well—which might not be the best way to organize things.

stories/feedback/CookieBanner.stories.tsx
melgaafary: good point. I will move it to components then

PR #3573, opened 2026-04-20
refactor(CSRF): Create reusable CSRF component

Spencer6497:
This PR removes the need for prop-drilling the CSRF token all the way from the root loader to the consuming components, instead relying on a reusable <Csrf /> component that calls useRouteLoaderData("root").

- **remove prop drilling**
- **create reusable CSRF component**


app/components/formElements/__test__/CsrfInput.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { render } from "@testing-library/react";
+import { Csrf } from "~/components/formElements/Csrf";
+import { CSRFKey } from "~/services/security/csrf/csrfKey";
+
+vi.mock("react-router", async () => ({
+  useRouteLoaderData: vi.fn().mockReturnValue({ csrf: "mocked-csrf-token" }),
+}));
+
+describe("Csrf", () => {
+  it("renders the hidden input with the csrf token", () => {
```
aaschlote, line 10: this test is not doing what's described. What do I mean?

1. It's not checking if it has an `input` element
2. It's not checking if the `input` it's hidden

app/components/formElements/__test__/CsrfInput.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+    const { getByTestId } = render(<Csrf />);
+    const csrf = getByTestId(CSRFKey);
```
aaschlote, line 12: maybe we can get the html element by the name, so we don't need to set the `data-test-id` to the input

app/components/formElements/Csrf.tsx
```diff
@@ -0,0 +1,10 @@
+import { useRouteLoaderData } from "react-router";
+import { type RootLoader } from "~/root";
+import { CSRFKey } from "~/services/security/csrf/csrfKey";
+
+export const Csrf = () => {
```
aaschlote: How about renaming this to `InputCsrf`? `Csrf` is a bit too generic, and InputCsrf better reflects that this component specifically handles rendering an input field.

aaschlote (approved):
I wrote some comments, besides that LGTM!

Small note:
In the test `app/components/content/emailCapture/__test__/EmailCapture.test.tsx`, the `useRouteLoaderData` has been mocked twice, maybe we can remove improve it.

app/components/formElements/__test__/CsrfInput.test.tsx
Spencer6497: I understand point 2, but isn't point 1 rather testing configuration/implementation instead of actual behavior?

app/components/formElements/__test__/CsrfInput.test.tsx
Spencer6497: sadly react testing library [doesn't support](https://testing-library.com/docs/react-testing-library/cheatsheet/) this kind of query :/

aaschlote: you could use `getByRole`, take a look in the file `HiddenInput.test.tsx` 😃

PR #3582, opened 2026-04-21
feat(AutoSuggest): Fuzzy Search

Spencer6497:
This PR adds fuzzy searching to address-based AutoSuggest.


aaschlote:
Was this result expected?  Maybe increasing the parameter `minMatchCharLength` to 5 characters it should work.
image

app/components/kern/formElements/autoSuggest/KernAutoSuggestInput.tsx
```diff
@@ -111,9 +121,13 @@ const KernAutoSuggestInput = ({
         setOptions([]);
         return;
       }
-      let filteredOptions = items.filter((item) =>
-        item.label.toLowerCase().includes(value.toLocaleLowerCase()),
-      );
+
+      let filteredOptions =
```
aaschlote, line 136: can we add a test for that? Either an unit test or E2E Test should be fine for me.

PR #3584, opened 2026-04-22
RAST-855 clean up strapi props and remove unused components

melgaafary:


app/services/cms/models/content/StrapiInlineNotice.ts
```diff
@@ -12,8 +10,6 @@ export const StrapiInlineNoticeSchema = z
     tagName: z.enum(["h1", "h2", "h3", "h4", "h5", "h6", "p", "div"]),
     look: z.enum(["warning", "tips", "success", "error"]),
```
ingriddorio, line 11: We don't have "tips" or "error" anymore. It should be "info" | "warning" | "danger" | "success";

PR #3593, opened 2026-04-23
RAST-878 missing stepper states styles

melgaafary:
warning and current state was missing, fixed now:
image

and fixed hover state for both warning and normal
image
image


app/components/kern/navigation/FlowStepperNavigation/getStepStyles.ts
```diff
@@ -20,24 +20,31 @@ export function getStepStyles(state: NavState) {
...
-      ? "var(--color-kern-neutral-200)"
-      : stateIsDisabled(state)
-        ? "var(--color-kern-neutral-025)"
-        : stateIsWarning(state)
-          ? "var(--color-kern-feedback-warning)"
-          : "var(--color-kern-neutral-025)",
+    arrowBg:
+      stateIsCurrent(state) && !stateIsWarning(state)
+        ? "var(--color-kern-neutral-200)"
+        : stateIsDisabled(state)
```
aaschlote: we might have some sonar issues here, can you please take a look? 

Image

aaschlote (approved):
We need to fix the border in the focus (probably needs to be iterate with UX) and improving a11y (specially for HCM).

But besides that, LGTM!

pgurusinga:
The number on the first stepper seems disappear 🤔 

image

melgaafary:
> The number on the first stepper seems disappear 🤔
> 
> image

there is an accessibility ticket for this. this ticket only addressing the design issues

PR #3600, opened 2026-04-24
feat(IBAN Input): Display bank name to users

Spencer6497:
This PR adds a feature to the IBAN component where a matching bank name is displayed when the user enters an IBAN.

- **add legacy bank name display component, display on preexisitng iban value**
- **add bank name to Kern IBAN input**
- **working sr changes**
- **add comments and tests**


aaschlote (changes requested):
This change significantly increases the bundle size from 27 kB to 327 kB due to importing `data/bankCodes.json` on the front end. It may negatively impact performance, especially on mobile devices, so we should investigate ways to address it.

Possible solutions:

1. Limit the imported data to only the necessary fields from `data/bankCodes.json`. This should substantially reduce the file size, although it may not be sufficient to bring the bundle down to an acceptable level.
2. . Move the logic for resolving bank names to the back end via an internal API, similar to how we fetch options for the `AutoSuggestionInput` component. With this approach, the bank code data could also be stored in SQLite, which may improve memory usage across the cluster.


Before
Image

After
Image

app/components/kern/formElements/input/KernIbanInput.tsx
```diff
@@ -13,13 +16,54 @@ const MaskedIbanInput: FunctionComponent<MaskedIbanInputProps> = IMaskMixin<
...
 const KernIbanInput = (props: InputProps) => {
+  const field = useField<string | undefined>(props.name);
+  const iban = field.value();
+  const [bankName, setBankName] = useState<string | null>(null);
+
+  // Debounce needed to not clobber the screen reader while typing
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      const matchedBankName = bankNameFromIBAN(iban);
+      setBankName(matchedBankName || null);
```
pgurusinga: possible?
```suggestion
      setBankName(matchedBankName ?? null);
```

app/components/kern/formElements/input/IbanInput/KernIbanInput.tsx
```diff
@@ -13,13 +16,54 @@ const MaskedIbanInput: FunctionComponent<MaskedIbanInputProps> = IMaskMixin<
...
+        <output
+          id={bankNameBadgeId}
+          className="kern-badge kern-badge-info border-2 border-kern-feedback-info bg-kern-feedback-info-background w-min"
+        >
+          <KernIcon name="info" className="fill-kern-feedback-info" />
+          <span className="kern-label kern-label--small text-nowrap">
+            {bankName}
+          </span>
+        </output>
+      )}
```
pgurusinga, line 58: Should this bank name be focusable?

Image

pgurusinga:
Also +1 for making the bank name as another endpoint and the component should fetch the data

app/components/kern/formElements/input/IbanInput/KernIbanInput.tsx
Spencer6497: As far as I know, no, since it's not an interactable element, and it's already being read via screen reader

app/components/kern/formElements/input/IbanInput/KernIbanInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <div key={props.name} className="flex flex-col gap-15">
+      <MaskedIbanInput
+        mask={"**** **** **** **** **** **** **** **** **"}
+        type="number"
+        prepareChar={(str) => str.toUpperCase()}
+        ariaDescribedBy={bankNameBadgeId}
+        {...props}
+      />
+      {/* Badge display of bank name for sighted users */}
+      {bankName && (
```
aaschlote, line 50: If  I enter an Iban, it shows correctly the bank name, but if I remove the iban it doesn't remove it.

image

app/components/kern/formElements/input/IbanInput/__test__/useBankData.test.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * @jest-environment jsdom
```
aaschlote: if you rename the file to `.tsx` you won't need this

app/components/kern/formElements/input/IbanInput/__test__/useBankData.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import { renderHook, waitFor } from "@testing-library/react";
+import { type BankData } from "~/components/kern/formElements/input/IbanInput";
+import { useBankData } from "~/components/kern/formElements/input/IbanInput/useBankData";
+
+const mockFetch = vi.fn();
+globalThis.fetch = mockFetch;
+
+describe("useBankData", () => {
```
aaschlote, line 8: can we have a test if the response is not `ok`?

app/components/kern/formElements/input/IbanInput/KernIbanInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+      if (iban && banks) {
+        const matchedBankName = bankNameFromIBAN(iban, banks);
+        setBankName(matchedBankName);
+      }
+    }, 1000);
+
+    return () => clearTimeout(timeout);
+  }, [iban, banks]);
+
+  const bankNameBadgeId = "bank-name-badge";
```
aaschlote: I would move the declaration of this function outside the component rendering, so above the line 19.

app/components/kern/formElements/input/IbanInput/useBankData.ts
```diff
@@ -0,0 +1,24 @@
...
+  const [bankData, setBankData] = useState<BankData>();
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const response = await fetch("/api/banks/list");
+
+        if (response.ok) {
+          const json = await response.json();
+          setBankData(json);
```
aaschlote, line 14: let's use zod to create the `BankData` and have a schema to validate it. Just to assure the value we get from the api is valid.

app/components/kern/formElements/input/IbanInput/useBankData.ts
```diff
@@ -0,0 +1,24 @@
...
+        }
+      } catch {
+        setBankData(undefined);
```
aaschlote, line 17: what about `setBankData([])`

app/components/kern/formElements/input/IbanInput/KernIbanInput.tsx
```diff
@@ -0,0 +1,74 @@
...
+const KernIbanInput = (props: InputProps) => {
+  const field = useField<string | undefined>(props.name);
+  const iban = field.value();
+  const [bankName, setBankName] = useState<string>();
+  const banks = useBankData();
+
+  // Debounce needed to not clobber the screen reader while typing
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      if (iban && banks) {
```
aaschlote: we can move condition `banks` outside of timeout in case we don't have banks we don't need to create a timeout function.

app/routes/api.banks.list.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const loader = async ({ request }: LoaderFunctionArgs) => {
+  const validatedRequest = await validateCsrfSessionFormless(request);
+
+  if (validatedRequest.isErr) {
+    logWarning(`Error: request to banks route without CSRF token`);
+    throw new Response(null, { status: 403 });
+  }
+
+  const bankData: BankData = (
```
aaschlote, line 15: we can optimise the response body if we return the name of the bank as a key and the `Bankleitzahl` as list of the numbers, so we don't repeat all the time name of the bank: It would be something:

```json
bankName: [111,222,333],
bankName2: [444,555,666],
```

app/routes/api.banks.list.ts
```diff
@@ -0,0 +1,26 @@
+import { type LoaderFunctionArgs } from "react-router";
+import { logWarning } from "~/services/logging";
+import { validateCsrfSessionFormless } from "~/services/security/csrf/validatedSession.server";
+import bankCodes from "../../data/bankCodes.json";
+import { type BankData } from "~/components/kern/formElements/input/IbanInput";
+
+export const loader = async ({ request }: LoaderFunctionArgs) => {
```
aaschlote, line 7: I would like to see test for this file.

data/bankCodes/data.json
aaschlote, line 1: Let's move this file into `data/bankCode/data.json` to follow the same structure as the other files in the `data` folder and add the command for the creation of this file in the `scripts` folder.

aaschlote (commented):
It looks much better, I wrote some comments, check it out this little bug in this [comment](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/3600/changes#r3166199224).

Feel free the ping me if you need a better clarification.

app/components/kern/formElements/input/IbanInput/__test__/useBankData.test.ts
Spencer6497: omfg that's soooo dumb. It took me forever to try and figure out why it wasn't working!

app/components/kern/formElements/input/IbanInput/useBankData.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from "react";
+import { type BankData } from "~/components/kern/formElements/input/IbanInput";
+
+export function useBankData() {
+  const [bankData, setBankData] = useState<BankData>();
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const response = await fetch("/api/banks/list");
```
pgurusinga, line 10: Nice! I am wondering if we can use useFetcher in here: https://reactrouter.com/api/hooks/useFetcher

> Fetchers **track their own, independent state and can be used to load data**, submit forms, and generally interact with [action](https://reactrouter.com/start/framework/route-module#action) and [loader](https://reactrouter.com/start/framework/route-module#loader) functions.

What do you think? I assume this lets us skip useEffect and not worry about the dependency array.

Spencer6497: we'd still have to use useEffect because it's async... I tried it out and it felt super clunky

PR #3624, opened 2026-04-28
feat(createPrometheusMetrics): expose metrics to prometheus

aaschlote:
### Problem

Our current Grafana dashboard relies on logs generated by `createPinoHttpLogger`. Building visualizations from logs requires significant effort in Grafana, and the resulting data becomes unreliable over longer timeframes. To achieve a more accurate and complete dashboard, we need to adopt a different data source.

### Solution

Introduce Prometheus metrics using the `prom-client` library. At present, we will implement two types of metrics: `Histogram` and `Counter`. These metrics will include the following labels: path, method, statusCode, browser, and device.

### Disclaimer

While it is possible to use metrics from Traefik, they do not include values such as userAgent or path. As a result, they cannot support detailed visualizations, such as error analysis by specific userAgent or request path.


app/services/logging/createPrometheusMetrics.ts
```diff
@@ -0,0 +1,161 @@
...
+    browser = "chrome";
+  } else if (/firefox/.test(ua)) {
+    browser = "firefox";
+  } else if (/safari/.test(ua) && !/chrome/.test(ua)) {
+    browser = "safari";
+  } else if (/opera|opr/.test(ua)) {
+    browser = "opera";
+  } else if (apiClientPatterns.test(ua)) {
+    browser = "api-client";
+  }
```
pgurusinga, line 94: Shouldn't we use a user-agent parser library instead? Right now, we’re missing support for other browsers, particularly in-app browsers. check this source: https://github.com/faisalman/ua-parser-js/blob/master/src/main/ua-parser.js

app/services/logging/createPrometheusMetrics.ts
```diff
@@ -0,0 +1,161 @@
...
+type UserAgentInfo = {
+  browser: string;
+  device: "mobile" | "desktop" | "unknown";
+};
+
+const apiClientPatterns =
+  /curl|wget|python-requests|axios|postman|thunder|insomnia|restclient|fetch|httpclient/i;
+
+const mobilePatterns =
+  /mobile|android|iphone|ipad|ipod|windows phone|blackberry|opera mini|kindle/i;
```
pgurusinga, line 64: ipod has browser?

app/services/logging/createPrometheusMetrics.ts
aaschlote: It would be nice to use one, but for now I want to keep simple, because I want to check the behaviour of the metrics. Not sure if the current configuration will work, so this PR is kind of testing it.

Btw, I guess we can't use the library you mention, it looks like we need to [pay](https://github.com/faisalman/ua-parser-js#package--pricing) and their [license](https://github.com/faisalman/ua-parser-js?tab=AGPL-3.0-1-ov-file) (AGPL-3.0-1) we might need to check it deeper as we don't use any dependency with this license now in our project.

app/services/logging/createPrometheusMetrics.ts
aaschlote: not sure, copilot suggestion 😅

PR #3632, opened 2026-04-28
refactor(KERN): remove legacy design FilesUpload

ingriddorio:
This PR remove legacy design from FilesUpload, update imports and tests


melgaafary (approved):

pgurusinga:
@ingriddorio @melgaafary have we discussed this with the ZOV team? Cause we might need the FileUpload for uploading evidence

ingriddorio:
> @ingriddorio @melgaafary have we discussed this with the ZOV team? Cause we might need the FileUpload for uploading evidence

@pgurusinga what exactly needs to be discussed? As far as I know, the feature will be working as before just looking like Kern.

pgurusinga (approved):

PR #3635, opened 2026-04-29
refactor(KERN): remove legacy design Tiles

ingriddorio:
This PR remove legacy design from Tiles, update imports and tests


app/components/formElements/tile/__test__/TileContent.test.tsx
```diff
@@ -0,0 +1,18 @@
+import { render, screen } from "@testing-library/react";
+import KernTileContent from "../KernTileContent";
+
+describe("TileContent", () => {
+  it("should render title and description", () => {
+    render(<KernTileContent tileTitle="Title" tileDescription="Description" />);
```
pgurusinga: For these props, would it make sense to use just `title` instead of `tileTitle`, same for `titleDescription`? Since we already know it's a tile component, the extra `tile` prefix might be unnecessary.

ingriddorio: Done!

pgurusinga (approved):

PR #3636, opened 2026-04-29
refactor(KERN): remove unused files and exports

ingriddorio:
This PR remove unused files and exports


app/components/content/Details.tsx
aaschlote, line 11: I guess we can delete this file and for the `Details.test.tsx` use the kern file, no?

app/components/formElements/schemaToForm/renderSchemaBasedFormElement.tsx
```diff
@@ -36,32 +31,3 @@ export const isSpecialComponentDescriptions = (
   val?: any,
 ): val is SpecialComponentDescription =>
   specialComponentDescriptions.includes(val);
-
-export const renderSchemaBasedFormElement = (
```
aaschlote, line 40: what about move the code of the function `renderSpecialMetaDescriptions` from `KernSchemaComponents` to this file and use this function there?

ingriddorio: Yeah, good idea. I can do that soon. I want to keep just Kern changes for now :)

melgaafary (approved):

PR #3637, opened 2026-04-29
RAST-845 Fix progress bar padding and add padding 80px 

melgaafary:


app/services/cms/models/__test__/StrapiPadding.test.ts
```diff
@@ -1,9 +1,12 @@
-import { strapiPaddingOptions, StrapiPaddingSchema } from "../StrapiPadding";
+import {
+  strapiPaddingOptions,
+  StrapiPaddingOptionalSchema,
+} from "../StrapiPadding";
 
 describe("StrapiPadding", () => {
   it("converts all valid options", () => {
```
aaschlote: This test doesn't really do what's described. Before we have a conversation of values, now the values exist already and the conversion is not more needed. Can you please adapt the test to check if the values receive as parameters are valid?

PR #3664, opened 2026-05-05
RAST-859 remove ds icons package

melgaafary:
replaced with kern


aaschlote (approved):
Do you mind to also update the licenses? With the command `pnpm run build:licenses`?

PR #3665, opened 2026-05-05
feat(Component): Add Numeric Stepper Component

Spencer6497:
This PR implements the NumberIncrement component

- **number increment scaffolding**
- **working, performant solution**
- **finish styling component**
- **add tests and storybook story**
- **rename schema**


app/components/formElements/inputs/number/NumberIncrement.tsx
```diff
@@ -0,0 +1,113 @@
...
+          className={classNames(
+            "bg-white max-w-52 text-center [appearance:textfield] [&::-webkit-inner-spin-button]:m-0 [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-outer-spin-button]:m-0 [&::-webkit-outer-spin-button]:appearance-none",
+            {
+              "bg-kern-feedback-danger-background!": field.error(),
+            },
+          )}
+          id={name}
+          {...field.getInputProps()}
+          value={field.value() ?? 0}
+          defaultValue={undefined}
```
aaschlote: why this is `undefined`?

app/services/validation/numberIncrement.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from "zod";
+
+export const createNumberIncrementSchema = (min: number = 0, max?: number) => {
```
aaschlote, line 3: can we have an unit test for the schema?

app/components/formElements/schemaToForm/renderZodString.tsx
```diff
@@ -46,6 +47,8 @@ export const renderZodString = (
     return (
       <AutoSuggestInput key={fieldName} {...matchingElement} {...inputProps} />
     );
+  if (matchingElement?.__component === "form-elements.number-increment")
```
aaschlote: wondering if we should add the `meta` inside the schema `createNumberIncrementSchema` to render this component, because the schema basically is related to this component. But checking this condition with `form-elements.number-increment` is fine for me too.

aaschlote (approved):
I wrote minor comments, besides that LGTM!

stories/components/NumberIncrement.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    name: "numberIncrement",
+    label: "NumberIncrement",
+    errorMessages: [
+      { code: "required", text: "Please enter a number" },
+      { code: "too_low", text: "Number too low" },
```
melgaafary, line 27: does it actually make sense to allow negative numbers? just wondering

app/components/formElements/inputs/number/NumberIncrement.tsx
Spencer6497: I can't find it in the docs anymore, but basically when `defaultValue` is defined, RVF treats it as an uncontrolled input, and then it starts clashing with out increment/decrement buttons. I'll add a comment

stories/components/NumberIncrement.stories.tsx
Spencer6497: Probably not? But I figured I'd design it as flexibly as possible and we can only use what we need

PR #3692, opened 2026-05-11
feat(KERN): add suffix to input label

ingriddorio:
This PR add suffix to input label to match Kern design


app/components/formElements/inputs/text/TextInput.tsx
```diff
@@ -44,6 +45,7 @@ const TextInput = function InputComponent({
       {label && (
         <label className="kern-label" htmlFor={name}>
```
aaschlote: don't you want to create a global component for this label? It looks like it's similar everywhere in the source code.

aaschlote (approved):
LGTM.

Please double check if the VoiceOver announce the suffix when it's available for the field.

ingriddorio:
> LGTM.
> 
> Please double check if the VoiceOver announce the suffix when it's available for the field.

Good one! I checked and it is announcing fine

melgaafary (approved):

PR #3694, opened 2026-05-12
feat(IBAN): Automatic Bank Name identification and controlled form inputs

Spencer6497:
This PR introduces a new pattern to forms: form inputs controlling other form inputs. In this example, when a user enters a valid IBAN that corresponds to an existing bank, the IbanInput field updates the bankName field with the matched name. This has wider application, for example, entering a Postleitzahl and automatically filling the city.

- **move BankNameBadge to its own file and hide for now**
- **make TextInput potentially controlled, fix useEffect in IbanInput**
- **add tests**
- **add clarifying comment to controlled prop, comment out bank name badge logic**
- **add accessibility features**


app/domains/pageSchemas.ts
```diff
@@ -152,9 +155,25 @@ export function xStateTargetsFromPagesConfig<T extends PagesConfig>(
 
 export type PagesConfig = Record<string, PageConfig>;
 
+export type FieldValueChangeHandlerProps = {
+  fieldValue: AllowedUserTypes;
```
aaschlote: maybe we can just name as `value`, because the type it's mentioned about `FieldValue` etc.

app/domains/pageSchemas.ts
```diff
@@ -152,9 +155,25 @@ export function xStateTargetsFromPagesConfig<T extends PagesConfig>(
...
+  originalFieldValue: AllowedUserTypes;
```
aaschlote: same here, maybe only `originalValue`

app/domains/pageSchemas.ts
```diff
@@ -152,9 +155,25 @@ export function xStateTargetsFromPagesConfig<T extends PagesConfig>(
...
+  controlledField: FieldApi<any>;
+  setControlledFieldSrValue: Dispatch<SetStateAction<string | undefined>>;
+};
+
+export type ControlledFieldConfig = {
+  controlledFieldName: string;
```
aaschlote: maybe just rename `fieldName` as the type it's `ControlledFieldConfig`.

app/domains/kontopfaendung/pkonto/antrag/pages.ts
```diff
@@ -34,6 +39,46 @@ export const kontopfaendungPkontoAntragPages = {
...
     },
+    controlledFieldConfig: {
+      controlledFieldName: "bankName",
+      handleFieldValueChange: async ({
+        originalFieldValue,
+        fieldValue,
+        controlledField,
+        setControlledFieldSrValue,
+      }: FieldValueChangeHandlerProps) => {
+        const banks = await fetchBanks();
```
aaschlote: every time I tip a digit in the field, it's done a request to server, can we find a way to cache it? 

Image

app/domains/kontopfaendung/pkonto/antrag/pages.ts
```diff
@@ -34,6 +39,46 @@ export const kontopfaendungPkontoAntragPages = {
...
+            }, 1000);
+
+            return () => clearTimeout(timeout);
+          }
+          setControlledFieldSrValue("");
+          controlledField?.setValue("");
+        }
+      },
+      getScreenreaderAnnouncementText: (controlledFieldSrValue: string) =>
+        `Bank identifiziert: ${controlledFieldSrValue}`,
```
aaschlote: maybe can we move it to translation?

app/domains/pageSchemas.ts
```diff
@@ -152,9 +155,25 @@ export function xStateTargetsFromPagesConfig<T extends PagesConfig>(
...
+  handleFieldValueChange: (
+    props: FieldValueChangeHandlerProps,
+  ) => MaybePromise<void | (() => void)>;
+  getScreenreaderAnnouncementText: (controlledFieldSrValue: string) => string;
```
aaschlote: maybe it's better `getScreenReaderAnnouncementText` as screen reader are two words

app/components/formElements/inputs/iban/IbanInput.tsx
```diff
@@ -1,68 +1,40 @@
...
-    setBankName(undefined);
-  }, [iban, banks]);
+>((props) => {
+  // Needs to be uncontrolled, so that the value gets masked on page load
+  return <TextInput {...props} controlled={false} />;
+});
+
+const IbanInput = (
+  props: InputProps & {
+    pageConfig?: ArrayPage | PageConfig;
```
aaschlote: maybe we can have only the field `controlledFieldConfig` as props here as we don't need the whole `pageConfig` object.

app/domains/pageSchemas.ts
```diff
@@ -152,9 +155,25 @@ export function xStateTargetsFromPagesConfig<T extends PagesConfig>(
 
 export type PagesConfig = Record<string, PageConfig>;
 
+export type FieldValueChangeHandlerProps = {
```
aaschlote, line 158: I think we need to add another field here to control when these rules should apply.

For example, right now this config is only used in the IbanInput component. But if I wanted to create a similar rule for updating the city based on the ZIP code, I would need to use this config in TextInput.

The problem is that both the ZIP code and city fields use TextInput, so the hook would run twice, even though it should only run for the ZIP code field.

app/components/hooks/__test__/useControlledField.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    field: vi.fn(),
+  }),
+}));
+
+describe("useControlledField", () => {
+  it("should return a screenreader announcement component", () => {
+    const { result } = renderHook(() =>
+      useControlledField(
+        "iban",
+        kontopfaendungPkontoAntragPages.bankdatenKontodaten
```
aaschlote: I would prefer to have a test not using an existing config from the `Flow`, but creating a own config for this test.

For example: If the `IBAN` value is amazing, it should print "You entered an amazing iban value".

Having own config for this test it removes the dependency from `Flow` config.

aaschlote (commented):
I added a few comments — this looks promising overall. I think there are still some improvements we could make, but we can address them later after discussing the comments below.

app/domains/pageSchemas.ts
Spencer6497: ahh sorry I was confusing the german and english words 🙈

app/domains/pageSchemas.ts
Spencer6497: Unfortunately, with react we can't call hooks conditionally, so we could either
* conditionally render a component with the hook, or without it, in all input fields
* Create internal hook logic to basically skip the handler function

I would tend towards the second solution

aaschlote: we decided to do later one, closing this thread

app/components/formElements/inputs/text/TextInput.tsx
```diff
@@ -28,6 +25,23 @@ export type InputProps = Readonly<{
   controlled?: boolean;
 }>;
 
+const getInputValue = (field: FieldApi<unknown>, controlled: boolean) => {
+  if (field.getInputProps().value !== undefined) {
+    return field.getInputProps().value;
+  }
```
Spencer6497, line 31: I'm a bit confused by this early return here... what case does this cover?

PR #3696, opened 2026-05-12
fix(KERN): add underline on hover for links with icon

ingriddorio:
This PR add underline to links with icon on hover and rename OpenSourceLicenses to remove Kern prefix

Screenshot 2026-05-12 at 12 18 56
Screenshot 2026-05-12 at 12 18 48
Screenshot 2026-05-12 at 12 18 36
Screenshot 2026-05-12 at 12 18 56


aaschlote (approved):
LGTM!

I'm wondering if need to bring back again the `Standalone` component, so we have an unique component to render `ahref` and once we need to adapt or fix some bug, we can touch only in one place and not many.

PR #3700, opened 2026-05-12
Use shadow inset to show border on item when focus

aaschlote:
Adding shadow inset so the border is visible when focus in the NavItem

Before

image

After
image

Alternative
The main problem is the `overflow-hidden` in the `nav` html element clips anything that extends outside the element’s box. We can could remove the `overflow-hidden`, but than the result it would be something like this:

image


melgaafary:
did you check with Tom or Marion? I feel like the after color is bit pale, it might not be accessible enough? 

I like the alternative version more. But I think Tom can give the final say

aaschlote:
> did you check with Tom or Marion? I feel like the after color is bit pale, it might not be accessible enough?
> 
> I like the alternative version more. But I think Tom can give the final say

Already asked Tom in the [ticket](https://digitalservicebund.atlassian.net/browse/RAST-835?focusedCommentId=49016), waiting him to reply

tom-haake:
Thanks @aaschlote! I like the alternative version, too. It is probably visually the closest variant to what we would aim for. Personally, I would be ok with the styling, and if you are up to it, we could adopt the focus styling of our links.

PR #3704, opened 2026-05-13
fix(KERN): move aria label from aside to nav

ingriddorio:
This PR move the aria label from aside to navigation for accessibility purposes

### DIFF
--- a/app/components/navigation/FlowNavigation.tsx
+++ b/app/components/navigation/FlowNavigation.tsx
@@ -8,10 +8,9 @@ export default function FlowNavigation({
   const rootLoaderData = useRouteLoaderData<RootLoader>("root");
 
   return (
-    <aside
-      aria-label={rootLoaderData?.accessibilityTranslations?.navigationLabel}
-    >
+    <aside>
       <nav
+        aria-label={rootLoaderData?.accessibilityTranslations?.navigationLabel}
         className={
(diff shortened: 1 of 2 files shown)


app/components/navigation/FlowNavigation.tsx
```diff
@@ -8,10 +8,9 @@ export default function FlowNavigation({
   const rootLoaderData = useRouteLoaderData<RootLoader>("root");
 
   return (
-    <aside
-      aria-label={rootLoaderData?.accessibilityTranslations?.navigationLabel}
-    >
+    <aside>
       <nav
+        aria-label={rootLoaderData?.accessibilityTranslations?.navigationLabel}
```
aaschlote, line 13: this change is not possible to do it, it breaks some html rules. You can check in the [thread here](https://digitalservicebund.atlassian.net/browse/ZOV-533?focusedCommentId=38264)

aaschlote: this it'll required a change on the structure of the page (check it out this [draft](https://www.figma.com/design/o5yG2MLun2xOhW0DhyQpOr/A2J-%7C-%E2%9D%96-Shared-Pages---Patterns?node-id=4122-2054&t=scIb9tvc6AbgBsFl-0) from Marion),  where it requires a big refactoring

ingriddorio: I see, thanks for letting me know :)

PR #3737, opened 2026-05-22
fix(KERN): SR announces tile description

ingriddorio:
This PR add descriptionID to TileContent for the SR to announce the description correctly


aaschlote (approved):
nice, let's add an unit test so we don't have this issue again.

LGTM!

PR #3751, opened 2026-05-26
RAST-835 fix side nav border issue

melgaafary:
image


aaschlote:
Not sure if you have seen, but this change breaks the borders on left and right side again 
image

melgaafary:
> Not sure if you have seen, but this change breaks the borders on left and right side again image

good catch, will fix it

aaschlote (approved):
LGTM!

Maybe to check later, when focus on current NavItem, the background color changes to white, not sure if we need to keep the background color for the current (dark blue).

Image

Image

melgaafary:
> LGTM!
> 
> Maybe to check later, when focus on current NavItem, the background color changes to white, not sure if we need to keep the background color for the current (dark blue).
> 
> Image Image

will have to check with Tom. Lets make another ticket if he wants to change it

PR #3757, opened 2026-05-27
fix(KERN): stepper and sidenav responsiveness and accessibility

ingriddorio:
This PR 
- change stepper circle size, 
- make sidenav text wrap on 200%, 
- fix arrow icon size 
- fix sidenav padding top on 200%

Screenshot 2026-05-27 at 15 54 15

Screenshot 2026-05-27 at 15 12 33

Screenshot 2026-05-27 at 15 11 10

Screenshot 2026-05-27 at 17 08 33


app/components/navigation/FlowStepperNavigation/getStepStyles.ts
```diff
@@ -58,7 +58,7 @@ export function getStepStyles(state: NavState) {
     arrowHoverBg: getArrowHoverBg(state),
 
     circle: classNames(
-      "flex justify-center items-center w-[20px] h-[20px] rounded-full mr-3",
+      "flex justify-center items-center w-[24px] h-[24px] rounded-full mr-3",
```
aaschlote: maybe we should avoid to use `px` and use either `rem` or `vw`.

You can check this [article](https://medium.com/@shakirulhkhan/why-should-you-not-use-px-3021aa0b5958) to check more about it.

melgaafary: rem or vw is much better for sure. But, that needs also change from designers. All designs use px

melgaafary (approved):

app/components/navigation/FlowStepperNavigation/getStepStyles.ts
aaschlote: we can use this [tool](https://nekocalc.com/px-to-rem-converter)  to convert px to rem

PR #3759, opened 2026-05-28
DRAFT - RAST-516 erbschein flow

melgaafary:
This includes the new engine. It will be separated before this PR is done


app/routes/action.delete-array-item.ts
```diff
@@ -27,7 +28,12 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   const cookieHeader = request.headers.get("Cookie");
   const flowSession = await getSession(cookieHeader);
 
-  const resultDeletion = deleteArrayItem(arrayName, index, flowSession);
+  const resultDeletion = deleteArrayItem(
+    arrayName,
+    index,
+    flowSession,
+    parsePathname(pathname).arrayIndexes,
```
aaschlote: we can get the `arrayIndexes` in the function `getArrayDataFromFormData` as there it's called the `parsePathname`

app/services/session.server/arrayDeletion.ts
```diff
@@ -37,14 +37,24 @@ export function getArrayDataFromFormData(formData: FormData): Result<
...
+    .map((segment, i) =>
+      indices[i] === undefined ? segment : `${segment}[${indices[i]}]`,
+    )
+    .join(".");
+
 export const deleteArrayItem = (
   arrayName: string,
   index: number,
   flowSession: Session,
+  arrayIndexes: number[] = [],
```
aaschlote, line 59: with this new parameter, the `index` parameter is not more needed, as it should exist this value in the `arrayIndexes`. What about use this parameter and remove the `index`?

aaschlote: actually never mind, the `delete` array comes from the `array overview page`, it won't have all the time the `arrayIndexes`

PR #3763, opened 2026-05-29
fix(KERN): array summary responsiveness and hcm

ingriddorio:
This PR add forced-colors-auto to icons and change spacing on mobile in 200%

Screenshot 2026-05-29 at 14 04 09


app/components/content/arraySummary/ArraySummary.tsx
```diff
@@ -58,7 +58,12 @@ const ArraySummary = ({
       <div>
         <Button
           look="secondary"
-          iconLeft={<Icon name="plus" className="text-kern-action-default" />}
+          iconLeft={
+            <Icon
+              name="plus"
+              className="text-kern-action-default forced-color-adjust-auto!"
```
aaschlote, line 64: we are adding the selector `forced-color-adjust-auto` more than 10 times for the `Icon` for every single call, should we set it directly on the component?

aaschlote (approved):
LGTM!

Minor thing, the focus on Bearbeitung button add the focus ring in the whole line and the Delete only in a short space.

Image

Image

app/components/content/arraySummary/ArraySummary.tsx
ingriddorio: Yes, I can do that in a following PR

ingriddorio:
> LGTM!
> 
> Minor thing, the focus on Bearbeitung button add the focus ring in the whole line and the Delete only in a short space.
> 
> Image Image

It is probably happening because the first is a link and the second a button. I will create a ticket to improve this.

PR #3776, opened 2026-06-01
fix(KERN): add forced-color-adjust-auto to icon component

ingriddorio:
This PR add forced-color-adjust-auto to icon component and remove from icon usages


aaschlote (approved):
nice clean up! LGTM!.

Image

melgaafary (approved):

PR #3812, opened 2026-06-08
fix(KERN): add word-break to rich text

ingriddorio:
This PR add word-break to rich text to fix spacing in mobile 200% and fix SR announcing the textbeispiele

**Before**

Screenshot 2026-06-08 at 15 09 29

**After**

Screenshot 2026-06-08 at 15 08 59


app/components/content/Details.tsx
```diff
@@ -8,6 +8,10 @@ export type DetailsProps = {
   content?: string;
 };
 
+function stripHtml(content: string): string {
```
melgaafary: looks like a nice function for utils?

melgaafary (approved):
LGTM , small nitpick. You can ignore

app/components/content/Details.tsx
ingriddorio: True!

PR #3820, opened 2026-06-09
Improve vorabcheck loader

aaschlote:


app/routes/shared/components/VorabcheckPage.tsx
```diff
@@ -42,7 +42,7 @@ export function VorabcheckPage() {
           row={2}
           id="flow-page-content"
         >
-          <ContentComponents content={cmsContent.pre_form} managedByParent />
+          <ContentComponents content={cmsContent.content} managedByParent />
```
Spencer6497, line 45: are we renaming this in strapi too? I've always found `pre_form` to be a bit confusing

app/services/cms/index.server.ts
```diff
@@ -101,12 +101,10 @@ export async function fetchMultipleTranslations(scopes: string[]) {
   const translationsParsed = strapiSchemas.translations.safeParse(strapiEntry);
 
   if (translationsParsed?.data?.length === 0) {
-    const error = new Error(
-      `CMS lookup for translations failed (filters: ${JSON.stringify(scopes)})`,
-    );
-    error.name = "StrapiPageNotFound";
```
Spencer6497, line 107: why did we get rid of this?

Spencer6497 (approved):
great work!

app/routes/shared/components/VorabcheckPage.tsx
aaschlote: for now no, only in the type that it goes to front-end. To change in Strapi we need to do a migration and a few workaround in your application (due the schema vs content file)

app/services/cms/index.server.ts
aaschlote: I want to reuse the function `retrieveContentData` for the `voracheck`, because to get the content data between the `formular` and `voracheck` are the same, besides fetching the translations (voracheck doesn't need translation at all). As the function `fetchMultipleTranslations` has been only used in `retrieveContentData` and there the translations are not required all the time (of course we need to have it, but we don't need to thrown an exception when it doesn't have it), I decided to get rid of this. Any concerns?

PR #3841, opened 2026-06-17
Migrate TGA to new flow engine - First release

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-177


app/services/flow/userFlowAction/generateUserDataToSave.ts
```diff
@@ -0,0 +1,48 @@
...
+
+type UpcomingUserData = {
+  sessionUserData: SessionData;
+  formUserData: UserData;
+  migrationData: UserData | undefined;
+};
+
+export const generateUserDataToSave = <C extends PageConfigMap>(
+  stepId: string,
+  upcomingUserData: UpcomingUserData,
```
melgaafary, line 19: what does upcoming in this context mean?

app/services/navigation/buildStepStatesFromStatusTree.ts
```diff
@@ -0,0 +1,44 @@
...
+  validFlowPaths: string[],
+): StepState[] {
+  const toStepStates = (
+    nodes: Record<string, StatusNode>,
+    parentPath = "",
+  ): StepState[] => {
+    return Object.entries(nodes).map(([pathSegment, node]) => {
+      const fullPath = `${parentPath}${pathSegment}`;
+
+      const nodeChildrenWithoutResultPage = node.children
```
melgaafary, line 17: I think we should do this in the engine? it shouldnt return result pages, what do you think?

melgaafary (approved):

app/services/navigation/buildStepStatesFromStatusTree.ts
aaschlote: I think the result page is very business case for the A2J flow pages, not sure if it does make sense to add in the engine, specially if we are planning to use the engine as dependency in the future.

app/services/flow/userFlowAction/generateUserDataToSave.ts
aaschlote: the upcoming data from the user input and migration data

PR #3853, opened 2026-06-19
refactor(FGR & TGA): Split strasseHausnummer field for two discrete fields

Spencer6497:
**NOTE**: We still need to update the relevant Strapi pages, and get approval from Product that we can go ahead with the change _without_ a feature flag.
- **refactor(FGR): Swap strasseHausnummer for two discrete fields**
- **refactor(Geld Einklagen): Swap combined strasseHausnummer for discrete street and house number fields**


aaschlote:
I would recommend (already mentioned in slack to have two tickets) to have one PR for every flow, it's too many changes and it's a bit risk since these flows are being used in production.

app/domains/fluggastrechte/formular/flugdaten/pages.ts
```diff
@@ -27,7 +28,8 @@ export const fluggastrechteFlugdatenPages = {
     stepId: "flugdaten/adresse-fluggesellschaft",
     pageSchema: {
       fluggesellschaft: schemaOrEmptyString(z.string()),
-      fluggesellschaftStrasseHausnummer: stringRequiredSchema,
+      fluggesellschaftStrasse: stringRequiredSchema,
+      fluggesellschaftHausnummer: germanHouseNumberSchema,
```
aaschlote: As we are asking for the country, we shouldn't add the schema for the street number for german houses. Better to allow the user whatever he/she wishes.

app/domains/fluggastrechte/formular/persoenlicheDaten/pages.ts
```diff
@@ -14,7 +15,8 @@ const persoenlicheDatenSchema = {
   title: z.enum(["none", "dr"]),
   vorname: stringRequiredSchema,
   nachname: stringRequiredSchema,
-  strasseHausnummer: stringRequiredSchema,
+  strasse: stringRequiredSchema,
+  hausnummer: germanHouseNumberSchema,
```
aaschlote: same here, allow the user enter whatever she/he wishes.

Spencer6497:
NOTE: E2E tests will pass again when the content has been adapted -- currently failing because of the Klagende Person fieldset having a different first field: in the content it's klagendePersonStrasseHausnummer, but should be klagendePersonStrasse.

aaschlote:
> NOTE: E2E tests will pass again when the content has been adapted -- currently failing because of the Klagende Person fieldset having a different first field: in the content it's klagendePersonStrasseHausnummer, but should be klagendePersonStrasse.

You can add this temporary snippet code in the `renderFieldSet` to make the E2E Test green again:

```if (
    formComponents
      .filter(({ name }) => name !== "klagendePersonStrasseHausnummer")
      .findIndex(({ name }) => name === fieldName) > 0
  ) {
    return null;
  }```

PR #3863, opened 2026-06-22
APL-177 Make new engine isDone schema-based and fix sidenav navigation

melgaafary:
### Problem
The geldEinklagen sidenav was showing incorrect state: sections marked as done before the user completed them, and sections showing as reachable when they shouldn't be.

### Root Causes & Fixes
**1. BFS used all-branches evaluation — simulate.ts**
Problem: Pass 2 (BFS reachability) used evaluateAllBranches, marking every passing guard's target as reachable. When forderung = "etwasAnderes", both the error branch AND the loose objectKeysNonEmpty fallback guard passed — so sachgebietInfo appeared reachable even on the error path.

Fix: Switched BFS to first-match-wins via evaluateRoute, matching Pass 1's behavior.

**2. isDone was position-based — statusTree.ts, createFlowSession.ts, compileFlow.ts**
Problem: isDone was computed by checking whether the linear walk had exited a section ("did the walk enter and leave?"). Since flows have unconditional fallback transitions between sections, the walk projected through the entire flow with any partial data. Marking every visited-and-exited section as done.

Fix: isDone is now schema-based (doneNodeKeys). After simulation, createFlowSession validates each visited page's schema fields against user data. A section is isDone only when every reachable page in it has its required fields filled.

**3. Section-boundary transitions were unconditional — geldEinklagen flow configs**
Problem: Transitions like klagendePersonKaufmann: "beklagtePersonGegenWen" let the simulation walk and BFS cross into the next section regardless of whether the current section was complete.

Fix: Added isSachgebietDone, isKlagendePersonDone, isBeklagtePersonDone, isGerichtSuchenDone, isKlageErstellenBeklagtePersonDone, isProzessfuehrungDone guards on all section-boundary transitions. These read pageData.subflowDoneStates, which is recomputed and persisted on every form submit. The guards live in a shared subflowDoneGuards.ts.

These two fixes (schema-based isDone + section-boundary guards) work together: the guards create a circular dependency (guard reads subflowDoneStates → computed from isDone → computed from walk position → needs guard to pass), which the schema-based isDone breaks by computing completion directly from field data independent of walk position.

**4. Sidenav links used section-prefix URLs — getSessionAndEngine.ts**
Problem: The statusTree generates section-level URL entries like /gericht-pruefen/forderung. No page has that exact stepId, so createFlowSession threw and the user was redirected to the start page when clicking any sidenav link.

Fix: When createFlowSession throws for an unrecognized path, a fallback runs a temporary simulation from the initial page and finds the first page in the linear walk whose path starts with the given section prefix, then redirects there.


melgaafary:
There is an improvement to this that I havent done yet. we can make the done automatic inside the engine, it will be generic based on the schema

### Engine auto-gates section-boundary transitions

The engine knows every node's section prefix (from its stepId). When evaluating a transition from node A to node B, if their section prefixes differ, the engine automatically checks whether the source section is done before following.

Flow configs go back to simple unconditional transitions (sachgebietMietePachtRaum: "klagendePersonFuerWen") — no guards needed.

app/services/flow/userDataAndFlow/getSessionAndEngine.ts
```diff
@@ -35,13 +59,20 @@ export const getSessionAndEngine = async (
...
       fullUserData as Parameters<typeof createFlowSession>[1],
       stepId,
     );
     return Result.ok({ flowSession, flowSessionEngine });
   } catch {
-    // In case the engine throws an error during creation, we want to catch it and redirect the user to the initial page of the flow, instead of showing an error page.
-    return Result.err({ redirectTo: flowId + newEngineConfig.initialPath });
+    // stepId may be a section-prefix URL (e.g. "/gericht-pruefen/forderung")
+    // with no corresponding page. Find the first matching page and redirect.
+    const firstPageInSection = findFirstPageInSection(
```
aaschlote, line 70: Let's discuss this function, as far I remember in case the page is not available, we should always redirect to the initial page of the flow.

PR #3864, opened 2026-06-22
fix(KERN): SR announces Textbeispiele

ingriddorio:
This PR change aria-describedby and connect id to annnounce textbeispiele when the focus is on the input


app/components/content/Details.tsx
```diff
@@ -37,14 +45,15 @@ export const Details = ({ title, content }: DetailsProps) => {
...
         </summary>
         <div
-          id="content"
-          aria-labelledby="summary"
+          id={contentId}
+          aria-labelledby={summaryId}
           className="pl-kern-space-x-large pt-kern-space-small text-kern-layout-text-default"
         >
+          <span id={contentId} className="sr-only">
+            Textbeispiele
```
aaschlote: can we move it to the translation files?

app/components/content/Details.tsx
```diff
@@ -1,16 +1,24 @@
...
+import { useId, useState, useEffect } from "react";
 import { GridItem } from "../layout/grid/GridItem";
 import { Icon } from "../common/Icon";
 import RichText from "../common/RichText";
-import { stripHtml } from "~/util/stripHtml";
 
 export type DetailsProps = {
   title?: string;
   content?: string;
+  onContentId?: (id: string) => void;
```
aaschlote: A good pattern for naming functions is to start with a verb that describes the action. `onContentId` doesn't convey its purpose clearly. Since it updates the parent component's state, consider renaming it to `setDetailsId`

app/components/content/Details.tsx
```diff
@@ -1,16 +1,24 @@
-import { useState } from "react";
+import { useId, useState, useEffect } from "react";
 import { GridItem } from "../layout/grid/GridItem";
 import { Icon } from "../common/Icon";
 import RichText from "../common/RichText";
-import { stripHtml } from "~/util/stripHtml";
 
 export type DetailsProps = {
   title?: string;
```
aaschlote: Can you check the possibility to make these props (title and content) requires? They shouldn't be optional.

app/components/content/Details.tsx
```diff
@@ -37,14 +45,15 @@ export const Details = ({ title, content }: DetailsProps) => {
...
+          aria-labelledby={summaryId}
           className="pl-kern-space-x-large pt-kern-space-small text-kern-layout-text-default"
         >
+          <span id={contentId} className="sr-only">
+            Textbeispiele
+          </span>
+
           {content && <RichText className="leading-[1.5]" html={content} />}
-          <div className="sr-only" aria-live="polite" aria-atomic="true">
-            {isOpen && content && `Textbeispiele: ${stripHtml(content)}`}
```
aaschlote, line 46: since you removed this function, we are getting a warning from knip as this function hasn't been used anywhere. I would suggest either remove or add a test so we don't see the knip warning.

Image

melgaafary (approved):

PR #3865, opened 2026-06-22
refactor(Overflow): Add global overflow/word-break instead of in specific places in the codebase

Spencer6497:
Noticed that we're adding a lot of "hyphens-auto" or overflow-break-word in a lot of different places, when we could just add a global style override. I can't really think of a scenario in which we *don't* want the default to be to break words. And where we want it to not wrap, we're usually overriding it anyways.

### DIFF
--- a/app/styles.kern.css
+++ b/app/styles.kern.css
@@ -48,6 +48,11 @@
 
   /* ------------------------------- */
 
+  body {
+    hyphens: auto;
+    overflow-wrap: break-word;
+  }
+
   .detail-summary:focus-within,
   .kern-tile:focus-within {
     border-radius: var(--kern-metric-border-radius-large, 0.25rem);
--- a/app/styles.kern.css
+++ b/app/styles.kern.css
@@ -49,8 +49,8 @@
   /* ------------------------------- */
 
   body {
-    hyphens: auto;
-    overflow-wrap: break-word;
+    @apply hyphens-auto;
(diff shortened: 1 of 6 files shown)


app/styles.kern.css
```diff
@@ -48,6 +48,11 @@
 
   /* ------------------------------- */
 
+  body {
+    hyphens: auto;
```
aaschlote: can we use tailwind directive `@apply` instead of writing css?

app/styles.kern.css
```diff
@@ -48,6 +48,11 @@
 
   /* ------------------------------- */
 
+  body {
```
aaschlote, line 51: Just wondering about the scope of this fix—how many files would we need to touch? I noticed that `hyphens-auto` and `wrap-break-word` are already used three times each in this repo. 

Applying this directly to the body selector sounds a bit rough.

Spencer6497: Quite a few more files as long as we keep discovering places in the app where word-break is breaking formatting, like [here](https://digitalservicebund.atlassian.net/jira/software/c/projects/APL/boards/1914?selectedIssue=APL-164) and [here](https://digitalservicebund.atlassian.net/jira/software/c/projects/APL/boards/1914?selectedIssue=APL-165). To avoid increasing the number of files affected by this issue, I think it would be much better to set a global default.

melgaafary (approved):

PR #3866, opened 2026-06-22
feat(Table): add table component

ingriddorio:
This PR add a new table component

Screenshot 2026-06-25 at 14 52 33

Screenshot 2026-06-25 at 14 53 08


app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+  isHeaderRow: z.boolean().optional(),
+  cells: z.array(StrapiTableCellSchema),
+});
+
+export const StrapiTableSchema = z.object({
+  __component: z.literal("page.table"),
+  id: z.string().optional(),
+  heading: z.string().optional(),
+  description: z.string().optional(),
+  columns: z.array(StrapiTableColumnSchema).default([]),
```
aaschlote: can we have empty columns in the table?

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+  rows: z.array(StrapiTableRowSchema).default([]),
```
aaschlote: can we have empty rows in the table?

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
+import { z } from "zod";
+
+const StrapiTableColumnSchema = z.object({
+  id: z.string().optional(),
```
aaschlote: you can use the schema `HasStrapiIdSchema`.

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+  label: z.string(),
+});
+
+const StrapiTableCellSchema = z.object({
+  id: z.string().optional(),
```
aaschlote: you can use the schema `HasStrapiIdSchema`.

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+  content: z.string(),
+  isHeader: z.boolean().optional(),
+});
+
+const StrapiTableRowSchema = z.object({
+  id: z.string().optional(),
```
aaschlote: you can use the schema `HasStrapiIdSchema`.

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const StrapiTableColumnSchema = z.object({
+  id: z.string().optional(),
+  label: z.string(),
+});
+
+const StrapiTableCellSchema = z.object({
+  id: z.string().optional(),
+  content: z.string(),
+  isHeader: z.boolean().optional(),
```
aaschlote: why this value is optional? When the value can be undefined if the component doesn't exist?

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+});
+
+const StrapiTableRowSchema = z.object({
+  id: z.string().optional(),
+  isHeaderRow: z.boolean().optional(),
```
aaschlote: why this value is optional? When the value can be undefined if the component doesn't exist?

app/components/content/Table.tsx
```diff
@@ -0,0 +1,50 @@
+import type z from "zod";
+import { type StrapiTableSchema } from "~/services/cms/models/content/StrapiTable";
+
+type TableProps = z.infer<typeof StrapiTableSchema>;
+
+const Table = ({ heading, description, columns, rows }: TableProps) => {
+  return (
+    <div>
+      {heading && <h2 className="kern-title">{heading}</h2>}
```
aaschlote: we shouldn't hardcode `h2` here, because we don't know what heading can be rendered on the page. Better to use `StrapiHeadingSchema` schema and render this section with `Heading`.

We had this issue with the `ArraySummary` component once.

app/components/content/Table.tsx
```diff
@@ -0,0 +1,50 @@
+import type z from "zod";
+import { type StrapiTableSchema } from "~/services/cms/models/content/StrapiTable";
+
+type TableProps = z.infer<typeof StrapiTableSchema>;
```
aaschlote: We shouldn't use the Strapi schema directly for component props here. Ideally, components should be independent of the CMS, which just provides the content. Much of the data from Strapi, like __component and sometimes id, isn't relevant to the component itself. The Strapi schema should only be used to validate external data; the component should be CMS-agnostic and render regardless of where the data comes from. 

Let's define a dedicated Props type that follows the current pattern in the repo.

app/components/content/Table.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const Table = ({ heading, description, columns, rows }: TableProps) => {
+  return (
+    <div>
+      {heading && <h2 className="kern-title">{heading}</h2>}
+      {description && <p className="kern-text">{description}</p>}
+
+      <table className="kern-table">
+        {heading && <caption className="kern-title">{heading}</caption>}
```
aaschlote: Checking the [Figma file](https://www.figma.com/design/Iu857Nsm8ZIhAjwFatKvZN/A2J-%7C-%E2%9D%96-Komponenten-2.0?node-id=17209-669&p=f&t=24LMQmV62GqHqtB8-0), the content of this heading it's different from above, so it shouldn't use the same data. I believe we need to create a new field here

app/components/content/Table.tsx
```diff
@@ -0,0 +1,50 @@
+import type z from "zod";
+import { type StrapiTableSchema } from "~/services/cms/models/content/StrapiTable";
+
+type TableProps = z.infer<typeof StrapiTableSchema>;
+
+const Table = ({ heading, description, columns, rows }: TableProps) => {
+  return (
+    <div>
+      {heading && <h2 className="kern-title">{heading}</h2>}
+      {description && <p className="kern-text">{description}</p>}
```
aaschlote: I believe we should use a markdown (RichtText) to render this paragraph, as the content creator might want to use bold, italic and configuration here.

app/services/cms/models/content/__test__/StrapiTable.test.ts
```diff
@@ -0,0 +1,49 @@
+import { describe, it, expect } from "vitest";
+import { StrapiTableSchema } from "../StrapiTable";
+
+describe("StrapiTableSchema", () => {
```
aaschlote, line 4: nice tests, but I would like to see some tests that it fails in the validation for the required fields.

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,27 @@
...
+  cells: z.array(StrapiTableCellSchema),
+});
+
+export const StrapiTableSchema = z.object({
+  __component: z.literal("page.table"),
+  id: z.string().optional(),
```
aaschlote: do we need this id to render the component? If not, we don't need to add it to the schema as this information is not relevant.

aaschlote (commented):
I think we should take another look at the data structure here. Right now, columns and rows are optional, but as we know, it feels a bit weird to render a table without them! 

Also, it would be awesome to see this component in Storybook with all its different variations, if possible.

app/components/content/Table.tsx
```diff
@@ -0,0 +1,50 @@
...
+              ))}
+            </tr>
+          </thead>
+        )}
+
+        <tbody className="kern-table__body">
+          {rows.map((row) => (
+            <tr key={row.id} className="kern-table__row">
+              {row.cells.map((cell, cellIndex) => {
+                const isHeader = cell.isHeader || cellIndex === 0;
```
m0dh4x: Hey, I think here if the isHeader is explicit set to `false` it would be ignored even the cellIndex is 0. so that means it is impossible to opt out the first cell of row-header, right? think here is better to use  nullish coalescing.

app/components/content/Table.tsx
m0dh4x: yes, actually if `heading` is provided the content will be rendered twice. :)

app/services/cms/models/content/StrapiTable.ts
m0dh4x: Hey, this value isn't used in the table.tsx? Could it removed or is there a behavior needed to implement?

app/components/content/__test__/Table.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+    );
+
+    expect(screen.getByText("Column 1")).toBeInTheDocument();
+    expect(screen.getByText("Column 2")).toBeInTheDocument();
+  });
+
+  it("should not render thead when columns are empty", () => {
+    render(<Table columns={[]} rows={[]} __component={"page.table"} />);
+
+    expect(screen.queryByRole("thead")).toBeNull();
```
m0dh4x: Huhu, afaik `thead` isn't an ARIA role so this returns always `null`.

app/components/content/__test__/Table.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+              },
+            ],
+          },
+        ]}
+        __component={"page.table"}
+      />,
+    );
+
+    const headerCell = screen.getByText("Header");
+    expect(headerCell.tagName).toBe("TH");
```
m0dh4x: here is only the `th` check, but i would also check that cell 0 is rendering as `td`. think thats the reason why the bug in `false || cellIndex` is undetected. :)

app/services/cms/models/content/StrapiTable.ts
ingriddorio: Nope, fixed!

app/services/cms/models/content/StrapiTable.ts
ingriddorio: Nope, fixed!

app/components/content/Table.tsx
ingriddorio: Alright! Thanks for that, I did not know :)

ingriddorio:
> I think we should take another look at the data structure here. Right now, columns and rows are optional, but as we know, it feels a bit weird to render a table without them!
> 
> Also, it would be awesome to see this component in Storybook with all its different variations, if possible.

@aaschlote yes, it makes sense to rethink the data structure. I added the Storybook but for now we will have just the current variation of the table.

app/components/content/Table.tsx
```diff
@@ -0,0 +1,89 @@
...
+          tagName="h1"
+          className="kern-heading-large"
+          managedByParent
+        />
+      )}
+      {description && <RichText className="kern-text" html={description} />}
+
+      <table className="kern-table">
+        <caption className="kern-title kern-heading-medium flex justify-start">
+          {title}
```
aaschlote: if the `title` is optional, should we avoid to render an empty `caption`?

app/components/content/Table.tsx
```diff
@@ -0,0 +1,89 @@
...
+        </caption>
+        {columns.length > 0 && (
```
aaschlote, line 50: can the `columns` be without any data?

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,29 @@
...
+  ...HasStrapiIdSchema.shape,
+  cells: z.array(StrapiTableCellSchema),
+});
+
+export const StrapiTableSchema = z.object({
+  __component: z.literal("page.table"),
+  heading: StrapiHeadingSchema,
+  title: z.string().optional(),
+  description: z.string().optional(),
+  columns: z.array(StrapiTableColumnSchema),
```
aaschlote: if we want the `columns` to not have any empty array, we can use `.nonempty()`

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,29 @@
...
+  rows: z.array(StrapiTableRowSchema),
```
aaschlote: if we want the `rows` to not have any empty array, we can use `.nonempty()`

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const StrapiTableCellSchema = z.object({
+  ...HasStrapiIdSchema.shape,
+  header: z.string(),
+  content: z.string(),
+});
+
+const StrapiTableRowSchema = z.object({
+  ...HasStrapiIdSchema.shape,
+  cells: z.array(StrapiTableCellSchema),
```
aaschlote: if we want the `cells` to not have any empty array, we can use `.nonempty()`

app/components/content/Table.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <GridItem
+      mdColumn={{ start: 1, span: 9 }}
+      lgColumn={{ start: 3, span: 9 }}
+      xlColumn={{ start: 3, span: 9 }}
+      className="flex flex-col gap-kern-space-default"
+    >
+      {heading && (
+        <Heading
+          {...heading}
+          tagName="h1"
```
aaschlote: why are you setting to be `h1`? The tag name should come from Strapi and not defined by code, specially for `h1` element, as it shouldn't be more than one on the page.

app/components/content/Table.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const Table = ({ heading, title, description, columns, rows }: TableProps) => {
+  return (
+    <GridItem
+      mdColumn={{ start: 1, span: 9 }}
+      lgColumn={{ start: 3, span: 9 }}
+      xlColumn={{ start: 3, span: 9 }}
+      className="flex flex-col gap-kern-space-default"
+    >
+      {heading && (
```
aaschlote, line 38: if the field is required, I believe we don't need this condition, right?

aaschlote (commented):
nice work, I believe we need to check if we need to restrict the arrays to not be empty (`[]`), so we assure the component will always render with columns and rows.

app/components/content/Table.tsx
ingriddorio: True!

app/services/cms/models/content/StrapiTable.ts
```diff
@@ -0,0 +1,29 @@
...
+});
+
+const StrapiTableRowSchema = z.object({
+  ...HasStrapiIdSchema.shape,
+  cells: z.array(StrapiTableCellSchema).nonempty(),
+});
+
+export const StrapiTableSchema = z.object({
+  __component: z.literal("page.table"),
+  heading: StrapiHeadingSchema,
```
aaschlote: As this field is [optional](https://github.com/digitalservicebund/a2j-rechtsantragstelle-strapi/pull/671/changes#r3490906675), we can use the schema `StrapiHeadingOptionalSchema`

app/components/content/Table.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  {column.header}
+                </th>
+              ))}
+            </tr>
+          </thead>
+        )}
+
+        <tbody className="kern-table__body">
+          {rows.map((row) => (
+            <tr key={String(row.id)} className="kern-table__row">
```
aaschlote: Can you check again if we need `String(row.id)`, maybe `key={row.id}` enough, no?

aaschlote (approved):
I wrote more two comments, but it's good to go.

LGTM!

app/components/content/Table.tsx
ingriddorio: Yes, key={row.id} is enough :)

PR #3871, opened 2026-06-25
APL-213 add pageData to the engine

melgaafary:


app/services/flow/newFlowEngine/createFlowSession.ts
```diff
@@ -107,6 +107,7 @@ export const createFlowSession = <C extends PageConfigMap>(
       return compiledFlow.getArrayInfo(path) !== undefined;
     },
     isFinal: compiledFlow.isFinal(currentPath) ?? false,
+    pageData: userData.pageData,
```
aaschlote, line 110: I was expecting the `pageData` to be inserted inside the `prunedUserData` and not in a new parameter

PR #3873, opened 2026-06-26
Migrate TGA to new flow engine - Second release

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-177


app/domains/__test__/flowSchemaNewEngine.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  if (!userInput || addArrayItemEvent) {
+    expect(pageSchema).toBeUndefined();
+  } else {
+    expect(pageSchema).toBeDefined(); // With userInput we expect it to validate against the pageSchema
+    const validationResult = z.object(pageSchema).safeParse(userInput);
+    expect(validationResult.error).toBeUndefined();
+  }
+}
+
+function testXstateArrayLinkages(
```
melgaafary: should this be `testNewEngineArrayLinkages`? or simply `testArrayLinkages`

app/domains/__test__/flowSchemaNewEngine.test.ts
```diff
@@ -0,0 +1,248 @@
...
+) => {
+  return expectedSteps.slice(0, idx + 1).reduce((acc, step) => {
+    let merged: ExpectedStepUserInput<UserData> = { ...acc, ...step.userInput };
+
+    // This assignment is useful for injecting pageData, like subflowDoneStates
+    merged.pageData = { ...merged.pageData, ...step.pageData };
+
+    // When adding an array item, increment the last array index or start with 0
+    if (step.addArrayItemEvent) {
+      merged.pageData = {
```
melgaafary, line 43: wouldnt this override line 39 if this condition is true?

app/domains/__test__/flowSchemaNewEngine.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      console.warn(
+        `Total of ${totalMissingStepCount} untested stepIds: `,
+        Object.fromEntries(missingStepsEntries),
+      );
+    }
+  });
+
+  Object.entries(flowSchemaTests).forEach(
+    ([testConfigName, { xstateConfig, testcases, newEngineConfig }]) => {
+      //TODO: Add later the flowId in the Testcase type, so we don't depend of the xstateConfig.id here. This is a temporary solution to get the flowId for the testcases.
```
melgaafary, line 220: do you want to create a ticket for this not to forget? we don't want any coupling of xstate and the new engine

app/domains/__test__/flowSchemaNewEngine.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  return expectedSteps.slice(0, idx + 1).reduce((acc, step) => {
+    let merged: ExpectedStepUserInput<UserData> = { ...acc, ...step.userInput };
+
+    // This assignment is useful for injecting pageData, like subflowDoneStates
+    merged.pageData = { ...merged.pageData, ...step.pageData };
+
+    // When adding an array item, increment the last array index or start with 0
+    if (step.addArrayItemEvent) {
+      merged.pageData = {
+        arrayIndexes: merged.pageData?.arrayIndexes
```
melgaafary, line 45: also this will only work for normal arrays. We will have to change it a bit if we will use this for nested array flows too. Just a note for late tho

melgaafary (commented):
🚀 🚀  nice work!

Just a few comments

app/domains/__test__/flowSchemaNewEngine.test.ts
aaschlote: ops, true, I forgot to rename it.

app/domains/__test__/flowSchemaNewEngine.test.ts
aaschlote: true, as I don't have any nested array flow to test it, I can't reproduce it, but once we have this scenario to test, we need to modify it.

app/domains/__test__/flowSchemaNewEngine.test.ts
aaschlote: I can do it, just add a TODO to make it later

app/domains/__test__/flowSchemaNewEngine.test.ts
aaschlote: probably, but in TGA I don't have a scenario for array to test it. So I'll make a change, but I can't guarantee it will work.

PR #3925, opened 2026-07-08
APL-247 New engine cascade pruning 

melgaafary:
### Cascading pruning strategy for the new flow engine

**Problem**
Stale answers stay in the session when a user changes an earlier answer and the flow takes a different branch. Pruning removes them from prunedUserData, but only in a single pass, and navigation still reads the raw session data. So a guard that checks the presence of a field still sees the stale value and routes to the wrong page.

Example from geld-einklagen: a user goes through the Miete branch and enters a secondary PLZ, then switches to Urheberrecht. The stale secondary PLZ still makes getPilotCourts find a pilot court, so the user lands on the pilot court page instead of the Abbruch result page.

**Solution**
New per-flow option on compileFlow:

pruningStrategy: "singlePass" | "cascading"; // default: "singlePass"
With "cascading", createFlowSession:

Re-simulates on the pruned data and re-prunes until nothing changes (removals can cascade, and branches selected by the clean data keep their fields).
Derives navigation from the stable data (nextPath, prevPath, isReachable, statusTree, isDone), so the flow behaves as if the stale fields were never there.
Iterations are capped at the page count as a safety net against oscillating presence-guards.

Default ("singlePass") is the current behavior, no other flow is affected.
geld-einklagen opts in via its flow config.

### Test fixes
Three geld-einklagen test cases had incomplete fixture data — fields from pages that aren't reachable with the rest of the data (e.g. fuerWenKlagen without any sachgebiet, klagendeHaustuergeschaeft without klagendeVerbraucher: "yes"). A real user can't produce these states, and cascading pruning now correctly treats such fields as stale. The fixtures were completed with the missing answers; all expected steps are unchanged.


aaschlote (approved):
Amazing work. LGTM!

PR #3927, opened 2026-07-08
New Flow Engine: Migrate Kontopfändung Wegweiser Vorabcheck

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-249

### DIFF
--- a/app/domains/kontopfaendung/wegweiser/__test__/testcasesWithUserInputs.ts
+++ b/app/domains/kontopfaendung/wegweiser/__test__/testcasesWithUserInputs.ts
@@ -1,9 +1,11 @@
 import type { FlowTestConfig } from "~/domains/__test__/TestCases";
 import { kontopfaendungWegweiserXstateConfig } from "../xStateConfig";
 import { type KontopfaendungWegweiserUserData } from "~/domains/kontopfaendung/wegweiser/userData";
+import { kontopfaendungWegweiserFlowConfig } from "../flowConfig";
 
 export const kontopfaendungWegweiserTestCases = {
   xstateConfig: kontopfaendungWegweiserXstateConfig,
+  newEngineConfig: kontopfaendungWegweiserFlowConfig,
   testcases: {
     yesKontopfaendung: [
       { stepId: "/start" },
@@ -38,7 +40,14 @@ export const kontopfaendungWegweiserTestCases = {
       { stepId: "/p-konto-probleme" },
     ],
     kinder: [
-      { stepId: "/kinder", userInput: { hasKinder: "yes" } },
+      {
+        stepId: "/kinder",
+        userInput: {
+          hasKinder: "yes",
+          hasPKonto: "ja",
+          hasKontopfaendung: "ja",
+        },
+      },
       {
         stepId: "/kinder-wohnen-zusammen",
         userInput: { kinderWohnenZusammen: "nein" },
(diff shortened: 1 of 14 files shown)


app/domains/kontopfaendung/wegweiser/flowConfig.ts
```diff
@@ -0,0 +1,193 @@
...
+    },
+  ]),
+) as {
+  [K in keyof typeof kontopfaendungWegweiserPages]: Omit<
+    (typeof kontopfaendungWegweiserPages)[K],
+    "stepId"
+  > & {
+    stepId: string;
+  };
+};
```
Spencer6497: It looks like this leading slash logic is repeated in a few places; maybe we could extract it to a reusable function?

app/domains/kontopfaendung/wegweiser/__test__/testcasesWithUserInputs.ts
```diff
@@ -52,7 +61,12 @@ export const kontopfaendungWegweiserTestCases = {
     partner: [
       {
         stepId: "/partner",
-        userInput: { verheiratet: "ja" },
+        userInput: {
+          verheiratet: "ja",
+          hasKinder: "yes",
+          hasPKonto: "ja",
+          hasKontopfaendung: "ja",
```
Spencer6497, line 68: I don't think I understand: why does it need all this additional context?

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
```diff
@@ -71,7 +64,7 @@ export const getNachzahlungSozialMehr500Strings = (
   return {
     hasNachzahlungSozialMehr500:
       userData.hasSozialleistungNachzahlung === "yes" &&
-      userData.sozialleistungNachzahlungHigherThan === "yes",
+      userData.arbeitgeberNachzahlungHigherThan === "yes",
```
Spencer6497, line 67: I know you didn't name this, but can we pleaaaaase rename this 😭

app/routes/shared/newEngineVorabcheck.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (resultValidatedSession.isErr) {
+    logWarning(resultValidatedSession.error);
+    throw new Response(null, { status: 403 });
+  }
+
+  const { pathname } = url;
+  const { flowId, currentFlow, stepId } =
+    getPageAndFlowDataFromPathname(pathname);
+
+  const compiledStaticFlow =
```
Spencer6497, line 95: why is it called "compiled"? What exactly are we compiling?

Spencer6497 (approved):
exciting stuff!! 🚀

app/domains/kontopfaendung/wegweiser/flowConfig.ts
aaschlote: yeah, I noticed it, I'll create a function. But this solution is only temporary, we should for new flows always add `/`, this solution is keep the new engine works together with the old implementation with the `xState`.

app/domains/kontopfaendung/wegweiser/__test__/testcasesWithUserInputs.ts
aaschlote: Due the many restrictions we have done to make the `back button` works, we need to have all the user data so the engine knows where it should come back.

app/domains/kontopfaendung/wegweiser/stringReplacements.ts
aaschlote: Maybe we can do this later, I also have seen a few fields with the values `ja` and `nein`. But this would require a change also in Strapi. I'll open a ticket for that.

app/routes/shared/newEngineVorabcheck.ts
aaschlote: When you create the static variable for the creation of the flowConfig.ts it's used the function `compileFlow`, I believe Chris mentioned that it would like to rename it. I'll have a meeting with him today and ask about it.

PR #3928, opened 2026-07-08
fix(acccessibility): make SR announce button and accordion status

ingriddorio:
This PR add aria label and aria expanded to the button so it announces the status "dimmed" and if the accordion is collapsed or expanded


app/components/content/autoGeneratedSummary/AutoGeneratedSummary.tsx
```diff
@@ -70,16 +70,20 @@ const AutoGeneratedSummary = ({
           <Button
             look="secondary"
             onClick={expandAll}
+            text={expandAllText}
             aria-disabled={allOpen}
+            aria-expanded={allOpen}
             className={allOpen ? "kern-btn--disabled pointer-events-none" : ""}
-            text={expandAllText}
+            aria-label={`${expandAllText}${allOpen ? ", Alle Abschnitte sind bereits aufgeklappt" : ""}`}
```
aaschlote: can we move this text to translation?

app/components/content/autoGeneratedSummary/AutoGeneratedSummary.tsx
```diff
@@ -70,16 +70,20 @@ const AutoGeneratedSummary = ({
           <Button
             look="secondary"
             onClick={expandAll}
+            text={expandAllText}
```
aaschlote: The props `expandAllText` and `collapseAllText` are with default values and they are just text, to improve this component I would recommend to move them also to translations

aaschlote (commented):
Can you check why it was created the props `showGlobalControls`? I don't see any reason why this props exist and it has a default value as well.

I would like also to see the new changes in a test if it's possible.

app/components/content/autoGeneratedSummary/AutoGeneratedSummary.tsx
ingriddorio: Done in a new PR

app/components/content/autoGeneratedSummary/AutoGeneratedSummary.tsx
ingriddorio: Done in a new PR

ingriddorio:
> Can you check why it was created the props `showGlobalControls`? I don't see any reason why this props exist and it has a default value as well.
> 
> I would like also to see the new changes in a test if it's possible.

@aaschlote 

About the 'showGlobalControls' I agree it is nice to check but it can be done in a new ticket/PR in another moment. This is out of the scope of the accessibility changes I did here and it is not breaking anything.

The default values you mentioned was moved into the translations file in another PR.

I added some accessibility tests 👍🏽

app/components/content/autoGeneratedSummary/__test__/AutoGeneratedSummary.test.tsx
```diff
@@ -102,3 +104,139 @@ describe("AutoGeneratedSummary", () => {
     expect(details[1]).toHaveAttribute("open");
   });
 });
+
+describe("AutoGeneratedSummary accessibility", () => {
+  beforeEach(() => {
```
aaschlote: you can call this `beforeEach` out of the describe and the another describe too and you won't need to have twice `beforeEach` in this file

app/components/content/autoGeneratedSummary/__test__/AutoGeneratedSummary.test.tsx
```diff
@@ -102,3 +104,139 @@ describe("AutoGeneratedSummary", () => {
...
+    const details = container.querySelectorAll("details");
+    expect(details[0]).toHaveAttribute("open");
+    expect(details[1]).toHaveAttribute("open");
+    expect(expandButton).toHaveAttribute(
+      "aria-label",
+      `${translations.autogeneratedSummary.expandAllText.de}, ${translations.autogeneratedSummary.expandAllAriaLabelText.de}`,
+    );
+  });
+
+  it("should not call expandAll when aria-disabled", async () => {
```
aaschlote: what this does it? I didn't get from this description

aaschlote (approved):
besides my comments and the warnings from Sonar (check below), LGTM!

Image

app/components/content/autoGeneratedSummary/__test__/AutoGeneratedSummary.test.tsx
ingriddorio: I changed the description. I hope it is clear now

PR #3933, opened 2026-07-09
Fix Bug: Apl 250 - user unable to continue flow related to renderHiddenField

m0dh4x:
## Summary

Fixes APL-250: users reported being unable to continue the flow written down [here](https://digitalservicebund.atlassian.net/browse/APL-250?atlOrigin=eyJpIjoiMTFhZjNiZTMyMTI1NGM3Yzg1ZTk0Mzk4ZjBhYzJlNWUiLCJwIjoiaiJ9)

## Fix

- Removed the `renderHiddenField` state/`onClick` workaround from `RadioGroup` and `Tile`. 

## How to reproduce (before this fix)

1. Open an unanswered radio or tile-group question in a fresh session.
2. In the browser console, check an option without dispatching a click, e.g.: `document.querySelector('input[type="radio"]').checked = true`


aaschlote (changes requested):
It looks like the issue on rendering and click on item is gone, however now we have another issue when the user has JS, forgets to select an option (Radio, Tile or Checkbox) and when clicks again the error message is not gone!

Image

aaschlote (changes requested):
It looks like we still have the original error on the Checkbox component, can you please take a look?

Image

m0dh4x:
@aaschlote @melgaafary seems I fixed also the checkboxes. Could you please double check? :)

aaschlote (changes requested):
@m0dh4x The problem is still happening, check it out below.

Image

aaschlote (approved):
Working. LGTM!

PR #3968, opened 2026-07-17
feat(Card): add component

ingriddorio:
This PR add the Card component for the use in the Homepage Redesign

Screenshot 2026-07-17 at 15 50 38

Screenshot 2026-07-17 at 15 50 24


melgaafary (approved):

app/components/content/card/Card.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <div className="kern-card__container bg-kern-neutral-025!">
+          <header className="kern-card__header">
+            <hgroup className="kern-hgroup">
+              <p
+                className="kern-label text-kern-layout-text-muted! font-normal!"
+                id={id}
+              >
+                {heading}
+              </p>
+              <Heading
```
aaschlote, line 31: I would prefer to not hardcode as `h2`, but given the content creator which `h` element should render on the page, specially we don't know the parent or children will render on the page.

app/components/content/card/Card.tsx
```diff
@@ -0,0 +1,51 @@
...
+                text={title}
+                managedByParent
+              />
+            </hgroup>
+          </header>
+          <section className="kern-card__body">
+            <p className="kern-body">{description}</p>
+          </section>
+          <footer className="kern-card__footer pt-kern-space-x-large!">
+            <Button aria-describedby={id} text={buttonLabel} look="secondary" />
```
aaschlote, line 43: should we give the option for the content creator decide what look the Button will render on the page or should be always  "secondary".

app/services/cms/models/content/StrapiCard.ts
```diff
@@ -0,0 +1,12 @@
+import z from "zod";
+import { HasStrapiIdSchema } from "../HasStrapiId";
+import { StrapiHeadingSchema } from "./StrapiHeading";
+
+export const StrapiCardSchema = z.object({
+  __component: z.literal("page.card"),
+  heading: StrapiHeadingSchema,
```
aaschlote, line 7: can we rename a better name of `heading` and `title`. It looks like the both props are the same. Maybe rename one to `heading` and another one to `subHeading` or `subLine`.

app/components/content/ContentComponents.tsx
```diff
@@ -128,6 +130,24 @@ function cmsToReact(
...
+            id: String(card.id),
+            heading: card.heading?.text ?? "",
+          }))}
+        />
+      );
+    case "page.card":
+      return (
+        <Card
+          {...componentProps}
+          id={String(componentProps.id)}
```
aaschlote, line 147: why do we need this line if the `id` is on the schema?

app/components/content/ContentComponents.tsx
```diff
@@ -128,6 +130,24 @@ function cmsToReact(
       );
     case "page.table":
       return <Table {...componentProps} />;
+    case "page.card-group":
+      return (
+        <CardGroup
+          cards={componentProps.cards.map((card) => ({
+            ...card,
+            id: String(card.id),
```
aaschlote, line 138: why do we need this line if the `id` is on the schema?

aaschlote (commented):
I have seen you already merged, but can you check my comments later?

PR #3971, opened 2026-07-20
Migrate schemas to auto suggestion schema

aaschlote:


app/domains/fluggastrechte/formular/flugdaten/pages.ts
```diff
@@ -26,7 +27,7 @@ export const fluggastrechteFlugdatenPages = {
   flugdatenAdresseFluggesellschaft: {
     stepId: "flugdaten/adresse-fluggesellschaft",
     pageSchema: {
-      fluggesellschaft: schemaOrEmptyString(z.string()),
+      fluggesellschaft: autoSuggestSchema(stringOptionalSchema)("airlines"),
```
Spencer6497, line 30: doesn't `schemaOrEmptyString` behave differently from `stringOptionalSchema`?

Spencer6497 (approved):
great work 🤝

app/domains/fluggastrechte/formular/flugdaten/pages.ts
aaschlote: Yeah, but if we use `autoSuggestSchema(schemaOrEmptyString(z.string()))("airlines")`, I lose access to the schema's meta. Also, since the `fluggesellschaft` field is only used for display, I'll chat with Kristin about replacing it with a FieldSet component so we can drop the field altogether.

PR #3985, opened 2026-07-21
feat(Homepage): move homepage to code

ingriddorio:
This PR moves the Homepage from Strapi to code


app/routes/shared/components/Homepage.tsx
```diff
@@ -0,0 +1,252 @@
...
+      title: translations.homepage.cardTitleAnleitung.de,
+      description: translations.homepage.cardDescriptionAnleitung.de,
+      buttonLabel: translations.homepage.cardButtonLabelAnleitung.de,
+      span: 6,
+    },
+  ];
+
+  return (
+    <>
+      <GridSection className="bg-kern-action-default" pt="40" pb="40">
```
aaschlote, line 16: Instead of call the GridSection and the following components, maybe you can just create an object and call `ContentComponents`.

It would be:

```
topContents: [
{
__component: "page.hero",
heading: {
text: translations.homepage.heroHeading.de,
....
},

...
]>

<ContentComponents contents={topContents} />
```

And do for the others too, it should reduce the size of this file and keep the logic of rendering the component on ContentComponents.

ingriddorio: I was trying to do in the way you suggested but it can not be done because of the centering of the page :(
I need to add the GridItem and hardcode the styling so it looks as expected.

app/routes/shared/components/Homepage.tsx
```diff
@@ -0,0 +1,262 @@
...
+import Box from "~/components/content/Box";
+import Hero from "~/components/content/Hero";
+import BMJVLogo from "~/assets/BMJV_Logo.svg?raw";
+
+const Homepage = () => {
+  const serviceCards = [
+    {
+      span: 4,
+      link: "/geld-einklagen",
+      id: translations.homepage.cardIdTGA.de,
```
aaschlote: I would prefer not having the `id` in the translation file, maybe add any random value for this property.

app/routes/shared/components/Homepage.tsx
```diff
@@ -0,0 +1,262 @@
...
+import { GridItem } from "~/components/layout/grid/GridItem";
+import { GridSection } from "~/components/layout/grid/GridSection";
+import { translations } from "~/services/translations/translations";
+import { Icon } from "~/components/common/Icon";
+import Box from "~/components/content/Box";
+import Hero from "~/components/content/Hero";
+import BMJVLogo from "~/assets/BMJV_Logo.svg?raw";
+
+const Homepage = () => {
+  const serviceCards = [
```
aaschlote: maybe we can have this variable and the `infoCards` outside of the component, so when we need to modify the content of the component, we don't touch the component file

app/routes/shared/components/Homepage.tsx
```diff
@@ -0,0 +1,262 @@
...
+      <GridSection pt="0" pb="80">
+        <Grid>
+          <GridItem
+            smColumn={{ start: 1, span: 12 }}
+            mdColumn={{ start: 1, span: 12 }}
+            lgColumn={{ start: 1, span: 12 }}
+            xlColumn={{ start: 1, span: 12 }}
+          >
+            <Heading
+              tagName="h2"
```
aaschlote, line 111: the `Card` component renders the title as `h2` also, wondering if this title is `h2` should the card titles be `h3`? Can you check it?

app/routes/_index.tsx
```diff
@@ -1,13 +1,11 @@
-import { useLoaderData } from "react-router";
-import ContentComponents from "~/components/content/ContentComponents";
 import { fetchPage } from "~/services/cms/index.server";
+import Homepage from "./shared/components/Homepage";
 
 export const loader = async () => {
   const { pageMeta, ...props } = await fetchPage("/");
```
aaschlote, line 5: maybe we don't need to call the function `fetchPage` and just hardcode the values from `meta` like the image below, as the whole content now on the homepage is hardcode.

Image

aaschlote (approved):
besides my comments, LGTM!

app/routes/shared/components/Homepage.tsx
ingriddorio: Following Kern, both are h2

app/routes/_index.tsx
ingriddorio: Without fetchPage the tile is not being rendered and test are failing. I will keep fetchPage for now.

app/routes/shared/components/Homepage.tsx
aaschlote: I understand on Kern style both are `H2`, however if you check on the mockup below you can see the titles on the cards are children of this title. And if you see today how it's our current structure on the Homepage, you can see we have a main title with `H2` and every service has a title with `H3`. I could be wrong, but on my perspective the titles for the services should be with `H3`. I would suggest to check with a Designer about it, because soon you apply the changes, to modify will need to be done via code. 

image

image

ingriddorio: So, 

Hero Heading is H1 
Headline above cards is H2
Text inside the card is a P
Heading in the cards are H3

PR #3990, opened 2026-07-22
APL-2292 array summary prevpath fix

melgaafary:
**Root cause:** createFlowSession.ts's "last occurrence" cycle logic (added in b72ad0c21 for the new angehoerige sub-flow) could pick a prevPath that requires more array indices than the current page provides — e.g. Back from a 0-index array-summary page resolving to a still-templated # path. No caller could safely turn that into a real URL; erbfolge's route crashed on it via resolveArrayCharacter's invariant.

**Fix:** when the picked predecessor needs more array depth than the current page, resolve concrete indices from the real (index-aware) visited context instead of returning a bare wildcard template.

### DIFF
--- a/app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
+++ b/app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
@@ -549,6 +549,44 @@ describe("createFlowSession", () => {
 
         expect(session.prevPath).toBe("/parents/0/children/0/detail");
       });
+
+      it("falls back to the first occurrence when the array is empty (add started but nothing submitted)", () => {
+        const arrayCompleteFlow = compileFlow({
+          pages: {
+            start: { stepId: "/start" },
+            overview: {
+              stepId: "/items",
+              arraySummary: {
+                name: "entries",
+                schema: z.array(z.object({ name: z.string() })),
+              },
+            },
+            itemDetail: {
+              stepId: "/items/#/detail",
+              pageSchema: { "entries#name": z.string() },
+            },
+            done: { stepId: "/done" },
+          },
+          initialStep: "start",
+          transitions: {
+            start: "overview",
+            overview: [
+              { target: "itemDetail", type: "addArrayItem" },
+              { target: "done" },
+            ],
+            itemDetail: "overview",
+            done: null,
+          },
+        });
+
+        const session = createFlowSession(
+          arrayCompleteFlow,
+          { pageData: { arrayIndexes: [] } } as any,
+          "/items",
+        );
+
+        expect(session.prevPath).toBe("/start");
+      });
     });
   });
 
--- a/app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
+++ b/app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
@@ -449,8 +449,105 @@ describe("createFlowSession", () => {
           "/items",
         );
 
-        // Back from the second overview should go to the item page (last occurrence)
-        expect(session.prevPath).toBe("/items/#/detail");
+        // Back from the second overview should go to the item page (last occurrence),
+        // with a concrete index — the overview page itself has no array index of its
+        // own to fall back on, so a bare "#" template would be unusable by callers.
+        expect(session.prevPath).toBe("/items/0/detail");
+      });
+
+      it("resolves the last completed item's own index, not always index 0", () => {
+        const arrayCompleteFlow = compileFlow({
+          pages: {
+            overview: {
+              stepId: "/items",
+              arraySummary: {
+                name: "entries",
+                schema: z.array(z.object({ name: z.string() })),
+              },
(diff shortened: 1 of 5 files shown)


app/services/flow/newFlowEngine/createFlowSession.ts
```diff
@@ -179,6 +189,46 @@ export const createFlowSession = <C extends PageConfigMap>(
...
   const prevNodeKey = skipFanOutOnlyBack(linearPrevNodeKey);
+
+  // The chosen prevNodeKey can sit deeper in array nesting than the current
+  // page (e.g. Back from a 0-wildcard array-summary page picked the last
+  // completed item's page, which has 1+ wildcards). The current page's own
+  // arrayIndexes can't fill those in — there's nothing at that depth to read.
+  // Resolve concrete indexes from the real (index-aware) visited context
+  // instead, so prevPath is always directly usable, never a bare "#" template.
+  const resolvePrevPath = (): string | undefined => {
+    if (prevNodeKey == null) return undefined;
```
Spencer6497: maybe this isn't an issue, but would scoping here possibly cause bugs in the future? i.e. `prevNodeKey` possibly being out-of-date when `resolvePrevPath()` is called, because the function was defined with a previous value? I know `createFlowSession` only gets theoretically called once, but wouldn't it be more bulletproof to pass-in all dependencies so we can guarantee they're up-to-date at call-time?

app/services/flow/newFlowEngine/createFlowSession.ts
```diff
@@ -179,6 +189,46 @@ export const createFlowSession = <C extends PageConfigMap>(
...
+      return compiledFlow.getPathFromNodeKey(
+        prevNodeKey as Extract<keyof C, string>,
+      );
+    }
+
+    // Most recently completed instance of that page wins (mirrors "last
+    // occurrence" cycle selection above).
+    const match = [...simulation.visitedContexts]
+      .reverse()
+      .find(
```
Spencer6497: nit: could use `findLast()`

app/services/flow/newFlowEngine/createFlowSession.ts
```diff
@@ -179,6 +189,46 @@ export const createFlowSession = <C extends PageConfigMap>(
...
+        ({ key, scopeData }) =>
+          key === prevNodeKey &&
+          isPageDone(
+            compiledFlow.getSchemaByNodeKey(key),
+            compiledFlow.getFieldNamesByNodeKey(key),
+            scopeData as Record<string, unknown>,
+          ),
+      );
+    if (!match) {
```
Spencer6497: for readability (though small performance hit), could move the `if (arrayWildcardCount)` conditional down here and `&&` them together, since they both call `compiledFlow.getPathFromNodeKey()`

app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
```diff
@@ -449,8 +449,105 @@ describe("createFlowSession", () => {
...
+          "/items",
+        );
+
+        expect(session.prevPath).toBe("/items/1/detail");
+      });
+
+      // Mirrors a real-world shape: a 0-wildcard array-summary page cycling
+      // through a *nested* array item (2 wildcards) before falling back to the
+      // summary. The picked predecessor sits two levels deeper than the current
+      // page, so both wildcards must be resolved, not just the first.
```
Spencer6497, line 500: 😵‍💫😵‍💫😵‍💫

Spencer6497 (approved):

app/services/flow/newFlowEngine/createFlowSession.ts
```diff
@@ -179,6 +189,46 @@ export const createFlowSession = <C extends PageConfigMap>(
...
   };
   const prevNodeKey = skipFanOutOnlyBack(linearPrevNodeKey);
+
+  // The chosen prevNodeKey can sit deeper in array nesting than the current
+  // page (e.g. Back from a 0-wildcard array-summary page picked the last
+  // completed item's page, which has 1+ wildcards). The current page's own
+  // arrayIndexes can't fill those in — there's nothing at that depth to read.
+  // Resolve concrete indexes from the real (index-aware) visited context
+  // instead, so prevPath is always directly usable, never a bare "#" template.
+  const resolvePrevPath = (): string | undefined => {
```
aaschlote: can we move this function to a new file? This function is big and it might need more refactoring in case we find more edge cases. Moving to a new file, we can add proper tests and make more readable the function .

melgaafary: I will remove it anyway, I tested multiple usecases, this solution will not work.

app/services/flow/newFlowEngine/createFlowSession.ts
melgaafary: it is actually an issue in some usecases. I created another solution

PR #3991, opened 2026-07-22
Feature: Enable deployment for feature branches

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-266


.github/workflows/ci-pipeline.yml
```diff
@@ -152,7 +162,7 @@ jobs:
         with:
           sarif_file: "trivy-results.sarif"
       - run: ./docker.sh --push prod
-      - run: ./docker.sh --sign
+      - run: ./docker.sh --sign prod
```
Spencer6497, line 165: why did this change? Was it signing all images before, and now only prod?

doc/e2e-tests.md
```diff
@@ -4,6 +4,11 @@
 >
 > Remember to install playwright cli and set the environment variable GERICHTSFINDER_ENCRYPTION_KEY
 
+## Commands
+
+- run: `pnpm run test:e2e`
+- UI: `pnpm run test:e2e:ui`
+
```
Spencer6497, line 11: ? random lol

docker.sh
```diff
@@ -22,7 +24,7 @@ function parseValidTarget() {
         exit 1
     fi
     case $2 in
-    app | content | prod) TARGET=$2 ;;
+    app | content | prod | app-testing-feature | prod-testing-feature) TARGET=$2 ;;
```
Spencer6497, line 27: do we really need both `app` and `prod` if its a test branch? Wouldn't it make more sense to just have `app-testing-feature`? Or am I forgetting something

Spencer6497 (approved):

.github/workflows/ci-pipeline.yml
aaschlote: before it was signed only the prod image, but now I need sign also the testing-feature image as well, this way I created this new arg to know what image be signed

doc/e2e-tests.md
aaschlote: I wanted to add a documentation for the testing-feature and I found a few things to be improved in our documentation, like this one.

docker.sh
aaschlote: yes we need, our Docker image that it runs in K8s needs two images, the `app` and the `content`, we pass them using the `build-arg` while building the production image. For the `testing-feature` image I need to build the `app` image with the current changes on the branch, but I don't need to `push` to our registry, only the one that it'll run in K8s.

melgaafary:
looks good! 🚀  thanks for updating the documentation as well! 🙇

PR #3993, opened 2026-07-23
refactor(New Flow Engine): Retrofit for new flows

Spencer6497:
This PR addresses two issues with the existing flow infrastructure to be able to handle newly-written flows: stepIds that start with a leading slash, and stepIds that feature an # in them.

### DIFF
--- a/app/domains/pageSchemas.ts
+++ b/app/domains/pageSchemas.ts
@@ -55,12 +55,15 @@ export const getAllPageSchemaByFlowId = (flowId: FlowId) => {
 export const getAllFieldsFromFlowId = (flowId: FlowId): FormFieldsMap => {
   const pagesConfig = pages[flowId];
   const fieldsMap: FormFieldsMap = {};
+  // TODO: Remove after old page configs are migrated to the new leading-slash format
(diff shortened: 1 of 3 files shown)


app/domains/pageSchemas.ts
aaschlote: Should you change this line too?

app/domains/pageSchemas.ts
```diff
@@ -58,7 +58,9 @@ export const getAllFieldsFromFlowId = (flowId: FlowId): FormFieldsMap => {
 
   for (const page of Object.values(pagesConfig)) {
     if (page.pageSchema && !isArrayParentPage(page)) {
-      const stepId = `/${page.stepId}`;
+      const stepId = page.stepId.startsWith("/")
```
aaschlote: I would say, add a TODO comment so we can remove this condition once we are done with the migration of the new flow

app/domains/pageSchemas.ts
Spencer6497: technically no, since only old flows reach this point (`isArrayParentPage` only tests for the presence of `arrayPages` in the pageSchema, which new flows don't have), buuuut my sonar fix also fixes this

PR #4003, opened 2026-07-24
refactor(New Flow Engine): Enable PDF downloads for new engine flows

Spencer6497:
This PR enables the PDF download of flows written using the new form engine. It also simplifies some types, increasing usability and type-safety.

- **roll pageData into InferredUserData, prune user data for new flow engine pdf, add TODO comments, refactor runSimulation to simulate.ts for reusability**
- **simplify types**
- **add tests**

### DIFF
--- a/app/services/flow/newFlowEngine/pruneUserData.ts
+++ b/app/services/flow/newFlowEngine/pruneUserData.ts
@@ -1,7 +1,12 @@
 import { ARRAY_WILDCARD } from "./compileFlow";
 import type { CompiledFlow } from "./compileFlow";
-import type { PageConfigMap, NodeKey, InferredUserData } from "./types";
-import type { PageData } from "../pageDataSchema";
+import {
+  type PageConfigMap,
+  type NodeKey,
+  type InferredUserData,
+} from "./types";
+import { type PageData } from "../pageDataSchema";
+import { runSimulation } from "~/services/flow/newFlowEngine/simulate";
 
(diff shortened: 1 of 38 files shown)


app/services/flow/newFlowEngine/pruneUserData.ts
```diff
@@ -70,3 +75,13 @@ export const pruneUserData = <C extends PageConfigMap>(
 
   return result as InferredUserData<C>;
 };
+
+export const getPrunedUserDataForPdf = <C extends PageConfigMap>(
```
aaschlote: we might need this function in other places, can we rename it to be generic? Maybe `getPrunedUserDataWithSimulation` or `getActualPrunedUserData`? Ticket for [reference](https://digitalservicebund.atlassian.net/browse/APL-244).

app/services/flow/newFlowEngine/types.ts
```diff
@@ -53,7 +53,7 @@ type ExtractNodeSchema<Node> = Node extends { pageSchema: infer S }
 
 export type InferredUserData<C extends PageConfigMap> = Partial<
   UnionToIntersection<ExtractNodeSchema<C[keyof C]>>
->;
+> & { pageData?: PageData };
```
aaschlote, line 56: nice, maybe we should evaluate later to make this value required, right now it's optional because we need to add direct into the userData of the domain.

aaschlote (approved):

PR #4015, opened 2026-07-27
APL-292 backbutton bug

melgaafary:
This reverts back the back button to go back to the previous page


app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
```diff
@@ -449,105 +452,7 @@ describe("createFlowSession", () => {
           "/items",
         );
 
-        // Back from the second overview should go to the item page (last occurrence),
-        // with a concrete index — the overview page itself has no array index of its
-        // own to fall back on, so a bare "#" template would be unusable by callers.
-        expect(session.prevPath).toBe("/items/0/detail");
-      });
-
-      it("resolves the last completed item's own index, not always index 0", () => {
```
aaschlote, line 458: why are we deleting this test?

melgaafary: I reverted back to a simpler back navigation. This test was for the more complicated version of the back button which goes back to the last finished page in this case it would be the last page of the added array item. Everywhere in the system we go back to the previous navigation page not the last array item. This was a mistake from my side, I was making the back button go to the last array item not the last navigation page

PR #4017, opened 2026-07-27
APL-297 Migrate Erbfolge as a flow using newEngine*

melgaafary:
# Migrate erbfolge onto the shared new-engine loaders (behind a feature flag)

## Summary

This PR moves erbfolge onto the shared new-engine loaders. Everything generic now
comes from the shared layer; everything erbfolge-specific lives in small, well-scoped
"extras" objects passed into that layer. The flow is registered as a real flow and
gated by a feature flag (`showNachlassErbscheinErbfolgeFlow`, **default off**)

## How it works now

**Registration.** Erbfolge is a normal entry in `flows.server.ts`, bundling the
compiled flow (`newEngineConfig`) and its `stringReplacements`. 

**Loader** — `loadVorabcheckData(args, erbfolgeVorabcheckExtras)`:
1. `getUserDataAndFlowNewEngine` parses the path, **enforces the feature flag**,
   builds the flow session, and redirects unreachable steps.
2. Erbfolge's `buildReplacements` resolves ancestor list-item names (`{{kinder#name}}`).
3. `retrieveContentData` renders the CMS page with the flow's `stringReplacements`
   (raw answers + missing-children data) merged with those ancestor names.
4. Erbfolge's `buildLoaderData` adds dynamic parent-select options, the
   kinder/elternteile array-summary data, the deceased's name, and parent-select
   fallback form elements.

**Action** — feature-flag check, then `runVorabcheckAction`: CSRF →
`validateFormUserData` → save session → `flowDestinationNewEngine` routes onward
from the engine's transitions (including into the `kinder-fehlen` exit page).

**Result page** — `loadResultData(args, erbfolgeResultExtras)`: same flag gate;
two hooks fire only on `/ergebnis/erbfolge`: `buildReplacements` injects the
`{{{requiredDocumentsHtml}}}` table, and `transformContent` runs
`calculateInheritance(...)` to fill the CMS `heirsList` List component.

## Key decisions & rationale

- **Extras as a simple ad-hoc object, injected into extracted loaders.** Extracted
  the loader/action bodies of `newEngineVorabcheck.ts` / `newEngineResult.ts` into
  `loadVorabcheckData` / `runVorabcheckAction` / `loadResultData` that take optional
  `extras`. The default `loader`/`action` exports delegate with no extras, so **every
  other flow is unchanged**.  No erbfolge only logic in shared files

- **Replacements split by what they can see.** Pure-userData replacements (the raw
  answers + the missing-children exit page) live in the flow's `stringReplacements`.
  Only the page-context ancestor names (which depend on the current array indexes)
  go through a new optional `extraReplacements` param on `retrieveContentData` — the
  static per flow hook can't know the current page.

- **Routes rewritten in place; The two thin route modules are the binding point that injects the extras, so `routes.ts` didn't
  change and nothing was deleted.

- **Array-summary intentionally hand-rolled.** The shared `getArraySummaryData` gates
  on a Strapi `statementKey === "yes"` and only reads top-level `userData[category]`;
  it can't handle erbfolge's *nested* arrays (`kinder#kinder`, resolved via array
  indexes) or its reachability based gating. So the bespoke derivation is deliberate,
  and `KinderSummary`/`ElternteilSummary` are kept as-is.

## Cleanup

- Removed the `ResultPageView` + `ResultPage` split. That split existed only so the
  old bespoke erbfolge route could inject custom React children; erbfolge now injects
  the heir list into the CMS `freeZone` instead, so nothing passes children. Collapsed
  into a single `ResultPage` (like `VorabcheckPage`).

## Testing

- New unit tests: `vorabcheckExtras` (array summaries, dynamic options, ancestor-name
  resolution), `stringReplacements` (missing-children exit page + escaping), and
  `resultExtras` (heir-list injection + required-documents, with step gating in both
  directions), driven against a real compiled flow session.


app/services/flow/server/progress.ts
```diff
@@ -58,7 +58,12 @@ export function progressLookupForMachine(machine: FlowStateMachine) {
 
 function computeVorabcheckProgress() {
   const vorabcheckEntries = Object.entries(flows).filter(
-    ([, flowConfig]) => flowConfig.flowType === "vorabCheck",
+    ([, flowConfig]) =>
+      flowConfig.flowType === "vorabCheck" &&
+      // New-engine flows get their progress from the new engine, so they never
+      // go through this xstate machine. Building one from their config is both
+      // unnecessary and, for a new-engine-only flow, impossible.
+      !("newEngineConfig" in flowConfig),
```
aaschlote, line 66: so this code it'll be deleted later, do we need this condition here?

Spencer6497 (approved):
Whew 😮‍💨 I'm not sure I fuuuuully understand all of it, but from what I gather, it looks good! Excellent work 🤝

app/domains/nachlass/erbschein/erbfolge/resultExtras.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (!ehepartnerName) return undefined;
+  return {
+    name: ehepartnerName,
+    gueterstand:
+      (userData as { gueterstand?: Gueterstand }).gueterstand ??
+      "communityOfAcquisitions",
+  };
+}
+
+export const erbfolgeResultExtras: ResultLoaderExtras = {
```
aaschlote, line 80: can we have also this function as the normal `stringReplacements`? or is it not possible?

app/routes/nachlass.erbschein.erbfolge.$.tsx
```diff
@@ -1,85 +1,48 @@
...
-  return `<ul>${names.map((name) => `<li>${escapeHtml(name)}</li>`).join("")}</ul>`;
-}
-function NachlassErbfolgePage() {
-  const loaderData = useLoaderData<typeof loader>();
+export const loader = (args: LoaderFunctionArgs) =>
+  loadVorabcheckData(args, erbfolgeVorabcheckExtras);
 
-  useFocusFirstH1();
+export const action = async (args: ActionFunctionArgs) => {
+  await throw404IfFeatureFlagDisabled(ERBFOLGE_FEATURE_FLAG);
```
aaschlote: I don't think you need this condition on the action, only in the loader

aaschlote (approved):
Besides my comments, LGTM!

app/services/flow/server/progress.ts
melgaafary: yes we still need this one otherwise I have to do a workaround for erbfolge

PR #4041, opened 2026-07-30
feat(New Flow Engine): Data Migration

Spencer6497:
This PR offers a potential implementation of the data migration feature for the new flow engine. Normally, it'll just grab whatever pruned data from the source flow matches the destination flow (same UserData keys). However, if a migrationDataMerger function is defined within the migration property, callers can specify exactly how the data gets merged, and define a mapping function to only take what they need.

- **groundwork for data migration**
- **build out migration infrastructure and add proof-of-concept usage**


app/domains/nachlass/erbschein/anfrage/index.ts
```diff
@@ -10,12 +10,36 @@ import {
...
+          yes: "yes" as const,
+          no: "no" as const,
+          unknown: undefined,
+        }[sourceData.ehevertrag ?? "unknown"],
+        verstorbeneFamilienstand: sourceData.familienstand,
+        verstorbeneVorname: sourceData.name?.split(" ")[0] ?? "",
+        verstorbeneNachname: sourceData.name?.split(" ")[1] ?? "",
+      };
+    },
+    buttonUrl: "/nachlass/erbschein/erbfolge", // do we need this?
```
aaschlote: we need in FGR because it has an button to restart the FGR Vorabcheck, but it's optional, you don't need to fill in our flow if you don't have it

app/domains/flows.server.ts
```diff
@@ -18,13 +18,26 @@ import { type Session } from "react-router";
...
+
+type FlowMigration<C extends PageConfigMap> = {
   source: FlowId;
   sortedFields: string[];
+  /**
+   * Custom merge function used in cases where source and destination flow have different schemas.
+   * @param sourceData userData of the source flow, which is being migrated from
+   * @returns merged userData belonging to the destination flow
+   */
+  migrationDataMerger?: MigrationDataMerger<C>;
```
aaschlote, line 40: add TODO to make this field required after the migration

app/services/flow/newFlowEngine/migrateSourceFlowDataToDestinationFlow.ts
```diff
@@ -0,0 +1,30 @@
...
+  destinationFlow: Flow<C>,
+  destinationFlowId: FlowId,
+) => {
+  if (destinationFlow.migration?.migrationDataMerger) {
+    return destinationFlow.migration.migrationDataMerger(sourceUserData);
+  }
+  const prunedData = createFlowSession(
+    sourceFlow.newEngineConfig!,
+    sourceUserData,
+    "/ergebnis/erbfolge",
```
aaschlote: not all the flows it'll have this page, so it's better to use the function `getPrunedUserDataFromSimulation` to retrieve the `prunedData`

app/services/session.server/getMigrationData.ts
```diff
@@ -18,6 +19,18 @@ export async function getMigrationData(
     return undefined;
 
   const userData = await getSessionData(migration.source, cookieHeader);
+  const sourceFlow = flows[migration.source];
+  if (
+    "newEngineConfig" in migrationFlowDestination &&
+    "newEngineConfig" in sourceFlow
```
aaschlote, line 26: Add a TODO here to remove this line later after the migration

aaschlote (commented):
You are missing to call the `getMigrationData` inside the function `getUserDataAndFlowNewEngine`, so we can render the component `MigrationDataOverview`.

It looks promising.

app/domains/flows.server.ts
Spencer6497: This field actually isn't required -- callers can actually leave the default behavior (only pull matching userData field names), or override the default behavior with a migrationDataMerger.

PR #4047, opened 2026-07-31
TGA Release 1 - Improvements

aaschlote:
A few improvements done on the [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/4027).

- Add workaround for the `geld-einklagen/formular` on the pageSchema and stepId on the `parsePathname`
- Add Textarea schema to the Fieldset schema
- Avoid creations of the children on the statusTree when `shouldCollapseIntoParentNavItem` is true
- Make possible to have two nested arrays inside one array. Changes from this [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/4058/commits)
- Create type of `UserData` of the flow service, checking first the `arraySummary` and later the `pageSchemas`.

These changes can be merged to main and it shouldn't effect our current `TGA` service. And it should make the [PR](https://github.com/digitalservicebund/a2j-rechtsantragstelle/pull/4027) a little smaller. 

Ticket: https://digitalservicebund.atlassian.net/browse/APL-266


app/components/content/summaryOverview/addArrayIndexToPathUrl.ts
```diff
@@ -9,5 +11,7 @@ export const addArrayIndexToPathUrl = (path: string, index: number): string => {
 
   segments.splice(startSegment, 0, index.toString());
 
-  return segments.join("/");
+  return segments
+    .join("/")
+    .replace(new RegExp(`/\\${ARRAY_WILDCARD}+/`, "g"), `/`);
```
Spencer6497: is the `+` necessary? Will there ever be a case where we see `##`? Could we instead use `{1}` for specificity?

melgaafary (approved):
LGTM :shipit:

Spencer6497 (approved):
LGTM as well:)

app/components/content/summaryOverview/addArrayIndexToPathUrl.ts
aaschlote: you are right, we shouldn't have this edge case. I'll change it

PR #4057, opened 2026-08-03
feat(New Flow Engine): Flow Transition Configs

Spencer6497:
This PR enables further feature parity of the new flow engine with the old engine, adding flow transition configs. This feature allows flows to be effectively guarded, in the case that a prerequisite flow isn't complete, like from /fluggastrechte/vorabcheck to the formular.


app/services/flow/newFlowEngine/compileFlow.ts
```diff
@@ -16,6 +17,7 @@ type Options<C extends PageConfigMap> = {
   initialStep: NodeKey<C>;
   transitions: TransitionConfigMap<C>;
   pruningStrategy?: PruningStrategy;
+  flowTransitionConfig?: FlowTransitionConfig;
```
aaschlote, line 20: we have already this element on the type `Flow`, so we don't need to have it here.

app/services/flow/newFlowEngine/flowTransitionValidationNewEngine.ts
```diff
@@ -0,0 +1,59 @@
...
+  eligibleSourcePages: string[];
+};
+
+type FlowTransitionResult = {
+  isEligible: boolean;
+  redirectTo?: string;
+};
+
+export function getFlowTransitionConfig(
+  currentFlow: CompiledFlow<PageConfigMap>,
```
aaschlote: Let's use the one from the `Flow` config.

app/services/flow/newFlowEngine/flowTransitionValidationNewEngine.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (!sourceFlowConfig) {
+    throw new Error(
+      `Source flow ${sourceFlowId} does not have a newEngineConfig, which is required for flow transitions.`,
+    );
+  }
+
+  const sourceFlowSession = createFlowSession(
+    sourceFlowConfig,
+    await getSessionData(sourceFlowId, cookieHeader),
+    sourceFlowConfig.initialPath,
```
aaschlote, line 39: double check if this value works. Reason:

User needs to reach the last page of the flow source (E.g.: `ergebnis/lastDestination`). Creation the flow session with the `currentPath` as the `initialPath`, it could be the flow session doesn't add the pages from the `eligibleSourcePages` as reachable.

aaschlote (commented):
It looks good, but I wrote some comments, please take a look.

app/services/flow/newFlowEngine/compileFlow.ts
Spencer6497: Riiiight for some reason my brain short circuited and thought that after the migration we'd be moving the existing `Flow` properties to `newEngineConfig`, instead of moving `newEngineConfig` to `Flow`... I'll change it back!

app/services/flow/newFlowEngine/flowTransitionValidationNewEngine.ts
Spencer6497: I tested this with the Erbscheinsanfrage flow, and also wrote [this test](https://github.com/digitalservicebund/a2j-rechtsantragstelle/blob/3ac2f4cedb20b7a4031600096cb42a73594b1d48/app/services/flow/newFlowEngine/__test__/flowTransitionValidationNewEngine.test.ts#L49), which should cover this case.

PR #4087, opened 2026-08-10
feat(Accessibility Mobile Sidenav): close the menu in mobile pressing escape key

ingriddorio:
This PR add logic to close the menu in mobile pressing escape key

Ticket: https://digitalservicebund.atlassian.net/browse/APL-243


app/components/navigation/SideNavMobile.tsx
```diff
@@ -87,6 +96,10 @@ const initializeKeyDownListeners = (
       summary.focus();
     }
   });
+
+  return () => {
+    document.removeEventListener("keydown", handleEscape);
```
Spencer6497, line 101: unrelated to this specific feature, but do the other event listeners also need to be cleared?

Spencer6497 (approved):

app/components/navigation/SideNavMobile.tsx
ingriddorio: Yeees, I was also thinking about it. Will check :)

ingriddorio: I changed it to clear the other event listeners as well

PR #4116, opened 2026-08-18
Improve autosuggestion input street

aaschlote:


app/components/formElements/schemaToForm/getDataListArgumentToAutoSuggestionInput.ts
```diff
@@ -0,0 +1,27 @@
...
+  const dataListType = fieldSchema.meta()?.type as DataListType;
+  const dataListArguments = fieldSchema.meta()?.dataListArguments as string[];
+
+  if (dataListType !== "streetNames" || !arrayIsNonEmpty(dataListArguments)) {
+    return "";
+  }
+
+  const zipCodeFieldNamesReference = dataListArguments.find(
+    (fieldName) =>
+      typeof userData[fieldName] === "string" && userData[fieldName].length > 0,
```
Spencer6497, line 19: We're certain that `userData`, at this point, will be sufficiently pruned, so that it doesn't just grab the first hit, in case something like `plzHospice` _and_ `plzPflegeheim` are both defined, right?

Spencer6497 (approved):

app/components/formElements/schemaToForm/getDataListArgumentToAutoSuggestionInput.ts
aaschlote: yes, the user data is pruned here :)

PR #4119, opened 2026-08-18
feat(Erbausschlagung Anfrage): add screen for user with appointment in court

ingriddorio:
This PR add pages to the flow for users with appointment in court

Ticket: https://digitalservicebund.atlassian.net/browse/APL-343

Screenshot 2026-08-18 at 15 19 03

Screenshot 2026-08-18 at 15 18 54

Screenshot 2026-08-18 at 15 18 40


aaschlote (changes requested):
Tha pages:

- `/gerichtstermin-bestaetigt`
- `/gerichtstermin-vereinbaren`
- `/datenverarbeitung`

should be inside the `/start` path, otherwise you will have the following problems:

- The side nav can't recognise the current item, because these pages don't have a parent `path` (you can already notice in the prints of this PR)
- The done functions doesn't apply correctly
- Having `shouldCollapseIntoParentNavItem` is workaround for not showing of the side nav, as it has been used in this PR.

app/domains/nachlass/erbausschlagung/anfrage/pages.ts
```diff
@@ -320,6 +329,14 @@ export const nachlassErbausschlagungAnfragePages = {
   abgabeZusammenfassung: {
     stepId: "/abgabe/zusammenfassung",
   },
+  abgabeBestaetigung: {
```
aaschlote, line 345: should we have this change in this PR?

aaschlote (approved):
LGTM!

I believe you need to fix the content below on Strapi.

Image

PR #4124, opened 2026-08-20
fix(Erbausschlagung): change optional schema to allow also empty strings

ingriddorio:
This PR change the erbausschlagungKinderArraySchema to accept empty string as well and don't forward user to `/kinder/warnung-nicht-ausgefuellt` page when the address is empty or not completely filled.

Ticket: https://digitalservicebund.atlassian.net/browse/ANL-37


app/domains/nachlass/erbausschlagung/anfrage/kinder/schema.ts
```diff
@@ -118,11 +118,11 @@ export const erbausschlagungKinderArraySchema = z
       ...commonErbausschlagungKinderFields,
       ...adultChild,
       ...livesSeparately,
-      strasse: stringOptionalSchema,
-      hausnummer: germanHouseNumberSchema.optional(),
-      plz: postcodeSchema.optional(),
-      ort: stringOptionalSchema,
-      adresseZusatz: stringOptionalSchema,
```
Spencer6497, line 125: is there a reason we can't preserve the original `stringOptionalSchemas`? As far as I understand it, we only need `schemaOrEmptyString` helpers, when we pass a specific, specially-validated schema in (like `postcode`)

aaschlote: in redis the data are saved with empty string and `.optional` allows `undefined` values. Zod fails because empty string is a value. We could touch the `stringOptionalSchema` to accept empty string, but this schema is used in more than 150 places 🥹

Image

ingriddorio: Yes, it is about "". Using schemaOrEmptyString I can bypass the ""

aaschlote (approved):
LGTM!

I'd recommend creating a ticket to update stringOptionalSchema so it accepts empty strings.

PR #4136, opened 2026-08-24
Make schema array to have at least one item

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-366


Spencer6497 (approved):
Thank youuuuu 🙏

PR #4142, opened 2026-08-26
feat(Auto-Generated Summary): Allow for field overrides

Spencer6497:
This PR enables the usage of overrides for the Auto-Generated Summary component. Namely, it's useful when a data structure has a purely in-code value that doesn't get displayed normally to the user, but requires extra processing, like `parentKindIndex`.

### DIFF
--- a/app/domains/flows.server.ts
+++ b/app/domains/flows.server.ts
@@ -6,10 +6,11 @@ import { fluggastrechteVorabcheck } from "~/domains/fluggastrechte/vorabcheck";
 import type { FlowTransitionConfig } from "~/services/flow/server/flowTransitionValidation";
 import type { Config } from "~/services/flow/server/types";
 import type { Replacements } from "~/util/applyStringReplacement";
+import type { ArrayFieldInfo, FieldItem } from "~/services/summary/types";
 import type { FlowId } from "./flowIds";
 import type { Guards } from "./guards.server";
 import { prozesskostenhilfeFormular } from "./prozesskostenhilfe/formular";
-import type { UserData } from "./userData";
+import type { UserData, AllowedUserTypes } from "./userData";
 import { geldEinklagenFormular } from "./geldEinklagen/formular";
 import { kontopfaendungPkontoAntrag } from "./kontopfaendung/pkonto/antrag";
 import { nachlassErbscheinWegweiser } from "~/domains/nachlass/erbschein/wegweiser";
@@ -49,6 +50,12 @@ type FlowMetaConfiguration = {
 
 export type FlowType = "vorabCheck" | "formFlow";
 
+export type SummaryFieldOverride = (
+  fieldInfo: ArrayFieldInfo,
+  rawValue: AllowedUserTypes,
+  userData: UserData,
+) => Pick<FieldItem, "question" | "answer"> | undefined;
+
 export type Flow<C extends PageConfigMap = PageConfigMap> = {
   flowType: FlowType;
   config: Config;
@@ -57,6 +64,11 @@ export type Flow<C extends PageConfigMap = PageConfigMap> = {
   migration?: FlowMigration<C>;
   flowTransitionConfig?: FlowTransitionConfig;
   stringReplacements?: (context: UserData) => Replacements;
+  /**
+   * Lets a flow override the auto-generated summary's question/answer for a
+   * specific field (e.g. internal fields with no CMS label, like parentKindIndex).
+   */
+  summaryFieldOverride?: SummaryFieldOverride;
   asyncFlowActions?: Record<
     string,
     (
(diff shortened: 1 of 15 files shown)


app/domains/nachlass/erbschein/shared/summaryFieldOverride.ts
```diff
@@ -0,0 +1,45 @@
...
+  const chosenValue = typeof rawValue === "string" ? rawValue : undefined;
+
+  const options = resolveParentOptions(fieldName, userData, arrayIndexes);
+  const parentName =
+    chosenValue === BOTH_PARENTS_VALUE
+      ? options.find((option) => option.value === BOTH_PARENTS_VALUE)?.text
+      : options.find((option) => option.value === chosenValue)?.text;
+
+  return {
+    question: "Kind von",
```
aaschlote: I would move it to the translation file and the string below too.

app/domains/nachlass/erbschein/shared/summaryFieldOverride.ts
```diff
@@ -0,0 +1,45 @@
...
+  "parentKindIndex",
+  "parentElternteilIndex",
+]);
+
+/**
+ * parentKindIndex/parentElternteilIndex (i.e. which parent a given descendant
+ * belongs to) are purely metadata and have no CMS label. Display them as
+ * "Kind von: <Elternname>" instead of the raw stored index and label.
+ */
+export const resolveParentIndexSummaryOverride: SummaryFieldOverride = (
```
aaschlote: I have a problem with the name for function starts with `resolve`, because what are we resolving here? It makes sense if the function return `Promise.resolve`, but it's not the case for this function.

app/services/summary/fieldEntryCreation.ts
```diff
@@ -85,6 +86,25 @@ const getValueAndArrayData = (
...
+ * Lets the calling flow override the question/answer for a field that has no
+ * CMS label (e.g. purely metadata fields like parentKindIndex).
+ */
+function resolveFieldOverride(
+  fieldInfo: ReturnType<typeof parseArrayField>,
+  value: AllowedUserTypes,
+  userData: UserData,
+  representativeStepId: string,
+): ReturnType<SummaryFieldOverride> {
+  try {
```
aaschlote, line 99: this `try` `catch` is because the possible throw exception from the function `getPageAndFlowDataFromPathname`?

app/services/summary/fieldEntryCreation.ts
```diff
@@ -85,6 +86,25 @@ const getValueAndArrayData = (
   return { value, arrayIndex, arrayBaseField };
 };
 
+/**
+ * Lets the calling flow override the question/answer for a field that has no
+ * CMS label (e.g. purely metadata fields like parentKindIndex).
+ */
+function resolveFieldOverride(
```
aaschlote: same as my comment above, what are resolving here?

app/services/summary/__test__/fieldEntryCreation.test.ts
```diff
@@ -134,6 +137,30 @@ describe("fieldEntryCreation", () => {
       );
     });
 
+    it("should handle field with an override", () => {
```
aaschlote, line 140: maybe you can mock the return value of the `resolveParentIndexSummaryOverride` and test if the `createFieldEntry` returns it.

app/domains/flows.server.ts
```diff
@@ -49,6 +50,12 @@ type FlowMetaConfiguration = {
 
 export type FlowType = "vorabCheck" | "formFlow";
 
+export type SummaryFieldOverride = (
```
aaschlote, line 53: this Override rule is only for array items or also for other item values? If it's only for the array items, maybe we can rename the type and functions to `SummaryFieldOverrideArrayItems`.

aaschlote (commented):
I wrote some comments, but it looks promising.

app/domains/nachlass/erbschein/shared/summaryFieldOverride.ts
Spencer6497: we are resolving the override, if the flow has one. "Resolve" in the sense of the verb "to fetch/obtain", not in the programmatic sense of a `Promise.resolve`, but I can see how that could be confusing. I'll change the prefix to just "get" :)

app/services/summary/fieldEntryCreation.ts
Spencer6497: see above comment. Will change it to `getFieldOverride`

app/domains/flows.server.ts
Spencer6497: Mmmm technically it's for all item values, but the call chain is reaaaally confusing inside of `createFieldEntry`, because it's always passing around an `ArrayFieldInfo`... which doesn't necessarily always describe an array. I'm gonna try and refactor it, if not for any other reason than to rename some stuff.

PR #4144, opened 2026-08-26
Feature: Create configuration to extend lifecycle of TGA data

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-370


app/services/session.server/index.ts
```diff
@@ -33,35 +37,49 @@ const generateHeader = (cookies: string[]) => {
...
-  httpOnly: true,
-  maxAge: 24 * 60 * 60,
-  secure: useSecureCookie,
-});
+const getCookie = () =>
+  createCookie("__session", {
+    secrets: [config().COOKIE_SESSION_SECRET],
+    sameSite: "lax",
+    httpOnly: true,
+    maxAge: getMaxAgeLifecycle(),
```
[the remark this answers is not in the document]
aaschlote: the "main" reference will be created again in Redis in case it's deleted

app/services/session.server/__test__/lifecycleSession.test.ts
```diff
@@ -0,0 +1,20 @@
+import { getLifecycleTimeBySessionUserData } from "../lifecycleSession";
+
+describe("getLifecycleTimeBySessionUserData", () => {
+  it("should return the default time to live for 'main' context", () => {
+    const result = getLifecycleTimeBySessionUserData("main");
+    expect(result).toBe(86400); // 24 hours in seconds
+  });
+
+  it("should return the lifecycle time for a specific flow context", () => {
```
m0dh4x, line 9: Hey @aaschlote nice work :) One idea, should we add a test for `getMaxAgeLifecycle`? The invariant "cookie lifetime ≥ longest Redis TTL" is what makes this work, and it's currently untested?

m0dh4x:
@aaschlote one think that runs through while i checked sth for deletion user date in /`action.delete-data.tsx` set cookie sets `__vault` but it should be `__vaultKey`, right? Could you fix it. Think its related? 
"

aaschlote:
> @aaschlote one think that runs through while i checked sth for deletion user date in /`action.delete-data.tsx` set cookie sets `__vault` but it should be `__vaultKey`, right? Could you fix it. Think its related? "

Ohh nice catch, I'll make the fix on `main`

m0dh4x (approved):
beside my comments LGTM

app/services/session.server/__test__/lifecycleSession.test.ts
aaschlote: yeah, I was wondering about it, but it requires to mock one external file, but I can do it.

PR #4154, opened 2026-08-28
feat(Erbscheinsanfrage): Auto-fill formular with Antragstellende Person data

Spencer6497:
This PR adds an asyncFlowAction that re-uses answers given by the Antragstellende Person to fill later parts of the Antrag, depending on their relationship to the Erblasser. This is because it is often the case that the Antragstellende Person is also Erbberechtigt, and would therefore have to redundantly re-enter their data when it comes time to fill out Beguenstigten/Ehepartner/Angehoerige.

Not included is the handling of missing data -- In the event that the Antragstellende Person is an Enkelkind or sibling/niece/nephew of the Erblasser, we have to effectively create "dummy" entries to hold their data: i.e., a parent with no valid data, apart from the fact that their `kinder` array contains the Antragstellende Person themselves (see `antragstellendePersonToDescendant`). For this, there is a follow-up ticket to display missing data to users.


app/domains/flows.server.ts
```diff
@@ -56,6 +56,12 @@ export type SummaryFieldOverride = (
   userData: UserData,
 ) => Pick<FieldItem, "question" | "answer"> | undefined;
 
+export type AsyncFlowAction<TUserData extends UserData = UserData> = (
```
aaschlote, line 59: nice improvement on the type, do u think can you adapt it to for the other function that are using the `asyncFlowAction`?

aaschlote (approved):
LGTM!

melgaafary:
looks good and tested locally

PR #4171, opened 2026-09-02
refactor(translations): separate translations into files

ingriddorio:


app/components/content/arraySummary/ArraySummary.tsx
```diff
@@ -63,7 +63,7 @@ const ArraySummary = ({
           disabled={disableAddButton}
           data-testid={`add-${category}`}
         >
-          {`${buttonLabel} ${translationProvider.arraySummary.arrayAddButtonLabel.de}`}
+          {`${buttonLabel} ${commonTranslations.common.add.de}`}
```
m0dh4x, line 66: Hey Ingrid, thanks for cleaning this up :) great! :) 

One thing though: the interpolation in `ArraySummary.tsx:66` breaks German
capitalization.

```tsx
{`${buttonLabel} ${commonTranslations.common.add.de}`} 

- before: "Kind hinzufügen"
- now:    "Kind Hinzufügen"?

File: app/domains/nachlass/erbschein/shared/__test__/summaryFieldOverride.test.ts [hunk 2]
```diff
@@ -55,6 +55,6 @@ describe("getParentIndexSummaryOverride", () => {
 
     const result = getParentIndexSummaryOverride(fieldInfo, "0", userData);
 
-    expect(result).toEqual({ question: "Kind von", answer: "Keine Angabe" });
+    expect(result).toEqual({ question: "Kind von", answer: "Keine Angaben" });
```
m0dh4x, line 58: Was this change from `Keine Angabe` to `Keine Angaben` intentional? I noticed that the test expectation was updated as well, but the test description still explicitly refers to `Keine Angabe`

app/components/content/arraySummary/ArraySummary.tsx
m0dh4x: Maybe it make sense to add a small regression test covering composed labels and existing fallback values, so future translation reorganizations do not unintentionally change user-facing wording? :)

PR #4173, opened 2026-09-03
Use zod validated instead of safe parse

aaschlote:
### Background
Zod version [4.5.0](https://github.com/colinhacks/zod/releases/tag/v4.5.0) introduced `zod.validate()` as an alternative to `safeParse()`. It provides a fast path to verify input validity without a full parse, making it up to 16x faster on invalid data. This PR implements `zod.validate()` wherever possible to improve performance in both our tests and production environments.


melgaafary (approved):
So much cleaner! thanks for doing this

m0dh4x (approved):
Hey @aaschlote  nice change and think for production is good. I'm wondering whether we should keep `safeParse()` in the tests cause safeParse() gives us more detail on what failed exactly while `z.validate` only gives us `expected true, received false`. Am I right?

aaschlote:
> Hey @aaschlote nice change and think for production is good. I'm wondering whether we should keep `safeParse()` in the tests cause safeParse() gives us more detail on what failed exactly while `z.validate` only gives us `expected true, received false`. Am I right?

yeah, this is true, if you only need to see if the input is valid from a schema, the `z.validate` is enough

m0dh4x:
> yeah, this is true, if you only need to see if the input is valid from a schema, the `z.validate` is enough

Yes, functionally it’s enough. My concern is only about debugging unexpected test failures, since `z.validate()` no longer provides the validation error. Is the performance benefit in tests worth losing that information?

aaschlote:
> > yeah, this is true, if you only need to see if the input is valid from a schema, the `z.validate` is enough
> 
> Yes, functionally it’s enough. My concern is only about debugging unexpected test failures, since `z.validate()` no longer provides the validation error. Is the performance benefit in tests worth losing that information?

I just updated the source code where we don't need more info about the error, the tests that it needs to check the error message, I keep them as `safeParse`.

m0dh4x:
> > > yeah, this is true, if you only need to see if the input is valid from a schema, the `z.validate` is enough
> > 
> > 
> > Yes, functionally it’s enough. My concern is only about debugging unexpected test failures, since `z.validate()` no longer provides the validation error. Is the performance benefit in tests worth losing that information?
> 
> I just updated the source code where we don't need more info about the error, the tests that it needs to check the error message, I keep them as `safeParse`.

Maybe we're just talking past each other. A concrete example is `app/domains/__test__/flowSchema.test.ts`. Previously, a failure of `expect(validationResult.error).toBeUndefined()` would show the received `ZodError`, including the affected field and validation rule. After changing it to `expect(z.validate(...)).toBe(true)`, an unexpected failure only shows `Received: false`. This is the diagnostic information I was referring to.

aaschlote:
> > > > yeah, this is true, if you only need to see if the input is valid from a schema, the `z.validate` is enough
> > > 
> > > 
> > > Yes, functionally it’s enough. My concern is only about debugging unexpected test failures, since `z.validate()` no longer provides the validation error. Is the performance benefit in tests worth losing that information?
> > 
> > 
> > I just updated the source code where we don't need more info about the error, the tests that it needs to check the error message, I keep them as `safeParse`.
> 
> Maybe we're just talking past each other. A concrete example is `app/domains/__test__/flowSchema.test.ts`. Previously, a failure of `expect(validationResult.error).toBeUndefined()` would show the received `ZodError`, including the affected field and validation rule. After changing it to `expect(z.validate(...)).toBe(true)`, an unexpected failure only shows `Received: false`. This is the diagnostic information I was referring to.

This is true, but the test wasn't asserting in details the values of the `validationResult.error` like message or field. So `expect(validationResult.error).toBeUndefined()` and `expect(z.validate(...)).toBe(true)` are basically the same. Also the `validationResult.error` contains values if is not valid.

PR #4176, opened 2026-09-04
Improve testing-feature environment deployment

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-394


.github/workflows/ci-pipeline.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           retention-days: 1
 
   testing-feature:
-    if: github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/APL-266-feature-release'
+    if: github.event_name == 'workflow_dispatch' && startsWith(github.ref, 'refs/heads/testing-feature/')
```
melgaafary, line 104: what do you think we make it `endsWith`?  just to keep the pattern we have for branch names `[ticket-number]-description` so it will be` [ticket-number]-description/testing-feature`

aaschlote: The main problem is on the `platform` infra police, where I need to create a regex rule for what branches to be accepted to be pulled to the cluster. I tried to create a flexible rule in this [PR](https://github.com/digitalservicebund/platform/pull/1325), but Kyverno doesn't support it at the moment.

I also believe start the branch name with `testing-feature/` is a bit easier to identify what branches are able to be deployed, so it would be like a folder structure.

```
testing-feature
 -> feature-1
 -> feature-2
 -> feature-3
```

And we can keep the current pattern later on: `testing-feature/[ticket-number]-description`.

Spencer6497 (approved):

PR #4179, opened 2026-09-07
Apl 371 remove nachlass from url

aaschlote:
Ticket: https://digitalservicebund.atlassian.net/browse/APL-371


app/domains/extraLoaderConfiguration.ts
```diff
@@ -17,6 +17,8 @@ const extraFlowFeaturesById: Partial<Record<FlowId, FlowExtras>> = {
   "/nachlass/erbschein/erbfolge": erbfolgeExtras,
```
Spencer6497: i assume these will also be deleted after the migration?

app/services/flow/userDataAndFlow/getUserDataAndFlow.ts
```diff
@@ -41,9 +41,7 @@ type ErrorResult = {
 };
 
 const flowIdFeatureFlag: Partial<Record<FlowId, FeatureFlag>> = {
-  "/nachlass/erbausschlagung/anfrage": "showErbausschlagungFlow",
-  "/nachlass/erbausschlagung/gericht-finden":
-    "showErbausschlagungGerichtFindenVorabcheck",
```
Spencer6497, line 46: are we not feature flagging this anymore?

Spencer6497 (approved):

app/services/flow/userDataAndFlow/getUserDataAndFlow.ts
aaschlote: this is the old function (with `xstate`), we don't need it, all the `nachlass` has been migrated with the new engine.

app/domains/extraLoaderConfiguration.ts
aaschlote: yes, I'll write a comment

melgaafary:
tested locally, works fine! 🚀

PR #4189, opened 2026-09-08
APL-420 Fix array nav collapse,

melgaafary:
Array sub-sections no longer leak into the side nav. When array item pages sit under an extra folder (e.g. `.../bankkonten/bankkonto/#/daten`), the nav showed a stray, unlabeled entry (the raw URL). A collapsed page now only attaches to sections a non collapsed page already creates, so array pages fold into their real parent (e.g. "Eigentum").


aaschlote (approved):
As we discussed, we need to add a new a regression test for the problem we saw when removed the function `getSegmentsBeforeFirstWildcard`.

Besides that, LGTM!

PR #4193, opened 2026-09-09
feat(Erbscheinsanfrage): Multi-field Validation

Spencer6497:
This PR enables multi-field validation in the Erbscheinsanfrage flow, namely validating that birth dates and death dates are logically coherent.

- **add multi-fields validation for verstorbene Person, add birthDateDeathDate validation helper, enable object parsing for hiddenInputs**
- **add validation to beguenstigsten field**
- **consolidate validator into one function**
- **validate angehoerige section, add unit test and update existing unit tests**
- **fix logical error, add test cases**
- **finish validating birthdays and death dates**


app/domains/multiFieldsFlowValidation.ts
```diff
@@ -12,6 +13,7 @@ const multiFieldsFlowValidation = {
     fluggastrechtMultiFieldsValidation as MultiFieldsStepIdValidation,
   "/geld-einklagen/formular":
     geldEinklagenMultiFieldsValidation as MultiFieldsStepIdValidation,
+  "/nachlass/erbschein/anfrage": nachlassErbscheinAnfrageMultiFieldsValidation,
```
aaschlote: you might change it, we will remove this flow id soon

app/components/formElements/inputs/hidden/HiddenInput.tsx
```diff
@@ -1,7 +1,25 @@
...
+    return (
+      <>
+        {Object.entries(inputProps.defaultValue).map(([key, value]) => (
+          <input
+            key={key}
+            data-testid={`hidden-input-${key}.${name}`}
+            name={`${name}.${key}`}
+            value={value as string}
+            hidden
+            readOnly
```
aaschlote: checking this [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly#overview), `readOnly` is not relevant for hidden inputs, do we really need it?

app/components/formElements/inputs/hidden/HiddenInput.tsx
```diff
@@ -1,7 +1,25 @@
...
+          />
+        ))}
+      </>
+    );
+  }
+  return <input {...inputProps} hidden readOnly />;
```
aaschlote: we need to keep the attribute `id` when render the input due dependencies of the library `@rvf/react-router`, can you check if it's rendering?

app/domains/nachlass/services/validation/validateBirthDateDeathDate.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      if (
+        type === "birthDate" &&
+        geburtsdatum &&
+        sterbedatum &&
+        toDate(geburtsdatum) > toDate(sterbedatum)
+      ) {
+        ctx.issues.push({
+          code: "custom",
+          message: translations.nachlass.birthDateAfterDeathDateError.de,
```
aaschlote, line 19: I would prefer to have a custom message that it can be changed on Strapi, so it makes easier for the content creator change the error message based on the page.

app/domains/nachlass/services/validation/validateBirthDateDeathDate.ts
```diff
@@ -0,0 +1,49 @@
+import { z } from "zod";
+import { type SchemaObject } from "~/domains/userData";
+import { translations } from "~/services/translations/translations";
+import { type DateObject, toDate } from "~/services/validation/dateObject";
+
+export function validateBirthDateDeathDate(
```
aaschlote: Instead of having a function that validate two types of validation, I would create two functions, `validateBirthDate` and `validateDeathDate`. They have a single responsibility, keep easy to understand what the function does and make the source logical smaller (and also having one less parameter).

app/domains/nachlass/services/__test__/validateBirthDateDeathDate.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  it("shouldn't return any errors if the birth date is before the death date", () => {
+    const birthDate = { day: "1", month: "1", year: "2000" };
+    const deathDate = { day: "1", month: "1", year: "2020" };
+    const result = validateBirthDateDeathDate(
+      "birthDate",
+      "deathDate",
+      "birthDate",
+    )(schema).safeParse({ birthDate, deathDate });
```
aaschlote: you can use `zod.validate` if you only need to see if the value is true or false.


```
const result = z.validate(
      validateBirthDateDeathDate("birthDate", "deathDate", "birthDate")(schema),
      { birthDate, deathDate },
    );
    expect(result).toBe(true);
```

app/components/formElements/inputs/hidden/HiddenInput.tsx
Spencer6497: Ah yes, you're correct, it removed the rendering of the id, but strangely enough, `getInputProps` doesn't officially support an `id` prop.... for now I'll just manually define `id={}` on the markup itself

app/components/formElements/inputs/hidden/HiddenInput.tsx
Spencer6497: fixed!

app/domains/nachlass/services/validation/validateBirthDateDeathDate.ts
Spencer6497: Mmmm normally I'd agree with you, but I think in this case it makes much more sense to have a single error message (that rarely changes) in one place, as opposed to ~30 in Strapi as references

app/domains/nachlass/services/validation/validateBirthDateDeathDate.ts
Spencer6497: I did have that initially, but didn't like the ergonomics of using it, plus it was sort of confusing that a single typo could logically disrupt the validation... but let me see how it looks now that all of the validation is in place

PR #4221, opened 2026-09-15
APL-420 new engine array page index shouldnt go out of bound

melgaafary:
1. New arrayIndexesInBounds helper walks the nested array path and checks each requested index against the actual array length (allowing index === length for the add-new-item slot).
2. After resolving the nodeKey, if the page is an array page and the requested index is out of bounds, it throws — mirroring the existing Invalid path throw.


app/services/flow/newFlowEngine/__test__/createFlowSession.test.ts
```diff
@@ -195,6 +195,26 @@ describe("createFlowSession", () => {
         /Invalid path/,
       );
     });
+
+    it("throws when an array index exceeds the array length", () => {
```
aaschlote, line 199: can we have a test for nested arrays too? Maybe until the second level is enough

aaschlote (approved):
LGTM!

PR #4234, opened 2026-09-18
fix(HiddenInput): Skip rendering of empty strings

Spencer6497:
This PR fixes a bug where HiddenInputs get rendered, even when the underlying data doesn't exist

- **skip rendering of empty hiddenInput values**
- **correctly merge migrationData into retrieveContentData**
- **refactor validateFormUserData to not return migrationData, to separate concerns**


app/components/formElements/inputs/hidden/HiddenInput.tsx
```diff
@@ -22,6 +22,9 @@ const HiddenSubfieldInput = ({
 const HiddenInput = ({ name }: { name: string }) => {
   const { getInputProps } = useField(name);
   const inputProps = getInputProps();
+  if (!inputProps.defaultValue) {
```
aaschlote, line 27: maybe we can add a comment why we are doing it

app/services/flow/contentData/retrieveContentData.ts
```diff
@@ -68,7 +68,7 @@ export const retrieveContentData = async (
 
   return getContentData(
     { cmsContent, translations },
-    userDataWithPageData,
+    { ...userDataWithPageData, ...migrationData },
```
aaschlote, line 71: why do we need this change?

aaschlote (approved):
Besides my comment, LGTM!

app/services/flow/contentData/retrieveContentData.ts
Spencer6497: otherwise migration data (like from FGR Vorabcheck) is missing from the hiddenInputs

